PR #1, opened 2018-01-03
A variety of clean ups

mrocklin:


daskernetes/__init__.py
```diff
@@ -52,31 +63,10 @@ def _make_pod(self):
...
-                            ),
-                            client.V1EnvVar(
-                                name = 'POD_NAME',
-                                value_from = client.V1EnvVarSource(
-                                    field_ref = client.V1ObjectFieldSelector(
-                                        field_path = 'metadata.name'
-                                    )
-                                )
-                            ),
-                        ],
```
mrocklin: These don't seem necessary to me yet.  Happy to be overruled though.

daskernetes/__init__.py
```diff
@@ -147,17 +137,15 @@ def main():
...
         args.namespace,
         args.worker_image,
         {},
-        scheduler.address
+        n_workers=1,
     )
-    adapative_cluster = Adaptive(scheduler, cluster)
-    tornado.ioloop.IOLoop.current().start()
+    client = Client(cluster)
+    print(client.submit(lambda x: x + 1, 10).result())
```
mrocklin: This is just here for testing at the moment.  We should probably set up a CI system with minikube at some point.  I ran into https://blog.travis-ci.com/2017-10-26-running-kubernetes-on-travis-ci-with-minikube .  Other suggestions welcome.

yuvipanda:
Aha, you are right we don't actually explicitly need to give it its IP or hostname, since kubernetes actually sets up network devices properly to have the pod IP!

daskernetes/__init__.py
```diff
@@ -1,26 +1,30 @@
...
+            namespace='dask',
+            worker_image='daskdev/dask:latest',
+            worker_labels=None,
+            n_workers=0,
+            threads_per_worker=1,
+            host='0.0.0.0',
     ):
-        self.log = logging.getLogger("distributed.deploy.adaptive")
+        self.cluster = LocalCluster(ip=host or socket.gethostname(),
+                                    n_workers=0)
```
mrocklin: LocalCluster also sets up things for us like the dashboard, which can be pleasant.  I'm happy to do things more raw with `Scheduler` though if preferred.

PR #14, opened 2018-01-04
Add IPython widget to KubeCluster

mrocklin:
This is kind of gimicky, but plays well

image


yuvipanda:
I like it! I also like the ipywidgets is optional!

PR #15, opened 2018-01-05
Create namespaces on demand

mrocklin:
I suspect that there are more idiomatic ways to do this, perhaps with `generate_name`?

Review and critical feedback would be appreciated.

Also, our tests are a bit wonky.  I suspect that we are not cleaning up correctly between tests.

### DIFF
--- a/daskernetes/core.py
+++ b/daskernetes/core.py
@@ -2,6 +2,7 @@ import logging
 import socket
 import argparse
 from urllib.parse import urlparse
+import uuid
 from weakref import finalize, ref
 
 from tornado import gen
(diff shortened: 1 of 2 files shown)


jacobtomlinson:
Agreed. I everything we do should be namespace agnostic. There are a few issues which already spring to mind:

- As @yuvipanda says with working RBAC the pod may not be able to create namespaces.
- The person deploying this onto the cluster may not be in a position to create namespaces (a developer with access to a shared cluster who has a namespace for their project).
- Namespaces will be used in different ways at different organisations and we should remain agnostic to that.

mrocklin:
OK, will do

> namespace should default to either current namespace or to specified one

Is "current namespace" something that we can query kubernetes for?  Is this typically stored in an environment variable perhaps?

yuvipanda:
@mrocklin https://github.com/jupyterhub/kubespawner/blob/master/kubespawner/spawner.py#L125 shows the easiest way.

yuvipanda:
Also UUIDs might not actually work for name defaults, since there's a 63 character limit on pod names. Perhaps something like 'username-${truncated-uuid}'?

PR #17, opened 2018-01-05
Use default namespace and unique name

mrocklin:
xref #15

### DIFF
--- a/daskernetes/core.py
+++ b/daskernetes/core.py
@@ -1,3 +1,4 @@
+import getpass
 import logging
 import os
 import socket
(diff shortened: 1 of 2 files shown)


daskernetes/core.py
```diff
@@ -38,8 +39,14 @@ def __init__(
 
         self.api = client.CoreV1Api()
 
+        if namespace is None:
+            namespace = _namespace_default()
+
+        if name is None:
+            name = 'dask-%s-%s' % (os.environ['USER'], str(uuid.uuid4())[:10])
```
yuvipanda: https://docs.python.org/3/library/getpass.html is probably a better bit to use here than calling $USER directly.

PR #22, opened 2018-01-07
WIP - Pangeo deployment

mrocklin:
I'm actively deploying off of this branch.  I'm putting it here just for comments.

### DIFF
--- a/daskernetes/core.py
+++ b/daskernetes/core.py
@@ -159,15 +159,12 @@ class KubeCluster(object):
         import ipywidgets
         layout = ipywidgets.Layout(width='150px')
         n_workers = ipywidgets.IntText(0, description='Workers', layout=layout)
-        threads = ipywidgets.IntText(1, description='Cores', layout=layout)
         actual = ipywidgets.Text('0', description='Actual', layout=layout)
         button = ipywidgets.Button(description='Scale', layout=layout)
-        box = ipywidgets.HBox([ipywidgets.VBox([n_workers, threads]),
-                               ipywidgets.VBox([button, actual])])
+        box = ipywidgets.VBox([n_workers, actual, button])
         self._cached_widget = box
 
         def cb(b):
-            self.threads_per_worker = threads.value
             n = n_workers.value
             self.scale_up(n)
 
--- a/daskernetes/core.py
+++ b/daskernetes/core.py
@@ -88,7 +88,7 @@ class KubeCluster(object):
             labels=None,
             n_workers=0,
             host='0.0.0.0',
-            port=8786,
+            port=0,
             env={},
             **kwargs,
     ):
--- a/daskernetes/core.py
+++ b/daskernetes/core.py
@@ -16,6 +16,27 @@ from kubernetes import client, config
 logger = logging.getLogger(__name__)
 
 
+def make_worker_spec(image='daskdev/dask:latest',
+                     threads_per_worker=1,
+                     env={}):
+    return client.V1PodSpec(
+        restart_policy='Never',
+        containers=[
+            client.V1Container(
+                name='dask-worker',
+                image=image,
+                args=[
+                    'dask-worker',
+                    '--nthreads', str(threads_per_worker),
+                    '--no-bokeh',
+                ],
+                env=[client.V1EnvVar(name=k, value=v)
+                     for k, v in env.items()],
+            )
+        ]
+    )
+
+
 class KubeCluster(object):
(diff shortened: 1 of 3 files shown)


daskernetes/core.py
```diff
@@ -55,69 +82,96 @@ class KubeCluster(object):
...
-        self.threads_per_worker = threads_per_worker
-        self.env = dict(env)
+        labels = dict(labels or {})
+        env = dict(env)
 
         # Default labels that can't be overwritten
-        self.worker_labels['org.pydata.dask/cluster-name'] = name
-        self.worker_labels['app'] = 'dask'
-        self.worker_labels['component'] = 'dask-worker'
+        labels['org.pydata.dask/cluster-name'] = name
```
mrocklin: @yuvipanda can I ask what the motivation is here?  Is this a standard namespacing scheme?

daskernetes/core.py
```diff
@@ -55,69 +82,96 @@ class KubeCluster(object):
...
+            except KeyError:
+                worker_spec = make_worker_spec()
+
+        worker_spec = deserialize_pod_spec(worker_spec)
+        self.worker_spec = worker_spec
+
+        self.worker_meta = client.V1ObjectMeta(
+                generate_name=name + '-',
+                labels=labels
+        )
```
mrocklin: Is this a sensible approach?  To get a pod spec from the user and create metadata ourselves?

daskernetes/core.py
```diff
@@ -294,6 +324,60 @@ def _namespace_default():
...
+    Returns
+    -------
+    kubernetes.client.V1Pod
+    """
+    if isinstance(x, str):
+        import yaml
+        with open(x) as f:
+            x = yaml.load(f)
+    if isinstance(x, dict):
+        x = client.ApiClient().deserialize(_FakeResponse(x), cls)
```
mrocklin: I would love to find a way to make this function work.  @yuvipanda I think that you pointed me to something in kubespawner but after several minutes searching through gitter logs I'm unable to find it.

mrocklin: the function below is the hack I used to get around this.  It's entirely specialized for my application.

daskernetes/core.py
yuvipanda: yeah, although I got it wrong! Docker labels are reversed like this, while k8s labels are not. I'l fix.

yuvipanda:
I'm going to pick out commits from here and make separate PRs.

yuvipanda:
First PR out of this: https://github.com/yuvipanda/daskernetes/pull/25

yuvipanda:
Ok, I closed #25, and rolled it into the bigger https://github.com/yuvipanda/daskernetes/pull/26. I looked at the worker YAML spec in the pangeo-data repository, and figured that the easy direct things that can be specified are:

1. limits
2. args

I added limits to https://github.com/yuvipanda/daskernetes/pull/26 but not args. I also added a much easier escape mechanism with deep-merging dictionaries. 

If this approach looks right to you, IMO next step is to talk to some folks who would be using these kinda clusters and figuring out what their ideal workflow looks like. I think there's multiple ideal workflows - full GUI management via a Jupyter extension, completely writing it via code, YAML files, etc. I think core should just be independent of that, and we can figure out which UX we should prioritize first.

Another important thing here is FUSE. Long term, I think giving workers this much permissions is a bad idea, but FlexVolume might not be good enough for you and you might need to wait for CSI to get out of alpha.

yuvipanda:
Anything leftover from here?

daskernetes/core.py
```diff
@@ -55,69 +82,96 @@ class KubeCluster(object):
...
         layout = ipywidgets.Layout(width='150px')
-        n_workers = ipywidgets.IntText(0, description='Workers', layout=layout)
-        threads = ipywidgets.IntText(1, description='Cores', layout=layout)
+
+        # Construct Bokeh dashboard link
+        # TODO: this currently assumes running within JupyterHub
+        # TODO: this code is entirely untested
+        link = '%sproxy/%d/status' % (os.environ['JUPYTERHUB_SERVICE_PREFIX'],
+                                      self.scheduler.services['bokeh'].port)
+        link = ipywidgets.HTML('<b>Dashboard:</b> <a href="%s" target="_blank">%s</a>' % (link, link))
```
mrocklin: We still need to figure out what to do here.

mrocklin: This is *very* useful in production, but it is also pretty specific.  I'm not sure yet the best way to make this externally configurable.

PR #26, opened 2018-01-23
Make KubeCluster more configurable

yuvipanda:
- Add config options for memory / cpu limits and guarantees
- Allow an easier escape hatch with `extra_container_config` and 
  `extra_pod_config`, which are dicts that are deep merged into the 
  pod / container configuration
- Minor other cleanups

### DIFF
--- a/daskernetes/core.py
+++ b/daskernetes/core.py
@@ -15,6 +15,29 @@ from kubernetes import client, config
 
 logger = logging.getLogger(__name__)
 
+def merge_dictionaries(a, b, path=None, update=True):
+    """
+    Merge two dictionaries recursively.
+
+    From https://stackoverflow.com/a/25270947
+    """
+    if path is None: path = []
+    for key in b:
+        if key in a:
+            if isinstance(a[key], dict) and isinstance(b[key], dict):
+                merge(a[key], b[key], path + [str(key)])
+            elif a[key] == b[key]:
+                pass # same leaf value
+            elif isinstance(a[key], list) and isinstance(b[key], list):
+                for idx, val in enumerate(b[key]):
+                    a[key][idx] = merge(a[key][idx], b[key][idx], path + [str(key), str(idx)], update=update)
+            elif update:
+                a[key] = b[key]
+            else:
+                raise Exception('Conflict at %s' % '.'.join(path + [str(key)]))
+        else:
+            a[key] = b[key]
+    return a
 
 class KubeCluster(object):
     """ Launch a Dask cluster on Kubernetes
@@ -64,6 +87,8 @@ class KubeCluster(object):
             host='0.0.0.0',
             port=8786,
             env={},
+            extra_container_config={},
+            extra_pod_config={},
             **kwargs,
     ):
         self.cluster = LocalCluster(ip=host or socket.gethostname(),
@@ -75,7 +100,7 @@ class KubeCluster(object):
         except config.ConfigException:
             config.load_kube_config()
 
-        self.api = client.CoreV1Api()
+        self.core_api = client.CoreV1Api()
 
         if namespace is None:
             namespace = _namespace_default()
@@ -89,6 +114,8 @@ class KubeCluster(object):
         self.worker_labels = (worker_labels or {}).copy()
         self.threads_per_worker = threads_per_worker
         self.env = dict(env)
+        self.extra_pod_config = extra_pod_config
+        self.extra_container_config = extra_container_config
 
         # Default labels that can't be overwritten
         self.worker_labels['org.pydata.dask/cluster-name'] = name
@@ -142,7 +169,7 @@ class KubeCluster(object):
         return self.scheduler.address
 
     def _make_pod(self):
-        return client.V1Pod(
+        pod = client.V1Pod(
             metadata=client.V1ObjectMeta(
                 generate_name=self.name + '-',
                 labels=self.worker_labels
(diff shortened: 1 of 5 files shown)


mrocklin:
I'm concerned that the current approach requires more configuration in the notebook than the average science user will be comfortable with.  For example the following code snippet is probably too much boiler plate for a science user to fully grok:

```python
   cluster = KubeCluster(
￼        loop=loop,
￼        n_workers=0,
￼        env={"TEST": "HI"},
￼        extra_container_config={
￼            "env": [ {"name": "BOO", "value": "FOO" } ],
￼            "args": ["last-item"]
￼        }
￼    )
￼```

And yet it might also be necessary for everyone in a certain group (I expect research groups to have similar specs like this).  

I'm still a fan of optionally placing the entire PodSpec in a yaml file if possible.  This conforms to a pre-existing specification (Kubernetes itself), and is easy for a group tech leader to configure and distribute to their group.

File: daskernetes/core.py [hunk 1]
```diff
@@ -165,24 +212,104 @@ def _make_pod(self):
...
+
+        for key, value in self.extra_pod_config.items():
+            self._set_k8s_attribute(
+                pod.spec,
+                key,
+                value
+            )
+        return pod
+
+    def _set_k8s_attribute(self, obj, attribute, value):
```
mrocklin: It looks like the only reference to self is to get out core_api.  Perhaps this should be a standalone function that accepts an api object?  That might also make it easier to test.

daskernetes/tests/test_objects.py
```diff
@@ -0,0 +1,108 @@
...
+            }
+        }
+    )
+
+    pod = cluster._make_pod()
+
+    assert pod.spec.containers[0].image_pull_policy == 'IfNotReady'
+    assert pod.spec.containers[0].security_context == {
+        'runAsUser': 0
+    }
```
mrocklin: I'm very glad to see this working well :)

mrocklin:
@rsignell-usgs @rabernat you two are probably early users for this.  The work here is central in allowing people to customize the environments of their dask clusters.  This is probably the sort of thing you would do to help support your research groups.

mrocklin:
We don't expect much of this to look familiar to you today, but it would be good to have your input on the kinds of interactions that you would be comfortable learning.

yuvipanda:
@mrocklin I don't disagree with you in any form or way :) I think it's only a question of factoring. We can easily build a YAML based object on top of this. I don't think we should use YAML in the core of the functionality we have here, because I think the UI for interacting with this should be made available in many ways:

1. GUI with a JupyterLab extension
2. YAML files that users manually write / admins write and put in images for them
3. Directly writing it in code

IMO the 'right' factoring is to keep the core agnostic to how users who don't want deep configurability might want to use it (since I strongly suspect that's going to be a JupyterLab UI rather than YAML! Everyone I've asked to write YAML eventually hates it, and we get lots of indent related errors), and then work on the various GUI aspects.

I think next step after this is to figure out both:

1. How a YAML spec should look like, how it should be distributed and loaded, etc
2. How a Jupyter UI for this should look like

Hope that makes sense.

daskernetes/core.py
yuvipanda: good idea, I'll do that!

I think the serialize function there could be a static or classmethod, I'll file a bug upstream.

rsignell-usgs:
As @yuvipanda suspected, I would like all of them, and likely would use the GUI with a JupyterLab extension the most.  But for sure others would prefer the YAML...

mrocklin:
> IMO the 'right' factoring is to keep the core agnostic to how users who don't want deep configurability might want to use it

Yeah, sorry, I wasn't trying to suggest that users *must* write YAML in order to configure their Pods, merely that we're able to pass through PodSpecs as are commonly defined in other established forms (like standard yaml spec, or the `kubernetes` Python API).  I become nervous when we start adding keywords to configure particular parts of the PodSpec, because this list invariably tends to grow and require tending ove time.

> How a YAML spec should look like, how it should be distributed and loaded, etc

My thought is that this would just be the same as someone would define a PodSpec in a normal Kubernetes file.  This way we know that our escape hatch is complete.

yuvipanda:
@mrocklin right, I think that is a very valid concern. So the extremes of that as I can see are:

1. Require an entire PodSpec template be passed in to constructor, and this template can come from YAML / Code. We can modify this to enforce some invariants (around labels, for example), but mostly not.
2. Have named constructor params for *everything*, escape hatch is 'make a PR!'

I think both these extremes are not ideal, and we should strike something of a middle ground - where the most basic setups are done by constructor arguments, hitting 60-80% of use cases, and then the rest has an escape hatch. This has worked well for us in kubespawner.

Do you think this is a valid framework to think about it?

If so, then the question becomes 'where is the line?'. IMO that line is fairly clear, and we're pretty much at it. We might want the ability to add annotations, but that should be it for the constructor. Everything else should go through the 'escape hatch'. You can pass in direct output of yaml.load to the escape hatch now and that should work...

It's good to write this down however, and if you think this makes sense I can document this in the comments of the class.

rabernat:
I'm trying to spin up on this to the point where I can give some feedback. But kubernetes is a new language to me.

I understand that this PR adds options for customization of dask workers. Does it also allow custom conda / pip packages to be installed? If so, how does that fit in to the various options discussed above?

yuvipanda:
@rabernat I think it's a little removed from there yet I think.

For me, the most useful thing to know would be 'what would your ideal workflow look like?'

yuvipanda:
@mrocklin upon more reflection, I wonder if:

> 1. Require an entire PodSpec template be passed in to constructor, and this template can come from YAML / Code. We can modify this to enforce some invariants (around labels, for example), but mostly not.

Is actually not a bad idea at all. We can provide convenience methods on top of it (essentially what we have now in the KubeCluster constructor), but the core can be kept simple.

mrocklin:
> Is actually not a bad idea at all. We can provide convenience methods on top of it (essentially what we have now in the KubeCluster constructor), but the core can be kept simple.

Having this as an option would make me happy.  Does this stop us from also accepting keywords that layer dictionaries on top of this base spec for particularly common cases?  Which keywords to accept becomes a complex question of course, I'm just curious if it's feasible to have the best of both worlds.

yuvipanda:
@mrocklin ok, I now have the core KubeCluster object only accepting a pod template, and there's a convenience function for generating said template. We can easily add a convenience function for generating it from yaml as well.

LMK how this looks! I got all tests to pass, although now you've to explicitly pass a '--worker-image' to pytest to tell it what image to use (since my local python is very different from the python used in dask upstream images).

I think easy ways to improve here would be:

1. Only image is required to make a podspec, we can read the image name from environment variable by default if it isn't given. We can also make other params default to reading from env variables, and populate this in our JupyterHub. This should address #28 
2. For a YAML loader, we can simply point to a path that has a YAML file. This file path can also default to a well known location, or be read from environment variable. This again can be set by admins.

yuvipanda:
I also wonder if I should just move the extra_config merging code to its own function, since then it can be chained with both yaml and non-yaml methods. We could also move the overrides back to KubeCluster constructor.

daskernetes/core.py
```diff
@@ -85,17 +84,15 @@ def __init__(
...
-        self.threads_per_worker = threads_per_worker
-        self.env = dict(env)
 
         # Default labels that can't be overwritten
-        self.worker_labels['org.pydata.dask/cluster-name'] = name
-        self.worker_labels['app'] = 'dask'
-        self.worker_labels['component'] = 'dask-worker'
+        worker_pod_template.metadata.labels['dask.pydata.org/cluster-name'] = name
+        worker_pod_template.metadata.labels['app'] = 'dask'
+        worker_pod_template.metadata.labels['component'] = 'dask-worker'
```
mrocklin: Should we do a copy before mutating the given input?

Also, should we expect a Pod or a PodSpec?  Do we want the user to specify a an ObjMeta?

daskernetes/core.py
```diff
@@ -85,17 +84,15 @@ def __init__(
 
         self.namespace = namespace
         self.name = name
-        self.worker_image = worker_image
-        self.worker_labels = (worker_labels or {}).copy()
-        self.threads_per_worker = threads_per_worker
-        self.env = dict(env)
```
mrocklin: If we want to we can continue supporting these kinds of common configurations by modifying the template in the `_make_pod` method.  I don't know if this is desired, but it might be useful if there are parameters that we expect users to want to change?

daskernetes/core.py
```diff
@@ -55,27 +52,29 @@ class KubeCluster(object):
     """
     def __init__(
             self,
+            worker_pod_template,
```
mrocklin: It might be nice to accept this as either a `kubernetes` library object, a dictionary, or a filename, and then normalize all options into a kubernetes object.

daskernetes/objects.py
```diff
@@ -0,0 +1,167 @@
...
+        raise ImportError("PyYaml is required to use yaml functionality, please install!")
+
+    apiclient = client.ApiClient()
+    with open(filepath) as f:
+        # FIXME: We can't use the 'deserialize' function since
+        # that expects a response object!
+        pod_template = apiclient.deserialize(
+            _FakeResponse(yaml.safe_load(f)),
+            client.V1Pod
+        )
```
mrocklin: I wasn't able to make this approach work before.  A number of components of the kubernetes object were silently dropped.  I had a bunch of hacks on top of this to get my use case to work.  Those hacks were pretty fragile.  https://github.com/mrocklin/daskernetes/blob/deploy/daskernetes/core.py#L338-L374

mrocklin: This particular problem was my main blocker.  I wasn't familiar enough with the `kubernetes` library to easily find a path out.  Most other issues I'm able to handle

daskernetes/core.py
yuvipanda: Yeah, I think we should accept Pod objects than PodSpec, since that's what all the k8s docs generally specify. And having users be able to do annotations / labels is also quite useful.

daskernetes/core.py
yuvipanda: I made individual classmethods for these variants instead. I'm generally not a fan of doing polymorphism style stuff in Python constructors, since I thinks this is clearer.

daskernetes/core.py
yuvipanda: I think I'm going to move the extra_* stuff to the constructor, and just have a 'common configuration' thing as another classmethod.

mrocklin:
Ah, I found the source of my `snake_case` issue.  `pod.to_dict()` returns a dictionaries with `snake_case` rather than `camelCase`.  I had wrongly assumed that this would provide dictionaries that we could round-trip with `deserialize`

PR #30, opened 2018-02-02
Add failing test for make_pod_from_dict

mrocklin:
This includes failures in the deserialize method that we saw when
deploying pangeo.pydata.org

### DIFF
--- a/daskernetes/tests/test_objects.py
+++ b/daskernetes/tests/test_objects.py
@@ -3,7 +3,7 @@ from time import sleep
 import yaml
 import tempfile
 from daskernetes import KubeCluster
-from daskernetes.objects import make_pod_spec
+from daskernetes.objects import make_pod_spec, make_pod_from_dict
 from distributed.utils_test import loop, inc
 from dask.distributed import Client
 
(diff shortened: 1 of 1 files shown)


daskernetes/tests/test_objects.py
```diff
@@ -129,3 +129,36 @@ def test_extra_container_config_merge(image_name, loop):
...
+            {
+                "args": [
+                    "dask-worker",
+                    "$(DASK_SCHEDULER_ADDRESS)",
+                    "--nthreads",
+                    "1"
+                ],
+                "image": "image-name",
+                "name": "dask-worker",
+                "security_context": {"capabilities": {"add": ["SYS_ADMIN"]},
```
yuvipanda: This needs to be securityContext, since we're asking for dicts to be in the same format as what Kubernetes expects (those are in camelCase, not _ case).

yuvipanda: I'm going to spend time fixing and documenting this properly.

mrocklin:
Ah ha, that would explain things.   Thanks!

PR #31, opened 2018-02-05
Update pod-merging branch

mrocklin:
A variety of small changes.

### DIFF
--- a/daskernetes/objects.py
+++ b/daskernetes/objects.py
@@ -1,4 +1,12 @@
+"""
+Convenience functions for creating pod templates.
+"""
 from kubernetes import client
+import json
+try:
+    import yaml
+except ImportError:
+    yaml = False
 
 def _set_k8s_attribute(obj, attribute, value):
     """
@@ -84,6 +92,9 @@ def make_pod_spec(
         cpu_limit=None,
         cpu_request=None,
 ):
+    """
+    Create a pod template from various parameters passed in.
+    """
     pod = client.V1Pod(
         metadata=client.V1ObjectMeta(
             labels=labels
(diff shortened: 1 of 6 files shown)


daskernetes/objects.py
```diff
@@ -86,20 +86,28 @@ def merge_dictionaries(a, b, path=None, update=True):
...
         extra_container_config={},
         extra_pod_config={},
-        memory_limit=None,
-        memory_request=None,
-        cpu_limit=None,
-        cpu_request=None,
+        memory_limit='4GB',
+        memory_request='4GB',
+        cpu_limit=1,
+        cpu_request=1,
```
mrocklin: Is it normal to provide resource defaults, or is it more typical to let these default to `None`?  I'm inclined to add limits, at least for memory.  Also, what are valid values here?  `G`, `GB`, `GiB` ?

daskernetes/tests/conftest.py
```diff
@@ -8,8 +8,5 @@ def pytest_addoption(parser):
 
 @pytest.fixture
 def image_name(request):
-    worker_image = request.config.getoption('--worker-image')
-    if not worker_image:
-        pytest.fail("Need to pass --worker-image. Image must have same python setup as host!")
-        return
+    worker_image = request.config.getoption('--worker-image') or 'daskdev/dask:latest'
```
mrocklin: Any objection with defaulting to `daskdev/dask`?

daskernetes/objects.py
yuvipanda: yeah, it's normal to default those to None.

daskernetes/tests/conftest.py
yuvipanda: In general, I think what we should do is:

1. Default those to None
2. Read from env variables as secondary default if they are None

I don't think we should use daskdev/dask as default, because:

1. It defaults to 'daskdev/dask:latest', and using latest anywhere is a ticking time bomb
2. It also will only work for people on the same version of python from conda.

I agree that users not having to know about images is a requirement for good UX. My suggestion is:

1. Default from something like DASKERNETES_CONTAINER_IMAGE
2. In JupyterHub deployment, set that to the same value as singleuser.image 

This would mean you have a single source of truth for image, and it won't drift out of use

mrocklin:
Pushed manually.  Closing.

PR #33, opened 2018-02-07
Continuous Integration, and Pytest Fixtures

mrocklin:
This PR has two major changes:

1.  It introduces continuous integration testing on Circle-CI 
2.  It changes the test suite to use pytest fixtures.  In particular these fixtures run tests within a new namespace and clean up that namespace on terminating the test.

cc @yuvipanda if you have time to review

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,8 +50,8 @@ jobs:
       - run:
           name: configure conda
           command: |
+              export PATH="$HOME/miniconda/bin:$PATH"
               if [ ! -d "/home/circleci/miniconda/envs/daskernetes-test" ]; then
-                export PATH="$HOME/miniconda/bin:$PATH"
                 conda update -q conda
                 conda env create -f ci/environment-${PYTHON}.yml --name=${ENV_NAME}
                 source activate ${ENV_NAME}
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,6 +35,13 @@ jobs:
           command: |
             sudo kubectl get deployment
       - run:
+          name: install miniconda
+          command: |
+              wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
+              bash miniconda.sh -b -p $HOME/miniconda
+              export PATH="$HOME/miniconda/bin:$PATH"
+              conda config --set always_yes yes --set changeps1 no
+      - run:
           name: configure conda
           command: ./ci/install-circle.sh
       - run:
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,60 @@
...
+      - run:
+          name: configure conda
+          command: |
+              export PATH="$HOME/miniconda/bin:$PATH"
+              conda update -q conda
+              conda env create -f ci/environment-${PYTHON}.yml --name=${ENV_NAME}
+              conda env list
+              source activate ${ENV_NAME}
+              pip install --no-deps --quiet -e .
+              conda list ${ENV_NAME}
```
mrocklin: TODO: we should figure out how to cache this.  @TomAugspurger does this in the dask-ml circle-ci config.

daskernetes/core.py
```diff
@@ -228,10 +229,23 @@ def scale_up(self, n, **kwargs):
...
+                ]
+                break
+            except client.rest.ApiException as e:
+                if 'ServerTimeout' in str(e):
+                    logger.info("Server timeout, retry #%d", i + 1)
+                    time.sleep(1)
+                    last_exception = e
+                    continue
+                else:
+                    raise
```
mrocklin: I had odd API failures previously that I associated to running under a very recently created namespace.  This *may* have fixed things, but I'm not very confident here.  cc @yuvipanda

.circleci/config.yml
mrocklin: I've now added caching.  This replaces around one minute of conda time with twenty seconds of download-cache time.

mrocklin:
Merging this soon if there are no further comments

daskernetes/core.py
```diff
@@ -228,10 +229,23 @@ def scale_up(self, n, **kwargs):
...
-        ]
+        for i in range(3):
+            try:
+                out = [
+                    self.core_api.create_namespaced_pod(self.namespace, self.pod_template)
+                    for _ in range(n - len(pods))
+                ]
+                break
+            except client.rest.ApiException as e:
+                if 'ServerTimeout' in str(e):
```
yuvipanda: There's probably a better way to check for this one, let me look.

yuvipanda: I think checking for e.status = 500 and e.reason = 'ServerTimeout' is a better check here, but otherwise this looks good.

daskernetes/core.py
yuvipanda: Yeah, namespaces are 'initializing' for a short while, and reject objects inside it. This should be fine.

requirements.txt
```diff
@@ -0,0 +1,2 @@
+dask[distributed]
+kubernetes
```
yuvipanda: We should pin this to 4.*, since they often have breaking changes.

PR #41, opened 2018-02-10
Automatically read worker yaml file from env variable

mrocklin:
This allows users to launch a KubeCluster() with no arguments if the
worker pod spec is in a yaml file pointed to by
the DASKERNETES_WORKER_SPEC_PATH environment variable.

This commit also creates a config.py file that contains daskernetes
configuration, which is currently just taken from environment variables.

### DIFF
--- a/daskernetes/config.py
+++ b/daskernetes/config.py
@@ -1,0 +1,5 @@
+import os
+
+config = {'-'.join(k.split('_')[1:]).lower(): v
+          for k, v in os.environ.items()
+          if k.startswith('DASKERNETES_')}
--- a/daskernetes/config.py
+++ b/daskernetes/config.py
@@ -1,3 +1,17 @@
+"""
+We capture all environment variables named
+
+    DASKERNETES_FOO_BAR=value
+
+And turn them into key-value pairs in a global ``config`` dictionary.
+We change the keys to remove ``DASKERNETES_``, replace underscores with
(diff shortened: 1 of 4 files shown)


daskernetes/tests/test_core.py
```diff
@@ -283,3 +282,36 @@ def test_scale_up_down(cluster, client):
...
+                "image": image_name,
+                "name": "dask-worker"
+            }]
+        }
+    }
+
+    with tmpfile(extension='yaml') as fn:
+        with open(fn, mode='w') as f:
+            yaml.dump(test_yaml, f)
+        config['worker-spec-path'] = fn
```
mrocklin: Can we think of a more concise-and-descriptive name here?

daskernetes/config.py
```diff
@@ -0,0 +1,5 @@
+import os
+
+config = {'-'.join(k.split('_')[1:]).lower(): v
```
yuvipanda: Add a docstring to this file?

daskernetes/tests/test_core.py
yuvipanda: I like worker-template-path better, but is not more concise. Am ok with whatever you pick

daskernetes/config.py
mrocklin: Done

yuvipanda:
Feel free to merge after tests!

PR #46, opened 2018-02-20
Allow customisation of worker name in environment

jacobtomlinson:
This PR allows you to customise the worker name template with the environment variable `DASKERNETES_WORKER_NAME_TEMPLATE`. If not specified it falls back to `dask-{user}-{uuid}`.

The template is then formatted with `user` and `uuid` as before but also it unpacks `os.environ` to allow the use of environment variables in the template.

For example in my use case I'm keen to configure the template to `dask-{JUPYTERHUB_USER}-{uuid}` as the `user` is always `jovyan`.

### DIFF
--- a/daskernetes/core.py
+++ b/daskernetes/core.py
@@ -158,8 +158,8 @@ class KubeCluster(object):
             namespace = _namespace_default()
 
         if name is None:
-            worker_name_template = config.get('worker-name-template', 'dask-{user}-{uuid}')
-            name =  worker_name_template.format(user=getpass.getuser(), uuid=str(uuid.uuid4())[:10], **os.environ)
+            worker_name = config.get('worker-name', 'dask-{user}-{uuid}')
+            name =  worker_name.format(user=getpass.getuser(), uuid=str(uuid.uuid4())[:10], **os.environ)
 
         self.pod_template = clean_pod_template(pod_template)
         # Default labels that can't be overwritten
--- a/daskernetes/core.py
+++ b/daskernetes/core.py
@@ -159,7 +159,7 @@ class KubeCluster(object):
 
         if name is None:
             worker_name = config.get('worker-name', 'dask-{user}-{uuid}')
-            name =  worker_name.format(user=getpass.getuser(), uuid=str(uuid.uuid4())[:10], **os.environ)
+            name = worker_name.format(user=getpass.getuser(), uuid=str(uuid.uuid4())[:10], **os.environ)
 
         self.pod_template = clean_pod_template(pod_template)
         # Default labels that can't be overwritten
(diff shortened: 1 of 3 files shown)


mrocklin:
This looks good to me.  I recommend a small test similar to this one:

```python
def test_diagnostics_link_env_variable(pod_spec, loop, ns):
    pytest.importorskip('bokeh')
    config['diagnostics-link'] = 'foo-{USER}-{port}'
    try:
        with KubeCluster(pod_spec, loop=loop, namespace=ns) as cluster:
            port = cluster.scheduler.services['bokeh'].port
            cluster._ipython_display_()
            box = cluster._cached_widget

            link = box.children[0]
            assert 'foo-' + getpass.getuser() + '-' + str(port) in link.value
            assert 'href' in link.value
    finally:
        del config['diagnostics-link']
```

If we start doing this with many config variables then we might consider adopting the single-line:

    worker_name_template = config.get('worker-name-template', 'dask-{user}-{uuid}')

Additionally we should keep a list of these options somewhere.  Currently that list is maintained here: http://daskernetes.readthedocs.io/en/latest/#configuration

jacobtomlinson:
Thanks for the quick review. 

Yeah I read the code back after submitting and thought "why didn't I just use `.get()`?.

I'll get the test added and update the docs.

mrocklin:
This looks good to me.  Final question: do we want to use `DASKERNETES_WORKER_NAME` instead of `DASKERNETES_WORKER_NAME_TEMPLATE`?  I believe that the use of template in `DASKERNETES_WORKER_TEMPLATE` before referred to the kubernetes specification.  If it's safe to shorten the name here that might be nice.  I don't have strong feelings here though.

mrocklin:
I'm looking into the test failures.  They seem to be unrelated.

jacobtomlinson:
Happy to shorten! 

I was being explicit that it is a template and therefore you can use the `{variable}` notation. However I see `DASKERNETES_DIAGNOSTICS_LINK` is the same and doesn't feel the need to say it is a template. I'll drop the suffix and push again.

Yeah didn't think it was me, looked to be timeouts on the main tests.

jacobtomlinson:
@mrocklin do you need me to do any more to this? Not sure if there's anything I can do to get the tests passing.

mrocklin:
Nope, you're good.  I'm working on tests now.  I'll merge this after that goes in.  I'll explicitly ping you if anything else needs to happen.  Thanks for working on this.

jacobtomlinson:
Thanks! No probs!

mrocklin:
Merging.  Thanks @jacobtomlinson !

jacobtomlinson:
🎉

Thanks for removing the extra space 😂.

PR #57, opened 2018-03-31
typo

BrendanMartin:


mrocklin:
Thanks @BrendanMartin !

BrendanMartin:
@mrocklin would you be able to help me out with something in regards to dask-kubernetes?

mrocklin:
That would depend on what it is.  I recommend raising a public issue here on GitHub.  That way others can help out as well if it turns out that your problem is more in line with their expertise.

BrendanMartin:
@mrocklin it's not technically an issue with the library, it's more like I can't figure out how to actually connect to Google Kubernetes via dask-kubernetes. Is that something I should open an issue on?

mrocklin:
It's probably the cleanest way today to ask a question of developers here.
If its sufficiently off topic then we'll just close it.  It's definitely
preferable to contacting any particular developer directly.  You might also
try asking questions on Stack Overflow.  Perhaps there is a tag for the
Google service that you're trying to use there as well?

PR #60, opened 2018-04-02
Default worker template path in /etc

ogrisel:
I find it convenient to look up for a default location for the default worker template. That makes it possible to skip one environment variable in the helm config by default.

This PR also fixes old references to `DASKERNETES_` env variables in the docstring of `KubeCluster`.

### DIFF
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -20,6 +20,8 @@ from .objects import make_pod_from_dict, clean_pod_template
 
 logger = logging.getLogger(__name__)
 
+DEFAULT_WORKER_TEMPLATE_PATH = "/etc/dask/kubernetes/worker-template.yaml"
+
 
 class KubeCluster(Cluster):
     """ Launch a Dask cluster on Kubernetes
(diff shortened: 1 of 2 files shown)


dask_kubernetes/core.py
```diff
@@ -20,6 +20,8 @@
 
 logger = logging.getLogger(__name__)
 
+DEFAULT_WORKER_TEMPLATE_PATH = "/etc/dask/kubernetes/worker-template.yaml"
```
mrocklin: We could also use `~/.dask/kubernetes/...` which is where we currently place our config files.  I think that this directory is configurable by setting the `DASK_CONFIG` environment variable (we should probably document this better).

ogrisel: The common way would be to lookup in that order:

- `/etc/dask/kubernetes/worker-template.yaml`
- `~/.dask/kubernetes/worker-template.yaml`
- `DASK_KUBERNETES_WORKER_TEMPLATE_PATH`

and take the last file that exists.

mrocklin: I'm fine with adopting that convention.  Should we eventually adopt this more broadly throughout the project?

ogrisel: Which other configuration files do you have in mind? You mean dask in general or just dask-kubernetes?

ogrisel: BTW, nowadays I think that using `~/.config/dask/` would be better than `~/.dask/` but if we switch we should be careful about backward dealing with compat.

mrocklin: I was referring to dask in general.

Hrm, looking at my own `~/.config` directory I now agree.  I'll raise an issue upstream.

ogrisel:
This is now discussed for the dask project as a whole at: https://github.com/dask/dask/issues/3367

jacobtomlinson:
It's been a while since this last had activity and dask/dask#3367 has now been closed.

Did we reach a consensus on how to go about this? @ogrisel are you happy to smooth out the conflicts?

mrocklin:
I believe that this is now automatically handled by configuration.  If you put a config file in `/etc/dask/whatever.yaml` then Dask will find it and include it in the configuration.  Closing.

PR #67, opened 2018-04-10
Do not pull daskdev/dask image for each test

ogrisel:
The following PR tries to use `'imagePullPolicy': 'IfNotPresent'` as much as possible in the pod specs used in the tests (only those that actually spawn worker pods) so as to accelerate the speed of the tests (both on the CI and on the local machine).

### DIFF
--- a/dask_kubernetes/tests/test_core.py
+++ b/dask_kubernetes/tests/test_core.py
@@ -34,7 +34,10 @@ def ns():
 
 @pytest.fixture
 def pod_spec(image_name):
-    yield make_pod_spec(image=image_name)
+    yield make_pod_spec(
+        image=image_name,
+        extra_container_config={'imagePullPolicy': 'IfNotPresent'}
+    )
 
 
 @pytest.fixture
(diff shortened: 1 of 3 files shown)


ogrisel:
Apparently, it does not seem to change the tests execution speed on Circle CI. However I think it is useful not to try to pull an image from the repository to make it possible to run `docker build -t somelocalimagetag .` (after `eval $(minikube docker-env)`) without pushing to a docker image repo and then running the dask-kubernetes tests only against that locally built and tagged docker image.

ogrisel:
I made a mistake in this PR. I am working on a fix.

ogrisel:
Ok I can run the tests with a private / local image with this PR.

mrocklin:
Thanks @ogrisel !  Sorry for the slow response.

ogrisel:
No problem, thanks for the merge. #69 is more interesting but minikube seems to be broken on circle ci and I cannot find the cause of the problem by googling.

PR #69, opened 2018-04-11
Delete pending pods first when manually scaling down

ogrisel:
When scaling down a cluster with pending pods still being provisioned in the background, one can often get an `AssertionError` in  `select_workers_to_close`. Furthermore, running workers are deleted before pending pods have a chance to get up and running when using a saturated shared kubernetes cluster. This trashes intermediate computation results for no valid reason and makes everything much slower than necessary.

This PR fixes all of that by always trying to scale down non-running pods first. Furthermore, we now explicitly scale down running pods by first scheduling `self.scheduler.retire_workers` in the event loop to make sure that we have the time to migrate intermediate results back to the remaining running workers.

I have added (stress) tests for all the above.

`test_scale_down_pending` can be quite slow when running on a cluster that allows for 20+ pods. #67 can accelerate by 30% the execution of that test (from 30s to 20s on my laptop).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,24 +4,29 @@ jobs:
   build:
     machine: true
     environment:
+      MINIKUBE_WANTUPDATENOTIFICATION: false
+      MINIKUBE_WANTREPORTERRORPROMPT: false
       CHANGE_MINIKUBE_NONE_USER: true
-    environment:
       PYTHON: "3.6"
       ENV_NAME: "dask-kubernetes-test"
+
     steps:
       - checkout
       - run:
           command: |
             curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/v1.8.4/bin/linux/amd64/kubectl && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
+      # Use minikube v0.25.2 as v0.26.0 requires systemd to manage the daemons which
+      # is not available in the Ubuntu 14.04 image used by Circle CI with the
+      # "machine: true" option.
       - run:
           command: |
-            curl -Lo minikube https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64 && chmod +x minikube && sudo mv minikube /usr/local/bin/
+            curl -Lo minikube https://github.com/kubernetes/minikube/releases/download/v0.25.2/minikube-linux-amd64 && chmod +x minikube && sudo mv minikube /usr/local/bin/
       - run:
           command: |
-            sudo minikube start --vm-driver=none
+            sudo -E minikube start --vm-driver=none
       - run:
           command: |
-            sudo minikube update-context
+            sudo -E minikube update-context
       - run:
           command: |
             JSONPATH='{range .items[*]}{@.metadata.name}:{range @.status.conditions[*]}{@.type}={@.status};{end}{end}'; until sudo kubectl get nodes -o jsonpath="$JSONPATH" 2>&1 | grep -q "Ready=True"; do sleep 1; done
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,7 @@ jobs:
             curl -Lo minikube https://github.com/kubernetes/minikube/releases/download/v0.25.2/minikube-linux-amd64 && chmod +x minikube && sudo mv minikube /usr/local/bin/
       - run:
           command: |
-            sudo -E minikube start --vm-driver=none
+            sudo -E minikube start --vm-driver=none --extra-config=kubelet.MaxPods=20
       - run:
           command: |
             sudo -E minikube update-context
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -312,69 +312,45 @@ class KubeCluster(Cluster):
         KubeCluster.scale_up
         KubeCluster.scale_down
         """
-        pods = self.pods()
+        pods = self._cleanup_succeeded_pods(self.pods())
         if n >= len(pods):
             return self.scale_up(n, pods=pods)
         else:
-            to_close = select_workers_to_close(self.scheduler, len(pods) - n)
+            n_to_delete = len(pods) - n
+            # Before trying to close running workers, check if we can cancel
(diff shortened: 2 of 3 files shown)


ogrisel:
Hum the minikube setup is broken on circle ci. I have tried with `--kubernetes=v1.9.4` but we still get error: E0411 when starting the kubelet daemon...

ogrisel:
@mrocklin @yuvipanda ok I found the source of the problem with circleci: the new minikube release (v0.26.0) relies on `systemd` when using `--vm-driver=none`. Systemd is not available on the ubuntu 14.04 image of circle ci used when `machine: true` is enabled. Therefore I decided to fallback to minikube v0.25.2 for now.

All the tests pass including the new ones. Total build runtime is now 3min 15s instead of 2min 30s as previously (because the new tests are a bit expensive as they create many containers). I still think it's worth it to run some scaling stress tests under heavy kubernetes API load.

dask_kubernetes/core.py
```diff
@@ -312,13 +312,62 @@ def scale(self, n):
...
+            if len(to_close) < len(self.scheduler.workers):
+                # Close workers cleanly to migrate any temporary results to
+                # remaining workers.
+                future = self.scheduler.retire_workers(
+                        workers=to_close, remove=True, close_workers=True)
+
+                def callback(future):
+                    self.scale_down(to_close)
+
+                self.scheduler.loop.add_future(future, callback)
```
ogrisel: @mrocklin I have very little experience with tornado / async programming, so I would appreciate your feedback on this pattern.

.circleci/config.yml
```diff
@@ -4,24 +4,29 @@ jobs:
...
       ENV_NAME: "dask-kubernetes-test"
+
     steps:
       - checkout
       - run:
           command: |
             curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/v1.8.4/bin/linux/amd64/kubectl && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
+      # Use minikube v0.25.2 as v0.26.0 requires systemd to manage the daemons which
+      # is not available in the Ubuntu 14.04 image used by Circle CI with the
+      # "machine: true" option.
```
mrocklin: Explicitly pinning the version sounds like a good idea regardless

dask_kubernetes/core.py
mrocklin: The `scale` method might be called by the user in the IPython interaction thread which is often different from the thread where the IOLoop is running.  As a result we want to avoid calling any coroutine like `retire_workers` directly.  Instead, we need to use the threadsafe method `IOLoop.add_callback`.  I suggest the following:

```python
@gen.coroutine
def f(to_close):
    yield self.scheduler.retire_workers(workers=to_close, ...)
    self.scale_down(to_close)

self.scheduler.loop.add_callback(f, to_close)
```

However this might become problematic if `scale_down` blocks (does it?  I'm not that knowledgable about the kubernetes API).   If so then we might want to offload this to a thread in a thread pool.

```python
from distributed.comm.utils import offload

@gen.coroutine
def f(to_close):
    yield self.scheduler.retire_workers(workers=to_close, ...)
    yield offload(self.scale_down, to_close)

self.scheduler.loop.add_callback(f, to_close)
```

dask_kubernetes/core.py
```diff
@@ -440,15 +482,16 @@ def _namespace_default():
...
     key = lambda ws: ws.info['memory']
-    to_close = set(sorted(s.idle, key=key)[:n])
+    to_close = set(sorted(scheduler.idle, key=key)[:n_to_close])
 
-    if len(to_close) < n:
-        rest = sorted(s.workers.values(), key=key, reverse=True)
-        while len(to_close) < n:
+    if len(to_close) < n_to_close:
+        rest = sorted(workers, key=key, reverse=True)
+        while len(to_close) < n_to_close:
```
mrocklin: Eventually we might want to leverage `Scheduler.workers_to_close`. It won't take into account succeeded pods but may be useful just to centralize logic.  

This doesn't need to be now though, I just wanted to increase awareness that this logic also existed in the core library.

dask_kubernetes/tests/test_core.py
```diff
@@ -313,6 +314,94 @@ def test_scale_up_down(cluster, client):
...
+        sleep(0.1)
+        assert time() < start + 10
+
+    # Put some data on this worker
+    future = client.submit(lambda: b'\x00' * int(1e6))
+    wait(future)
+
+    # Rescale the cluster many times without waiting: this should put some
+    # pressure on kubernetes but this should never fail nor delete our worker
+    # with the temporary result.
```
mrocklin: If you wanted to verify that we never delete the worker with the temporary result then we should probably verify that the following is the same before and after

    cluster.scheduler.tasks[future.key].who_has

ogrisel:
> However this might become problematic if scale_down blocks (does it? I'm not that knowledgable about the kubernetes API). If so then we might want to offload this to a thread in a thread pool.

It's not an async call but it does not wait for kubernetes to actually delete the pods to return. It's just making a transaction with kubernetes API to schedule the pod deletion on its side as far as I understand. Maybe @yuvipanda knows better.

Still offloading looks simple to implement so we can use it to make it more cooperative with the event loop in case of slow interaction with the kubernetes API.

dask_kubernetes/tests/test_core.py
ogrisel: Thanks, I will do that.

dask_kubernetes/core.py
ogrisel: Yes I saw that and was wondering why you did not use it in the first place. I wanted this PR to only fix the original problem but indeed it might be a good idea to centralize logics in a following PR.

ogrisel:
@mrocklin I think I addressed your comments in my last commit.

mrocklin:
This looks good to me.  Thank you @ogrisel .  Merging.

ogrisel:
Great! Have a nice WE, it's beer time in France. See you next week ;)

mrocklin:
Enjoy!

PR #70, opened 2018-04-15
Explicitly pass Scheduler to KubeCluster

mrocklin:
This can be useful when adding adaptive scheduling to an existing cluster.

### DIFF
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -15,6 +15,7 @@ except ImportError:
 from tornado import gen
 from distributed.deploy import LocalCluster, Cluster
 from distributed.config import config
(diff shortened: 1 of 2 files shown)


mrocklin:
My intention here is to start towards enabling adaptivity on the helm chart by adding KubeCluster to a running scheduler.  Unfortunately I'm blocked right now on an issue with my local testing environment (see #71 ) If anyone has any suggestions they would be welcome.

mrocklin:
cc @jacobtomlinson

jacobtomlinson:
Given that things have shifted quite a bit with recent changes I'm going to close out this PR as stale. But I have raised #185 to note that this is desired functionality.

PR #75, opened 2018-05-05
[RFC] Configuration

mrocklin:
We're updating configuration in https://github.com/dask/dask/pull/3432

This affects downstream projects and makes it a bit easier for them to handle their own configuration files.  

This PR shows how that configuration system might be used within dask-kubernetes.  Review would be welcome.  I recommend reading the documentation provided within https://github.com/dask/dask/pull/3432 first.

cc @jacobtomlinson


mrocklin:
I've removed the RFC label.  This is hypothetically ready to merge.  Critical feedback welcome.

dask_kubernetes/core.py
```diff
@@ -381,6 +392,8 @@ def scale_up(self, n, pods=None, **kwargs):
         --------
         >>> cluster.scale_up(20)  # ask for twenty workers
         """
+        if dask.config.get('kubernetes.count.max') is not None:
+            n = min(n, dask.config.get('kubernetes.count.max'))
```
jhamman: Would it be worth issuing a warning when `n > kubernetes.count.max`? This may lead to surprising behavior when you can't scale beyond a certain threshold because of a parameter you didn't set.

PR #81, opened 2018-06-25
Include username in metadata

mrocklin:
I'm a little curious to know if there are good reasons against this.


jacobtomlinson:
I don't see why not!

mrocklin:
Great.  Merging.  Thanks @jacobtomlinson !

PR #87, opened 2018-07-20
Add flag to block until scaling finishes

jacobtomlinson:
Sometimes I've found that I want to create a cluster but wait for it to scale to the desired size before starting to execute work on it.

This PR adds a `wait` flag which allows users to tell the `KubeCluster.scale()` method to block until scaling completes.

### DIFF
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -333,7 +333,7 @@ class KubeCluster(Cluster):
         return self.core_api.read_namespaced_pod_log(pod.metadata.name,
                                                      pod.metadata.namespace)
 
-    def scale(self, n):
+    def scale(self, n, wait=False):
         """ Scale cluster to n workers
 
         Parameters
@@ -341,9 +341,13 @@ class KubeCluster(Cluster):
         n: int
             Target number of workers
 
+        wait: bool
+            Flag to wait for the scaling to happen before returning
+
         Example
         -------
         >>> cluster.scale(10)  # scale cluster to ten workers
+        >>> cluster.scale(10, wait=True)  # scale and block until workers are connected
 
         See Also
         --------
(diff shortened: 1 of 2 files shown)


mrocklin:
I agree that this would be useful.  I'd like to bring up a few thoughts:

1.  This seems like the sort of thing that might be useful across all of the cluster objects, and so maybe it should go into dask/distributed instead
2.  Explicitly calling `time.sleep` will stop asynchronous operation (like tornado/asyncio/async-await), which is useful for advanced applications, and also very useful for testing.   We might have to be a bit more clever here
3.  One solution here might be to use [SchedulerPlugins](http://distributed.readthedocs.io/en/latest/plugins.html) to explicitly trigger events whenever a new worker reaches the scheduler.  Calling `wait` on the cluster object might register a `tornado.locks.Event` or `tornado.locks.Condition` that gets triggered when the right number of workers have arrived.  
4.  @jhamman implemented a SchedulerPlugin for dask-jobqueue in order to improve adaptive operation (he wanted adaptive to be aware of pending jobs).  This work happened in https://github.com/dask/dask-jobqueue/pull/63 and might be the sort of thing we could think about generalizing.  Any thoughts on this @jhamman?

jhamman:
IIRC, `Cluster.scale()` is executed inside the event loop. Coincidentally, I'm thinking the lack of the wait option may be partially to blame for my failing tests in https://github.com/dask/dask-jobqueue/pull/97. 

The one thing the `SchedulerPlugin` does not offer at the moment is a way to track pending jobs so that is currently handled in dask-jobqueue.

Question, does the Cluster superclass have any pluggins attached by default? This sort of functionality seems to belong best there but its not clear to me if the pluggin route is the best way to achieve that.

mrocklin:
> Question, does the Cluster superclass have any pluggins attached by default? This sort of functionality seems to belong best there but its not clear to me if the pluggin route is the best way to achieve that.

Not today, but it could.  There could be a standard plugin that downstream clusters would use.  Presumably they would report whenever they started a worker and then the plugin would place that information into a pending structure similar to how dask-jobqueue does now.  We would implement a `wait` method on the Cluster object that would block until all jobs currently in the pending queue of the plugin had been removed.

mrocklin:
Also cc @jcrist for dask-yarn (though I think he's away this week)

jacobtomlinson:
Closing in favour of dask/distributed#2138

PR #89, opened 2018-07-28
Fix broken worker-template configuration

fbergroth:
I believe #75 broke reading pod template from `kubernetes.worker-template`.

If `kubernetes.worker-template` is set, `pod_template` becomes a truthy value (by the line I removed), and the following code will never run:
https://github.com/dask/dask-kubernetes/blob/c4728ba4ecf87c175f3b2020f3ffb0a91860b6d0/dask_kubernetes/core.py#L159-L161

Later on, since `make_pod_from_dict` was never called, the following code will raise an exception since `pod_template` at this point is still a `dict`.
https://github.com/dask/dask-kubernetes/blob/c4728ba4ecf87c175f3b2020f3ffb0a91860b6d0/dask_kubernetes/core.py#L193

### DIFF
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -148,7 +148,6 @@ class KubeCluster(Cluster):
             env=None,
             **kwargs
     ):
-        pod_template = pod_template or dask.config.get('kubernetes.worker-template')
         name = name or dask.config.get('kubernetes.name')
         namespace = namespace or dask.config.get('kubernetes.namespace')
         n_workers = n_workers if n_workers is not None else dask.config.get('kubernetes.count.start')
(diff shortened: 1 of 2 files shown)


fbergroth:
I'm not very familiar yet with this library, but your suggestion seems reasonable. Do you want me to add a test as well?

mrocklin:
It would be nice to add a test, yes.  My hope is that my example above,
along with the existing tests in this library should be sufficient to guide
you to a correct test.

On Sat, Jul 28, 2018 at 12:33 PM, Fredrik Bergroth <notifications@github.com

fbergroth:
Alright, I added a simple test for this.

mrocklin:
Looks good.  Thanks @fbergroth !  Merging

Issue #90, opened 2018-08-03
Scaling up only works initially 


fbergroth:
I'm trying to do some batch jobs using dask-kubernetes. Initially i run:
```
cluster.adapt(minimum=1, maximum=6)
```

Which works fine, and I soon have 6 worker-pods running.

However, when a pod dies, another one isn't recreated, instead I get flooded with these messages (about every second):
```
2018-08-03 08:24:50 [INFO] distributed.deploy.adaptive: CPU limit exceeded [132485 occupancy / 5 cores]
2018-08-03 08:24:50 [INFO] distributed.deploy.adaptive: Scaling up to 6 workers
```
Any idea what's going on?

jacobtomlinson:
Interesting! When a worker dies for me I tend to get a `KilledWorker` error and the whole thing exits.

Are you able to provide some example code so I could have a go at reproducing this?

fbergroth:
Previously I just used `scale_up(6)`, and also saw `KilledWorker`, the future errored, but I could continue executing.

This time my worker pod got evicted, and dask-kubernetes fails to replace it.

I'm not quite sure how to reproduce it, I might have some time next week, but any help would be appreciated.

fbergroth:
Okay, so it seems that the cluster doesn't have enough resources to spawn another pod. It should be checked in `scale_up` I guess, and the following note seems related to this as well:

https://github.com/dask/dask-kubernetes/blob/e8e919aa60f267ceb3986bd019d598730bd1a8cd/dask_kubernetes/core.py#L444

fbergroth:
I think I found the issue, see #91

PR #91, opened 2018-08-05
Exclude failed pods from list of pods

fbergroth:
Digging a bit deeper into #90, I think I found the issue.

So I started 6 workers, one of them got evicted, and then my logs got flooded with
```
2018-08-03 08:24:50 [INFO] distributed.deploy.adaptive: CPU limit exceeded [132485 occupancy / 5 cores]
2018-08-03 08:24:50 [INFO] distributed.deploy.adaptive: Scaling up to 6 workers
```
This is the status of my evicted pod: 
```
"status": {
    "message": "The node was low on resource: memory.",
    "phase": "Failed",
    "reason": "Evicted",
    "startTime": "2018-08-05T11:29:55Z"
}
```
It seems to be that failed pods are not excluded in the podlist, only Successful pods are. So `to_create` in `scale_up` ends up being 0.

The simple fix is the one attached, but maybe more changes should be made.
- Should we cleanup failed pods?
- Should `_cleanup_succeeded_pods` be renamed?
- Do you have some other fix in mind?

### DIFF
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -395,7 +395,12 @@ class KubeCluster(Cluster):
                     self.namespace,
                     kubernetes.client.V1DeleteOptions()
                 )
-                logger.info('Deleted pod: %s', pod.metadata.name)
+                pod_info = pod.metadata.name
+                if pod.status.reason:
+                    pod_info += ' [{}]'.format(pod.status.reason)
+                if pod.status.message:
+                    pod_info += ' {}'.format(pod.status.message)
+                logger.info('Deleted pod: %s', pod_info)
             except kubernetes.client.rest.ApiException as e:
                 # If a pod has already been removed, just ignore the error
                 if e.status != 404:
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -349,7 +349,7 @@ class KubeCluster(Cluster):
         KubeCluster.scale_up
         KubeCluster.scale_down
         """
-        pods = self._cleanup_succeeded_pods(self.pods())
+        pods = self._cleanup_terminated_pods(self.pods())
         if n >= len(pods):
             return self.scale_up(n, pods=pods)
         else:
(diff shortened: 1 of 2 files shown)


dask_kubernetes/core.py
```diff
@@ -404,7 +404,7 @@ def _delete_pods(self, to_delete):
     def _cleanup_succeeded_pods(self, pods):
         terminated_pods = [p for p in pods if p.status.phase == 'Succeeded']
         self._delete_pods(terminated_pods)
-        return [p for p in pods if p.status.phase != 'Succeeded']
+        return [p for p in pods if p.status.phase not in {'Succeeded', 'Failed'}]
```
jacobtomlinson: I think this should be a list `[]` rather than a dictionary `{}`.

fbergroth: This is a set literal, which I think is more idiomatic since we check for inclusion and don't care about order.

jacobtomlinson: Ah nice I had not seen that syntax before. That makes sense then!

jacobtomlinson:
Many thanks for your contribution!

fbergroth:
Nice, thanks.

@jacobtomlinson any chance you can cut a new release with my latest contributions?

jacobtomlinson:
@fbergroth I haven't cut a release of this repo yet. But now seems like a good time to do one.

@mrocklin I assume you have done the previous releases. What is the process for this repo?

jacobtomlinson:
Ok great that's useful thanks. I wondered if there would be a CI/CD pipeline for this, but that explains why I couldn't find one.

My pypi username is `jacob.tomlinson`.

mrocklin:
Added

jacobtomlinson:
Thanks

fbergroth:
@jacobtomlinson friendly ping :-)

jacobtomlinson:
@fbergroth apologies I'm traveling at the moment and don't want to cut a release in case something goes wrong and I'm not around to fix it.

@mrocklin any chance you can do this?

PR #101, opened 2018-09-28
expand environment variables in pod specs

### LINKED ISSUE #92: Using environment variables to update values in worker-spec.yml
In the dask configuration module, it is possible to use environment variables to update configuration values. It doesn't seem that is currently possible here as the "worker-spec.yml" file is actually a kubernetes spec, rather than a dask config file. I'm curious if there is a way to do that now or if it would be feasible to add such functionality. 

somewhat related to #88, #45

Background for my particular use case: https://github.com/jupyterhub/kubespawner/pull/193, https://github.com/jupyter/repo2docker/pull/363

jhamman:
closes #92

I've taken the minimal distruption path here. Only environment variables in the pod spec are expanded. Two other ways this could be done include:

1. expand environment variables in the config.py module. This would only work for variables set in the dask configuration files, not for the worker spec yaml. This would open up the option of expanding environment variables other than the ones in the pod spec. I don't need this at the moment but it may be worth considering.
1. in the `make_pod_from_dict` function (in objects.py). This would further tie the behavior to just expanding variables to the pod spec, and nothing else.

cc @jacobtomlinson and @mrocklin 

xref: https://github.com/dask/dask/pull/3893

### DIFF
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -157,6 +157,7 @@ class KubeCluster(Cluster):
 
         if not pod_template and dask.config.get('kubernetes.worker-template', None):
             d = dask.config.get('kubernetes.worker-template')
+            d = dask.config.expand_environment_variables(d)
             pod_template = make_pod_from_dict(d)
 
         if not pod_template and dask.config.get('kubernetes.worker-template-path', None):
@@ -165,6 +166,7 @@ class KubeCluster(Cluster):
             fn = fn.format(**os.environ)
             with open(fn) as f:
                 d = yaml.safe_load(f)
+            d = dask.config.expand_environment_variables(d)
             pod_template = make_pod_from_dict(d)
 
         if not pod_template:
@@ -275,6 +277,7 @@ class KubeCluster(Cluster):
             raise ImportError("PyYaml is required to use yaml functionality, please install it!")
         with open(yaml_path) as f:
             d = yaml.safe_load(f)
+            d = dask.config.expand_environment_variables(d)
             return cls.from_dict(d, **kwargs)
 
     @property
--- a/dask_kubernetes/tests/test_core.py
+++ b/dask_kubernetes/tests/test_core.py
@@ -204,6 +204,42 @@ def test_pod_from_yaml(image_name, loop, ns):
                 assert all(client.has_what().values())
 
 
+def test_pod_from_yaml_expand_env_vars(image_name, loop, ns):
+    try:
+        os.environ["FOO_IMAGE"] = image_name
+
+        test_yaml = {
+            "kind": "Pod",
+            "metadata": {
+                "labels": {
+                    "app": "dask",
+                    "component": "dask-worker"
+                }
+            },
+            "spec": {
+                "containers": [{
+                    "args": [
+                        "dask-worker",
+                        "$(DASK_SCHEDULER_ADDRESS)",
+                        "--nthreads",
+                        "1"
+                    ],
+                    "image": '${FOO_IMAGE}',
+                    'imagePullPolicy': 'IfNotPresent',
+                    "name": "dask-worker"
+                }]
+            }
+        }
+
+        with tmpfile(extension='yaml') as fn:
+            with open(fn, mode='w') as f:
+                yaml.dump(test_yaml, f)
+            with KubeCluster.from_yaml(f.name, loop=loop, namespace=ns) as cluster:
+                assert cluster.pod_template.spec.containers[0].image == image_name
+    finally:
+        del os.environ['FOO_IMAGE']
+
+
 def test_pod_from_dict(image_name, loop, ns):
     spec = {
         'metadata': {},


ci/environment-3.6.yml
```diff
@@ -4,13 +4,11 @@ channels:
 dependencies:
   - flake8
   - ipywidgets
-  - dask=0.17.4
-  - distributed=1.21.8
+  - dask
+  - distributed
```
mrocklin: I would prefer to keep these explicitly pinned if possible, just so it's clear what we're testing against.  I think that we should bump the versions to 0.19.2 and 1.23.2, the latest versions as of today.

jhamman:
I'm getting a test failure in `test_reject_evicted_workers`. I dare say I didn't break this one!

mrocklin:
That's an odd failure.  @jacobtomlinson do you have any understanding of why this might fail?  This is more a kubernetes issue than a Dask issue.  I'm reproducing the relevant portion of the test here for convenience:

```python
def test_reject_evicted_workers(cluster):
    cluster.scale(1)

    start = time()
    while len(cluster.scheduler.workers) != 1:
        sleep(0.1)
        assert time() < start + 60

    # Evict worker
    [worker] = cluster.pods()
    cluster.core_api.create_namespaced_pod_eviction(
        worker.metadata.name,
        worker.metadata.namespace,
        kubernetes.client.V1beta1Eviction(
            delete_options=kubernetes.client.V1DeleteOptions(grace_period_seconds=300),
            metadata=worker.metadata))

    # Wait until pod is evicted
    start = time()
    while cluster.pods()[0].status.phase == 'Running':
        sleep(0.1)
        assert time() < start + 60

    [worker] = cluster.pods()
    assert worker.status.phase == 'Failed'
```

Under what circumstances would this pod be pending at this point?

mrocklin:
I've peeled off the version changes from this PR into https://github.com/dask/dask-kubernetes/pull/104 .  Let see if it fails as well.

mrocklin:
I've force pushed to your branch.  I hope you don't mind.

mrocklin:
Stil getting this error: 

```
_________________________ test_reject_evicted_workers __________________________

cluster = KubeCluster("tcp://10.142.7.162:39295", workers=1)

    def test_reject_evicted_workers(cluster):
        cluster.scale(1)
    
        start = time()
        while len(cluster.scheduler.workers) != 1:
            sleep(0.1)
            assert time() < start + 60
    
        # Evict worker
        [worker] = cluster.pods()
        cluster.core_api.create_namespaced_pod_eviction(
            worker.metadata.name,
            worker.metadata.namespace,
            kubernetes.client.V1beta1Eviction(
                delete_options=kubernetes.client.V1DeleteOptions(grace_period_seconds=300),
                metadata=worker.metadata))
    
        # Wait until pod is evicted
        start = time()
        while cluster.pods()[0].status.phase == 'Running':
            sleep(0.1)
            assert time() < start + 60
    
        [worker] = cluster.pods()
>       assert worker.status.phase == 'Failed'
E       AssertionError: assert 'Pending' == 'Failed'
E         - Pending
E         + Failed

dask_kubernetes/tests/test_core.py:359: AssertionError
```

Restarting to see if it is intermittent

jhamman:
@mrocklin - Passed on a second go. Interesting.

mrocklin:
I'm now rebuilding the previous PR at
https://circleci.com/gh/dask/dask-kubernetes/168

mrocklin:
OK, I've run the CI in this PR and the CI on master about ten times now.  This PR failed with the error above a couple of those times while master hasn't failed yet with that error.   

Master did just fail with the following exception (which seems a bit different)

```

=================================== FAILURES ===================================
___________________________ test_scale_down_pending ____________________________

cluster = KubeCluster("tcp://10.142.5.118:38827", workers=3)
client = <Client: scheduler='tcp://10.142.5.118:38827' processes=3 cores=3>

    def test_scale_down_pending(cluster, client):
        # Try to scale the cluster to use more pods than available
        nodes = cluster.core_api.list_node().items
        max_pods = sum(int(node.status.allocatable['pods']) for node in nodes)
        if max_pods > 50:
            # It's probably not reasonable to run this test against a large
            # kubernetes cluster.
            pytest.skip("Require a small test kubernetes cluster (maxpod <= 50)")
        extra_pods = 5
        requested_pods = max_pods + extra_pods
        cluster.scale(requested_pods)
    
        start = time()
        while len(cluster.scheduler.workers) < 2:
            sleep(0.1)
            # Wait a bit because the kubernetes cluster can take time to provision
            # the requested pods as we requested a large number of pods.
            assert time() < start + 60
    
        pending_pods = [p for p in cluster.pods() if p.status.phase == 'Pending']
        assert len(pending_pods) >= extra_pods
    
        running_workers = list(cluster.scheduler.workers.keys())
        assert len(running_workers) >= 2
    
        # Put some data on those workers to make them important to keep as long
        # as possible.
        def load_data(i):
            return b'\x00' * (i * int(1e6))
    
        futures = [client.submit(load_data, i, workers=w)
                   for i, w in enumerate(running_workers)]
        wait(futures)
    
        # Reduce the cluster size down to the actually useful nodes: pending pods
        # and running pods without results should be shutdown and removed first:
        cluster.scale(len(running_workers))
    
        start = time()
        pod_statuses = [p.status.phase for p in cluster.pods()]
        while len(pod_statuses) != len(running_workers):
            if time() - start > 60:
                raise AssertionError("Expected %d running pods but got %r"
>                                    % (len(running_workers), pod_statuses))
E               AssertionError: Expected 2 running pods but got ['Running', 'Running', 'Running']

dask_kubernetes/tests/test_core.py:445: AssertionError
```

mrocklin:
Sorry for the long delay with this @jhamman 

Lets merge it in and see what happens

PR #104, opened 2018-10-17
Update base requirements

mrocklin:


mrocklin:
@jhamman am I missing anything here?

jhamman:
@mrocklin - this looks right. Perhaps the failure in #101 was a fluke. I'd be happy to see this go in before I return to the environment variable issue.

mrocklin:
> Perhaps the failure in #101 was a fluke

Yeah, that wouldn't be surprising.  At some point we should run these tests several times to try to flush out intermittent failures.

PR #105, opened 2018-10-25
Bump dask versions in CI environment

mrocklin:


mrocklin:
Ah, this is still using the 0.19.2 docker image

PR #106, opened 2018-10-26
Update index.rst

TomAugspurger:


TomAugspurger:
Right now, the "Deploy > Kubernetes" dropdown on `dask.org` points to the docs for this repo. We may want to change that link to http://docs.dask.org/en/latest/setup/kubernetes.html, which will link to the helm docs and to here.

I think having the two projects cross-reference each other is also a good idea.

mrocklin:
Sounds good to me.  Merging.

PR #110, opened 2018-11-15
Feat/default toleration

### LINKED ISSUE #109: Add default toleration to worker pod
I have setup an instance of dask-kubernetes that uses taints to prevent `kube-system` components from scheduling onto the nodes. This really helps when using Google's autoscaler, since empty nodes are easily batch deleted.

The configuration that I needed to add is here:
https://github.com/pangeo-data/atmos.pangeo.io-deploy/blob/0ddab66e1a37f9f2a9ac85c8fdad7301b9e3882d/deployments/atmos.pangeo.io/image/.dask/config.yaml#L36-L44
```
      tolerations:
      - key: "k8s.dask.org_dedicated"
        operator: "Equal"
        value: "worker"
        effect: "NoSchedule"
      - key: "k8s.dask.org/dedicated"
        operator: "Equal"
        value: "worker"
        effect: "NoSchedule"
```

It might be nice to add a default toleration, so we standardize what taints are recommended for using dask-kubernetes.

dsludwig:
Fixes #109 

Adds the default tolerations:

`k8s.dask.org/dedicated=worker` & `k8s.dask.org_dedicated=worker`. These can be used to create nodes that will only schedule dask workers.

### DIFF
--- a/dask_kubernetes/objects.py
+++ b/dask_kubernetes/objects.py
@@ -134,6 +134,22 @@ def make_pod_spec(
                     env=[client.V1EnvVar(name=k, value=v)
                             for k, v in env.items()],
                 )
+            ],
+            tolerations=[
+                client.V1Toleration(
+                    key='k8s.dask.org/dedicated',
+                    operator='Equal',
+                    value='worker',
+                    effect='NoSchedule',
+                ),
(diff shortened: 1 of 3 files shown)


mrocklin:
cc @jacobtomlinson or @yuvipanda for review

Issue #84, opened 2018-07-09
Spawn Dask scheduler on separate Kubernetes pod


chrish42:
This feature request is a followup to #82. Having at least the option to spawn the Dask scheduler on a remote Kubernetes pod would enable more use cases for Dask-kubernetes, including at least the "run Jupyter on my notebook, but do the Dask computation on a remote cluster" one.

mrocklin:
I agree that this would be useful.  It also seems doable to me, but would
require moderate effort.

I'm a bit curious how often people have access to a Kubernetes cluster when
they're not in the cluster themselves.

jacobtomlinson:
This is definitely something we would use in anger. 

Running Jupyter on a laptop and then connecting to a cloud based cluster for computation would be really useful.

mrocklin:
@jacobtomlinson is this something that is important enough to you all to
help develop?  I imagine that some parts of this, like making adaptive work
across a network, I'll probably have to do, but I'm curious if there are
parts for which I can lean on others.  My guess is that the first pass of
just putting the scheduler in another pod is probably something that you
are as if not more capable of than I am :)

jacobtomlinson:
@mrocklin I will have to check my schedule and see if I can dedicate time to this.

chrish42:
@mrocklin From my perspective, everyone is 5 minutes (and a credit card) away from having access to a Kubernetes cluster when they outgrow their laptop for computations. Especially if #85 is also implemented. Then they don't need to install anything else but dask-kubernetes; they just instantiate a `dask_kubernetes.Cluster` object, passing it the cluster credentials given when they created the cluster in the GCP (or other) console. It's getting to a JupyterHub-like setup on Kubernetes, where people can launch notebooks, etc. that requires more work (and/or money), and has potentially a smaller audience.

jgerardsimcock:
We would really like to explore the idea of having there be a central scheduler that users via jupyter notebook/terminal could log into and run large batch jobs. I'd like to understand how to do something like the financial modeling team in the [dask use cases](http://dask.pydata.org/en/latest/use-cases.html)

What is the best place to start on getting this moving?

mturok:
+1

arpit1997:
Is it still open? Can I try this?

@jgerardsimcock 
if we spawn dask scheduler in another pod and let scheduler manage jobs wouldn't that be enough? Am I missing something here?

[...]

PR #111, opened 2018-11-16
Feat/cluster auth

### LINKED ISSUE #85: Add ability to explictly specify Kubernetes cluster
Currently, `KubeCluster()` is hardcoded to launch Dask workers in the "current" Kubernetes cluster, this being either the cluster where the current process is running ("in-cluster config") or as a fall-back the default cluster in the kubeconfig file. As this may not be the right strategy for all use cases, it would be nice to be able to be able to optionally specify more explicitly which cluster to use. Here are the choices I currently see possible:

- in-cluster
- default from kubeconfig
- non-default cluster (other context) from kubeconfig
- other cluster (URL to Kubernetes API server + TLS certificate for connection... is anything else needed?)

dsludwig:
Fixes #85 

Adds three classes that implement a single `ClusterAuth` interface:

* `InCluster` - load configuration from container environment
* `KubeConfig` - load configuration from a kubeconf file
* `KubeAuth` - explicitly provide configuration

Currently, the configuration is globally set. The interface could instead instantiate an API client object with the appropriate configuration, which would be less messy. The only reason I left it global is for `_cleanup_pods`, which instantiates its own client object. I guess the finalizer could take an `auth` argument, which would allow it to instantiate a client in the same way as `KubeCluster`.

### DIFF
--- a/dask_kubernetes/__init__.py
+++ b/dask_kubernetes/__init__.py
@@ -1,4 +1,5 @@
 from . import config
+from .auth import ClusterAuth, KubeAuth, KubeConfig, InCluster
 from .core import KubeCluster
 from .objects import make_pod_spec, make_pod_from_dict
 
--- a/dask_kubernetes/auth.py
+++ b/dask_kubernetes/auth.py
@@ -1,0 +1,157 @@
+"""
+Defines different methods to configure a connection to a Kubernetes cluster.
+"""
+import logging
+
+import kubernetes
+
+logger = logging.getLogger(__name__)
+
+
+class ClusterAuth(object):
+    """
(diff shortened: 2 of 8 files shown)


dsludwig:
I think this becomes more useful if #84 is implemented. I found it useful just to be able to explicitly load a particular kubeconfig file/context for my own testing, but I think you're right - in general, the networking would not be setup for workers to connect back to a local scheduler.

mrocklin:
cc @jacobtomlinson for review if he has time (probably not until next week,
given timezones)

mrocklin:
I just noticed that this never got merged.  Should it be merged?

PR #114, opened 2018-12-20
Add Python 2.7 compatibility

rbubley:


mrocklin:
Interesting!

This raises the fascinating question of "are we willing to accept contributions to maintain Python 2 compatibility?"

@jacobtomlinson @yuvipanda you two seem to do most of the maintenance here.  What is your preference?

My loose preference is to accept this, trigger a release, and then not promise to maintain Python 2.7 support.  Groups stuck on Python 2 would then still have *some* solution, even if it was old.

yuvipanda:
Thanks for making this effort, @rbubley. I appreciate folks who do the hard, unglamorous work of trying to keep backwards compatibility to reach maximum number of users.

That said, I personally have a somewhat strong belief that python 2.x support should not be added to any new libraries that don't already have it. There are extenuating circumstances - places where a script needs to run without being able to rely on any external libraries, for example. But for something like dask-kubernetes, where you already need kubernetes, and this most likely runs in a container, I don't think we should support python2.x.

I'm also exploring https://github.com/tomplus/kubernetes_asyncio as a way forward for all kubernetes / python API interactions, and any libraries that are asyncio native are often 3.x only.

rbubley:
I should probably provide the motivation for this PR.

I concur that we want everyone to migrate to Python 3, of course, but some client/worker code has not yet been migrated from Python 2.7.

Dask/Distributed did discuss enabling the scenario where client/worker were on 2.7 and scheduler was on 3, but I think this was not accepted (see discussion at https://github.com/dask/distributed/pull/989); the implication being that if you need clients/workers on 2.7 you need the scheduler on 2.7, too.

In the context of a dask kubernetes implementation that leads us to this PR.

jacobtomlinson:
Seconding the thanks for the effort, however I also agree that generally we should not be adding support for Python 2.7. 

While I understand what @mrocklin is saying about accepting but not agreeing to maintain I feel like that may set a bad precedent.

I would be interested to know what the main dask and distributed projects are doing in terms of Python 2.7 support and their roadmap.

hugovk:
For what it's worth, here's the pip installs for Dask Kubernetes from PyPI for December 2018:

| category | percent | downloads |
|----------|--------:|----------:|
|      3.6 |  98.00% |     8,832 |
|      3.7 |   0.80% |        72 |
| null     |   0.74% |        67 |
|      2.7 |   0.37% |        33 |
|      3.5 |   0.08% |         7 |
|      3.4 |   0.01% |         1 |
| Total    |         |     9,012 |

Source: `pypistats python_minor dask-kubernetes --last-month`

Looking at the chart at https://pypistats.org/packages/dask-kubernetes, I've no idea what caused that huge spike of Python 3.6 downloads in December!

image

So here's more representative November stats:

| category | percent | downloads |
|----------|--------:|----------:|
|      3.6 |  84.31% |       881 |
| null     |   8.71% |        91 |
|      3.7 |   4.59% |        48 |
|      2.7 |   1.63% |        17 |
|      3.5 |   0.77% |         8 |
| Total    |         |     1,045 |

Source: `pypistats python_minor dask-kubernetes -m 2018-11`

And October's are similar:

| category | percent | downloads |
|----------|--------:|----------:|
|      3.6 |  82.65% |       781 |
| null     |  12.06% |       114 |
|      2.7 |   2.01% |        19 |
|      3.7 |   1.69% |        16 |
|      3.5 |   1.27% |        12 |
|      3.4 |   0.32% |         3 |
| Total    |         |       945 |

Source: `pypistats python_minor dask-kubernetes -m 2018-10`

jacobtomlinson:
> Can you expand on this?

We are fighting a battle within our organisation to get everyone off Python 2 before the EOL at the end of the year. Adding Python 2 support to a project which has not previously had Python 2 support opens the door to new requests of that nature. 

As this project is still very young there will likely be many bug fixes over time and core features implemented. As kubernetes moves forwards quickly we will aim to keep pace. This would result in us quickly leaving our Python 2 friends behind.

>  If some large organization comes to Dask and wants to try it out but uses Python 2 today we currently say "nope, you have to switch over your entire architecture before you can try things out" which is unlikely to occur.

This is compelling, I'm just keen to be mindful of what I said before. Perhaps a good stance would be to accept PRs from the community adding Python 2 support (like this one) but intentionally not encouraging the core maintainers to not worry about breaking the support.

mrocklin:
> Perhaps a good stance would be to accept PRs from the community adding Python 2 support (like this one) but intentionally not encouraging the core maintainers to not worry about breaking the support.

Yes, sorry, that's what I meant to propose earlier.  I'm inclined to merge this, cut a release immediately, and then never touch Python 2 again (especially after @yuvipanda uses an async kubernetes library).  I'm happy to do this work so that no one else has to deal with it.

mrocklin:
@yuvipanda any objection to my doing this?

yuvipanda:
@mrockline no objections. +1 on cutting release and back again

yuvipanda:
@mrockline no objections. +1 on cutting release and back again

mrocklin:
OK.  This is done.  https://github.com/dask/dask-kubernetes/commit/458da55c2e44a15d9147dfbb9c2a80bf67598bc7 

This was released as 0.7

mrocklin:
Dropping Python 2 in https://github.com/dask/dask-kubernetes/pull/117

PR #115, opened 2018-12-22
Pass through return value from close

mrocklin:
This enables dask_kubernetes to close clusters asynchronously

    await cluster.close()

This follows on to https://github.com/dask/distributed/pull/2437


mrocklin:
Yeah, I think it's fine regardless.

Issue #121, opened 2019-02-06
Kubernetes Scheduling Performance Example


rmccorm4:
Hi,

I'm new to `dask` and `dask-kubernetes` and I'm trying to put together an example that shows a performance increase from using `dask-kubernetes` with multiple pods/workers by performing computations in parallel. Do you have anything like this?

This was my attempt, but it's hard to tell if the pods are even being given a chunk of the work to do. 

```
from dask_kubernetes import KubeCluster

cluster = KubeCluster.from_yaml('worker-spec.yml')
cluster.adapt(minimum=1, maximum=100)  # or dynamically scale based on current workload

import sys
import dask
import numpy as np
import dask.array as da
N = 1_000_000

def compute_sum(num_arrays=10):
    mat = np.arange(N)
    mat = da.from_array(mat, chunks=-1)

    results = []
    for i in range(num_arrays):
        # Delay computations to be computed in parallel at the end
        total = dask.delayed(sum)(mat)
        # Store computations in a list to be computed in parallel
        results.append(total)

    return results

if __name__ == '__main__':
    if len(sys.argv) != 3:
        sys.exit('Usage: python launch_pods.py <NUM_ARRAYS> <NUM_WORKERS>')

    num_arrays = int(sys.argv[1])
    num_workers = int(sys.argv[2])
    cluster.scale_up(num_workers)
    computation_list = compute_sum(num_arrays=num_arrays)
    totals = dask.compute(*computation_list)
    print(totals)
    # Validate answer
    print((N-1)*N/2)
```

**Additionally, I couldn't get more than 1 pod to be created from using `cluster.adapt()`. Could you show an example or a workload that would cause `cluster.adapt()` to make the scheduler spawn multiple pods?**

P.S. I'm using the default worker-spec.yml and running `minikube` 

```
# worker-spec.yml

kind: Pod
metadata:
  labels:
    foo: bar
spec:
  restartPolicy: Never
  containers:
  - image: daskdev/dask:latest
    imagePullPolicy: IfNotPresent
    args: [dask-worker, --nthreads, '2', --no-bokeh, --memory-limit, 6GB, --death-timeout, '60']
    name: dask
    env:
      - name: EXTRA_PIP_PACKAGES
        value: fastparquet git+https://github.com/dask/distributed
    resources:
      limits:
        cpu: "2"
        memory: 6G
      requests:
        cpu: "2"
        memory: 6G
```

rmccorm4:
[...]

PR #124, opened 2019-02-11
Add usage example to quickstart

jacobtomlinson:
In response to dask/dask-kubernetes#121 I've added a quick start example for using the cluster.


mrocklin:
Tests fail here, but those failures are in master.  I ran into one of them when doing the async work (now merged into `dev`) and the other is from using an old version of the `distributed` library.  Merging.

PR #125, opened 2019-02-11
Update core.py

bphi:
`pod_template` should be `V1Pod` not `V1PodSpec`


yuvipanda:
Thanks for the doc fix, @bphi!

Issue #123, opened 2019-02-08
Make scaling up and down more robust


mrocklin:
As a reminder, Dask users scale clusters using a Cluster API with the scale
method.

```python
cluster.scale(10)  # set the cluster to ten workers total
```

Currently in Dask-kubernetes we implement this API with the following two methods

```python
async def _scale_up(self, n: int):
    ...

async def _scale_down(self, workers: List[str])
    ...
```

However, our logic in these methods isn't foolproof, and we can get into some
bad situations if the user scales up and down rapidly without waiting for
thigns to settle.  This case happens somewhat frequently, especially when using
Dask's [adaptive scaling](http://docs.dask.org/en/latest/setup/adaptive.html) which calls scale based on the current activity
and perceived need in the scheduler.

There is some effort in `scale_up` and `scale_down` to be clever around:

-  Removing terminated pods
-  Removing pods that are Pending before removing actual workers
-  Accounting for Pending as well as Running pods when we scale up

However our current tests show that we're not very successful in ensuring that
we don't run into a bad situation.  In particular if you look through the test
suite for tests named `test_scale_*` that ar marked with `@pytest.mark.xfail`
you'll see some failures where combinations of calls of scaling up and down
result in a state where the user has asked for a certain number of workers and
we fail to deliver that to them.

Ideally calling `scale` would pin a target number of workers and we would
continuously endeavor to meet this target.  In principle Kubernetes could
handle this for us, but we want to have manual control over pods so that we can
scale them down mindfully of current results.  Dask already has functions to
determine which workers to kill given a target number, and move data off of
them, look for `select_workers_to_close` and `retire_workers`.

*Note: This became a more pressing issue when we switched to using async/await style code (I'm not sure why).  Near term work here should target the `dev` branch.*

JoeJasinski:
Hi, I'm new to this project, so I don't know all of the ins and outs yet. If I understand correctly, pods are individually spawned and scaled based on the yaml template (or other means of specification). 

However, I was wondering if it would make sense to use a kubernetes Deployment instead of individual pods to scale up and down the cluster. If a deployment is used, then kubernetes itself can re-spawn pods that die instead of having to implement methods to manually do this housekeeping. 

I realize that there is probably a reason for this choice to use individual pods, but was just wondering if using deployments had been considered. Or maybe I'm misunderstanding how it works...

jacobtomlinson:
[...]

PR #132, opened 2019-03-22
[WIP] Fea/scale w Async Queue

quasiben:
resolves https://github.com/dask/dask-kubernetes/issues/123

This PR currently implements a timeout with assertion that pods are in a `Running` state.  This is executed 3 times in try loop when scaling up clusters.  It's a very naive implementation and the timeout is hardcoded (this could an optional setting).  

We can improve this check with exponential backoff (this is how k8s works when asking for a resource) or implement a state-machine or internal queueing system to more closely evaluate state.  

More commits to come


dask_kubernetes/core.py
```diff
@@ -508,6 +508,35 @@ def f(to_close):
     def scale_up(self, n, **kwargs):
         self.scheduler.loop.add_callback(self._scale_up, n, **kwargs)
 
+    async def _pod_status(self, pods):
```
mrocklin:
```suggestion
   def _pod_status(self, pods):
```

This doesn't appear to need to be asynchronous

mrocklin: Actually, this function only appears to be used once, and is relatively simple.  I recommend that we just inline this logic into the list comprehension where it is used.

dask_kubernetes/core.py
```diff
@@ -508,6 +508,35 @@ def f(to_close):
...
+        for p in pods:
+            status[p.metadata.name] = p.status.phase
+        return status
+
+    async def check_status(self):
+        while True:
+            pods = await self._pods()
+            p_status = await self._pod_status(pods)
+            result = ['Running' == status for name, status in p_status.items()]
+            await asyncio.sleep(3.2)
```
mrocklin: This seems like an odd place for a sleep, and also an odd value.  

I would expect to sleep only if we didn't pass everything, and I would also hope that the iteration time would be smaller than 3s.

dask_kubernetes/tests/test_async.py
```diff
@@ -477,9 +477,7 @@ def pod_spec(image_name):
     assert set(cluster.scheduler.workers) == {b}
 
 
-@pytest.mark.xfail(
-    reason="The delay between scaling up, starting a worker, and then scale down causes issues"
-)
```
mrocklin: Can I ask you to also see if this resolves `test_scale_down_pending`?

dask_kubernetes/core.py
```diff
@@ -508,6 +508,35 @@ def f(to_close):
...
+            if not all(result):
+                continue
```
mrocklin: What if we scale down after scaling up within the timeout above?  Some of the pods might have a status of "done" or "cancelled" or something, right?

mrocklin:
Another approach, which I'm not sure would scale well, would be for `scale` to store its most recent value on the Cluster object.

```python
>>> cluster.scale(20)
>>> cluster._manual_scale_target
20
```

And then we could periodically check the status of the pods, and make any changes if things had gotten out of alignment.  So if the last user request was for 20 pods and for some reason we only had 19, then we would ask Kubernetes for another.  If the user had asked for zero workers, and later we find out that surprisingly one has stuck around, then we would just try to delete it again.  This lets us be a bit sloppy about tracking individual requests, since we'll always try to bring things back in a dumb way to the user's defined goal.

At first I was concerned that this might get expensive with many pods (1000s is common enough) so it would be hard to do this check every second or so, but it looks like this all goes into one request, so maybe it isn't that bad?

Thoughts?

quasiben:
I believe what you've outlined is similar to the state machine idea -- periodically check the request of and move towards matching request and status.  The periodic check can be tunable but I would expect at least a 5 second frequency?  When I've used/seen this used in the past it was in the context of a deployed thing being functional in <15 seconds.  Is that too slow ?  

Also, how do we check if a worker is stuck ?  Or do you mean the pod status changes from 'Running' to something else ?

While we discuss I'll build another PR which implements this idea

quasiben:
In this PR I am going to assume that we want to handle adaptive scaling with low latencies.  This means being resilient to number of worker changes in a ~3 second window. 

- 0 s: 5 workers
- 3 s:  10 workers
- 6 s: 2 workers
...

In last commit (3d59f3d434f0fe75cd67a32aa31f687b7a0e3f8b), I've cheated a bit by checking the total number of workers vs the target -- abusing the suggestion of `_manual_scale_target`.

```python
result = [status_check == 'Running' for name, s in status.items()]
if sum(result) < self._manual_scale_target:
    continue
else:
    # we have met the target
    return True
````

By doing the above, we can be speedy about changes to the number of workers.  Updates to `_manual_scale_target` can be independent of our check of number of running workers.  Additionally, we are not bound by a periodic check -- the above is triggered when whenever we scale up.  As for scaling down, the code still simply calls delete and returns.  

The logic of scaling up is also simplified in that we try for 3 times to meet a target and timeout in each of those tries after `status-timeout` (default 5 seconds)

**Comment by quasiben** [comment]
fixing other tests now as well

File: dask_kubernetes/core.py [hunk 5]
```diff
@@ -500,27 +514,58 @@ def f(to_close):
...
 
-    async def _scale_up(self, n, pods=None, **kwargs):
+    async def check_status(self, status_check=PodStatus.running):
+        while True:
+            pods = await self._pods()
+            phases = {p.metadata.name: p.status.phase for p in pods}
+            result = [status_check.value == status for name, status in phases.items()]
+
+            # throttle loop
+            await asyncio.sleep(0.5)
```
mrocklin: Ideally if things go smoothly the first time it would be good to not have any unnecessary slowdown.  This 500ms delay is probably better placed at the end of a loop like this, hopefully after a potential break/return.

dask_kubernetes/core.py
```diff
@@ -614,7 +664,7 @@ def _cleanup_pods_sync(namespace, labels):
     for pod in pods.items:
         try:
             api.delete_namespaced_pod(
-                pod.metadata.name, namespace, kubernetes.client.V1DeleteOptions()
+                name=pod.metadata.name, namespace=namespace, body=kubernetes.client.V1DeleteOptions()
```
mrocklin: +1 on explicit keywords, thanks

dask_kubernetes/core.py
```diff
@@ -500,27 +514,58 @@ def f(to_close):
...
+            if result is None:
+                continue
+            elif sum(result) < self._manual_scale_target:
+                continue
+            else:
+                return True
+
+    async def status_with_timeout(self, timeout):
+        t = asyncio.create_task(self.check_status())
+        done, pending = await asyncio.wait({t}, timeout=timeout)
```
mrocklin: I'm curious, is there a nicer way to do this?   Does asyncio require the `create_task` function to be run before `wait`?  (I don't know as much about asyncio as I do about tornado)

dask_kubernetes/core.py
mrocklin: As a reference point, for interactive workloads (which dask-kubernetes is sometimes used for) we probably are about anything over 100ms, especially if they start to build on each other.

If we're just doing a periodic cleanup check then yeah, a few seconds or whatever is fine.  But when you hit compute you want to see workers start up more or less immediately.

mrocklin: The comment above is just for general context, not for this particular sleep.

dask_kubernetes/core.py
```diff
@@ -28,6 +29,16 @@
...
 
 
+class PodStatus(Enum):
+    pending = 'Pending'
+    running = 'Running'
+    succeeded = 'Succeeded'
+    failed = 'Failed'
+    unknown = 'Unknown'
+    completed = 'Completed'
+    crashloopbackoff = 'CrashLoopBackOff'
```
mrocklin: I have mixed thoughts about specifying things like this.  I imagine it's a matter of style.  I've personally always preferred using raw values directly rather than providing an abstraction layer between them.  This enum approach might be better as people use things like type systems and nicer editors though.  Maybe it's time for my style to change.

quasiben: After chatting with @mrocklin we came to the conclusion that for small codebases and readability we should switch to global variables rather than introduce ENUMs

dask_kubernetes/core.py
```diff
@@ -453,46 +484,53 @@ def logs(self, pod=None):
...
+                )
+                yield offload(self.scale_down, to_close)
+
+            self.scheduler.loop.add_callback(f, to_close)
+            return
 
-            # Terminate all pods without waiting for clean worker shutdown
-            self.scale_down(to_close)
+        # Terminate all pods without waiting for clean worker shutdown
+        return await self._scale_down(to_close)
```
mrocklin: It seems like we want to wait for things to retire gracefully if we can, and then only shut things down hard if we weren't able to shut things down cleanly in a reasonable amount of time.

I would have expected us to just `await self.scheduler.retire_workers(...)` above rather than schedule it as a separate callback.

dask_kubernetes/core.py
```diff
@@ -511,44 +549,18 @@ def f(to_close):
             except kubernetes.client.rest.ApiException as e:
                 # If a pod has already been removed, just ignore the error
                 if e.status != 404:
-                    raise
+                    raise kubernetes.client.rest.ApiException(e)
```
mrocklin: Why this change?  Calling `raise` will raise the exception that we're curently handling.  I think it also retains the original traceback.

dask_kubernetes/core.py
```diff
@@ -511,44 +549,18 @@ def f(to_close):
...
-        t = asyncio.create_task(self.check_status())
-        done, pending = await asyncio.wait({t}, timeout=timeout)
-
-        # if there are any pending tasks we have timed out
-        if len(pending):
-            return False
-
-        assert all([d.result() for d in done])
-        return True
+        self.task_queue.put_nowait(self._scale_down_with_pending_check)
```
mrocklin: You may have wanted `_scale_up` here

dask_kubernetes/tests/test_async.py
```diff
@@ -519,24 +519,28 @@ def pod_spec(image_name):
...
-        pytest.skip("Require a small test kubernetes cluster (maxpod <= 50)")
+        # It's probably not reasonable to run this test against a larger clusters
+        # artificially restrict to 50 pods
+        max_pods = 50
+
     extra_pods = 5
     requested_pods = max_pods + extra_pods
-    cluster.scale(requested_pods)
-
+    cluster.scale(20)
```
mrocklin: Why this change?

dask_kubernetes/core.py
mrocklin: Or maybe I should just say that there is value to waiting for the workers to transfer data before killing them.  Otherwise during adaptive workloads we end up repeating a lot of computation as our cluster scales itself down and loses results.

dask_kubernetes/core.py
```diff
@@ -275,11 +280,37 @@ def asynchronous(self):
...
+                # Get a scale_up/down out of the queue.
+                func = await self.task_queue.get()
+                # execute function
+                await func()
+                # cleanup messy pods
+                pods = await self._cleanup_terminated_pods(pods)
+
+                self.task_queue.task_done()
+
+            await asyncio.sleep(1)
```
mrocklin: I wonder if we might wait here for either 1s or when there is a new element in the task queue, whichever is first.  That might help to improve resonsiveness.

dask_kubernetes/core.py
```diff
@@ -500,27 +512,58 @@ def f(to_close):
                     raise
 
     async def _cleanup_terminated_pods(self, pods):
-        terminated_phases = {"Succeeded", "Failed"}
+        terminated_phases = {PodStatus.succeeded, PodStatus.failed}
```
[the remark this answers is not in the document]
quasiben: switched to globals

dask_kubernetes/core.py
quasiben: switched to a queueing system

dask_kubernetes/core.py
quasiben: yup!

dask_kubernetes/tests/test_async.py
quasiben: a leftover from debugging

quasiben:
This PR is currently failing a lot of tests on CircleCI.  I'm somewhat unfamiliar with CircleCI and I don't see where the output of the failing tests should be located.  I'm looking here:
https://circleci.com/gh/dask/dask-kubernetes/278

mrocklin:
It would normally follow there, but that test is hanging I think, and so circle-ci maybe just stops the process entirely before pytest has a change to pill details?

You might try including a `-s` to include debug information during the tests, or you might try removing that test.

dask_kubernetes/core.py
```diff
@@ -156,7 +163,7 @@ def __init__(
         env=None,
         auth=ClusterAuth.DEFAULT,
         asynchronous=False,
-        **kwargs
+        **kwargs,
```
mrocklin: This will kill Python 3.5 compatibility.  This is likely a black issue.  You might want to update to black master and make sure that the rest of your file is compatible with Python 3.5.  If that's the case then black won't put these in there (I ran into this today as well, and found it frustrating).

dask_kubernetes/core.py
```diff
@@ -440,45 +500,57 @@ def logs(self, pod=None):
 
         See Also
         --------
-        KubeCluster.scale_up
+        KubeCluster.scale_upc
```
mrocklin:
```suggestion
        KubeCluster.scale_up
```

dask_kubernetes/core.py
```diff
@@ -440,45 +500,57 @@ def logs(self, pod=None):
...
+        running_workers = list(self.scheduler.workers.keys())
+        running_ips = set(urlparse(worker).hostname for worker in running_workers)
+        running_names = [
+            v.name for w, v in self.scheduler.workers.items() if v.has_what
+        ]
+
+        # Dask is Fast! A worker can register with the scheduler before k8s
+        # moves the pod from `PENDING` to `RUNNING` below we remove
+        # any workers in a `PENDING` state which have non-zero
+        # memory or currently processing a task
```
mrocklin: @jacobtomlinson have you run into this before?

dask_kubernetes/core.py
quasiben: Thanks for the note!

quasiben:
@mrocklin finally have passing tests -- note I commented out test_sync.py.  Should I remove this file entirely ?

mrocklin:
> @mrocklin finally have passing tests -- note I commented out test_sync.py. Should I remove this file entirely ?

Unfortunately we still need these tests to work.  These test that this code works when run from a notebook or normal script where the user isn't writing async/await code.

dask_kubernetes/tests/test_sync.py
```diff
@@ -243,34 +243,34 @@ def test_pod_from_dict(image_name, loop, ns):
...
+#                     "image": image_name,
+#                     "imagePullPolicy": "IfNotPresent",
+#                     "name": "worker",
+#                 }
+#             ]
+#         }
+#     }
+
+#     with KubeCluster.from_dict(spec, loop=loop, namespace=ns) as cluster:
+#         cluster.adapt()
```
quasiben: @mrocklin What should `cluster.adapt` do here ?  No workers have been requested yet

mrocklin: `cluster.adapt()` starts up a coroutine that watches the activity in the scheduler and, if anything shows up, starts calling scale as necessary.

So, to start, nothing at all would happen.  But later on after we submit some work we would expect the cluster to automatically scale up.

quasiben:
closing in favor of #162

PR #136, opened 2019-04-24
k8s API changed in 9.0

athornton:
Fixes #118, fixes #135.

PR #137, opened 2019-04-25
Use versioneer for versioning

jcrist:
This is commonly used by other dask projects, and avoids having to keep
versions stable between `setup.py` and `__init__.py` (and also marks dev
versions appropriately).


jcrist:
Note that tests will fail until #136 is merged.

mrocklin:
+1

jcrist:
Error seems unrelated.

PR #142, opened 2019-04-30
Add widgets for logs

jacobtomlinson:
Add a couple of new log classes which extend `str` and `dict` to add some ipywidgets. This is aimed to make viewing logs easier.

Each log is a string which is rendered as a `<code>` block in an `HTML` widget. Then all logs are a dictionary of the log strings wrapped in an accordion view.

### Before
image

### After
image

Would appreciate some pointers on testing ipywidgets stuff.

### DIFF
--- a/dask_kubernetes/logs.py
+++ b/dask_kubernetes/logs.py
@@ -1,0 +1,28 @@
+class Log(str):
+    """A container for logs."""
+
+    def _widget(self):
+        from ipywidgets import HTML
+        return HTML(
+            value="<pre><code>{logs}</code></pre>".format(logs=self)
+        )
+
+    def _ipython_display_(self, **kwargs):
+        return self._widget()._ipython_display_(**kwargs)
+
+
+class Logs(dict):
+    """A container for multiple logs."""
+
+    def _widget(self):
+        from ipywidgets import Accordion
+        accordion = Accordion(children=[log._widget() for log in self.values()])
+        [accordion.set_title(i, title) for i, title in enumerate(self.keys())]
+        return accordion
+
+    def _ipython_display_(self, **kwargs):
+        return self._widget()._ipython_display_(**kwargs)
+
+
+
+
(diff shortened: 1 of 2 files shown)


mrocklin:
That's quite neat!

dask_kubernetes/logs.py
```diff
@@ -0,0 +1,28 @@
+class Log(str):
+    """A container for logs."""
+
+    def _widget(self):
+        from ipywidgets import HTML
```
mrocklin: What happens if `ipywidgets` isn't installed?

Also, do we need `ipywidgets` here?  Would implementing a `_repr_html_` method work instead?

```python
def _repr_html_(self):
    return "<pre><code>... "
```

I think that this has the advantage of also rendering nicely in static notebooks.

Perhaps that wouldn't work for the accordion though, so maybe this isn't worth it.

mrocklin:
We might also consider this within the core dask/distributed project and the `Client.get_scheduler.logs` and `Client.get_worker_logs` methods.

mrocklin:
I think that this is great, and would be happy to see it merged.  

@jacobtomlinson given the less-than-active maintenance response on some of these repositories I encourage you to self-merge in the future after a suitable time.  I typically say something like "Merging in 24 hours if there are no further comments" and then merge away the next day.

jhamman:
@mrocklin - what do you think we should here? I personally think this can be merged as is.

dask_kubernetes/logs.py
jhamman: > What happens if ipywidgets isn't installed?

I don't think `obj._widget` is called if ipywidgets is not importable.

mrocklin:
> @mrocklin - what do you think we should here? I personally think this can be merged as is.

Same, I say go for it

mrocklin:
Also, FWIW, I personally prefer squash-and-merge merging if you're comfortable with that

PR #145, opened 2019-05-03
move default tolerations to clean_pod_template

jhamman:
This is a followup PR to #109 and #110. 

The basic idea is that the prior work only applied the default toleration to `pod_template`s created `make_pod_spec`. This PR moves the existing toleration specifications to `clean_pod_template` to make sure the tolerations are set for all instantiations of `KubeCluster`. 

cc @jacobtomlinson and @yuvipanda

### DIFF
--- a/dask_kubernetes/objects.py
+++ b/dask_kubernetes/objects.py
@@ -134,22 +134,6 @@ def make_pod_spec(
                     env=[client.V1EnvVar(name=k, value=v)
                             for k, v in env.items()],
                 )
-            ],
-            tolerations=[
-                client.V1Toleration(
-                    key='k8s.dask.org/dedicated',
-                    operator='Equal',
-                    value='worker',
-                    effect='NoSchedule',
-                ),
-                # GKE currently does not permit creating taints on a node pool
-                # with a `/` in the key field
-                client.V1Toleration(
-                    key='k8s.dask.org_dedicated',
-                    operator='Equal',
-                    value='worker',
-                    effect='NoSchedule',
-                ),
             ]
         )
     )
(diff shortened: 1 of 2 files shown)


jhamman:
This can be tested on binder.pangeo.io: https://binder.pangeo.io/v2/gh/jhamman/pangeo_cmip6_examples/testing_tolerations

These changes produce pods with the following tolerations:

```yaml
  tolerations:
  - effect: NoSchedule
    key: k8s.dask.org/dedicated
    operator: Equal
    value: worker
  - effect: NoSchedule
    key: k8s.dask.org_dedicated
    operator: Equal
    value: worker
  - effect: NoExecute
    key: node.kubernetes.io/not-ready
    operator: Exists
    tolerationSeconds: 300
  - effect: NoExecute
    key: node.kubernetes.io/unreachable
    operator: Exists
    tolerationSeconds: 300
```

This is despite the fact the user (me) did not specify any tolerations.

yuvipanda:
@jhamman if you don't have any nods that have that taint, do the pods schedule in other nodes? Want to make sure this doesn't require people use taints.

jhamman:
@yuvipanda - that is my understanding. This doesn't even encourage pods to be placed on nodes with these taints, just excludes pods that don't tolerate them.

BTW, in the course of testing this, I've decided we may also want pod affinity rules, so I've opened #146.

jacobtomlinson:
This is also my understanding. This allows us to have dedicated worker nodes. But workers will also run on all nodes.

mrocklin:
@quasiben if you have a moment can I ask you to check out this PR?

quasiben:
I am very much in support of taints/tolerations.  My only thought is maybe including a line in the docstring (or better yet, docs) for the appropriate taint to take advantage of the setting.  I think it should be something like the following, correct ?

> kubectl taint nodes HOSTS k8s.dask.org_dedicated=worker:NoSchedule

I'd be happy to add this in a separate PR.

jhamman:
@quasiben - this is actually already in the docs: https://kubernetes.dask.org/en/latest/#deployment-details. This PR just enforces that these tolerations are implemented for all `KubeCluster` constructor paths.

It sounds like everyone is in favor of the change in concept. This is my first time writing k8s api code so I would appreciate some review of the implementation.

Issue #146, opened 2019-05-05
Feature Request: Default pod affinity 


jhamman:
Dask-kubernetes is currently configured to include two default pod/taint tolerations (see #109, #110, #145). This feature allows dask-worker pods to be the sole inhabitants of a node/node-pool, provided the cluster is configured with the corresponding taints. Another scheduling pattern would be to attract dask-worker pods to a specific set of nodes. While this has been done successfully with nodeSelectors in the past, it is not a particularly user-friendly option.

I'm wondering if we could come up with a universal dask-worker label and then have dask-kubernetes set a `preferredDuringSchedulingIgnoredDuringExecution` affinity on each pod. Has anyone tried this? Are there specific downsides to doing this? My understanding is that the nodeSelector approach will be eventually deprecated in favor of affinity rules.

I think something like this would work (haven't tested):

```yaml
spec:
  affinity:
    nodeAffinity:
      preferredDuringSchedulingIgnoredDuringExecution:
      - weight: 100
        preference:
          matchExpressions:
          - key: k8s.dask.org/node-purpose
            operator: In
            values:
            - worker
``` 

The main benefits here are:

1. This doesn't require users/admins to set anything, existing applications without labels would continue to function as normal
2. It would allow admins to direct traffic to specific nodes without asking the user to configure their pod template

Thoughts from @yuvipanda, @jacobtomlinson, @betatim?

jhamman:
For reference, the KubeSpawer includes the following affinity rule for single user notebook pods:
```yaml
  affinity:
    nodeAffinity:
      preferredDuringSchedulingIgnoredDuringExecution:
      - preference:
          matchExpressions:
          - key: hub.jupyter.org/node-purpose
            operator: In
            values:
            - user
        weight: 100
```

consideRatio:
If I understand you correctly @jhamman, you are proposing to add a default preferred nodeAffinity to the worker pods created by dask-kubernetes. This default affinity would relate to a specific dask label. I like that, and see no practical downside to doing so.

consideRatio:
### About the Z2JH default nodeAffinity
KubeSpawner get's configured by the Z2JH Helm chart / jupyterhub_config.py to have that affinity preference by default, but it can be configured to be required or to have no such affinity at all.

See: http://z2jh.jupyter.org/en/latest/reference.html#scheduling-userpods-nodeaffinity
And: http://z2jh.jupyter.org/en/latest/reference.html#scheduling-corepods-nodeaffinity

[...]

PR #147, opened 2019-05-06
Default affinity

jhamman:
At this point, this is for illustration purposes only. It builds on #145 and gives #146 a try. By default, this now creates dask-worker pods with the following affinity and tolerations:

```yaml
spec:
  affinity:
    nodeAffinity:
      - preference:
          matchExpressions:
            operator: In
            values:
            - worker
        weight: 100
  tolerations:
  - effect: NoSchedule
    key: k8s.dask.org/dedicated
    operator: Equal
    value: worker
  - effect: NoSchedule
    key: k8s.dask.org_dedicated
    operator: Equal
    value: worker
  - effect: NoExecute
    key: node.kubernetes.io/not-ready
    operator: Exists
    tolerationSeconds: 300
  - effect: NoExecute
    key: node.kubernetes.io/unreachable
    operator: Exists
    tolerationSeconds: 300
```

If my understanding is correct: A kubernetes cluster / node-pool with the following labels/taints should attract and accept dask-workers with this spec.

labels:
 - `k8s.dask.org/node-purpose : worker`

taints:
- `k8s.dask.org/dedicated=worker`
- `k8s.dask.org_dedicated=worker`

I should note that I have a node-pool with these options set and my dask-workers are not landing in the desired place yet so I'm probably still missing at least one piece of the puzzle.

### DIFF
--- a/dask_kubernetes/objects.py
+++ b/dask_kubernetes/objects.py
@@ -180,7 +180,7 @@ def make_pod_from_dict(dict_):
     )
 
 
-def clean_pod_template(pod_template):
+def clean_pod_template(pod_template, match_node_purpose='prefer'):
     """ Normalize pod template and check for type errors """
     if isinstance(pod_template, str):
         msg = ('Expected a kubernetes.client.V1Pod object, got %s'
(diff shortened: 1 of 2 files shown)


mrocklin:
@quasiben if you have a moment can I ask you to check out this PR?

jhamman:
Thanks @jacobtomlinson - I need to add a unit test or two but now that its passed your sniff test, I can forge ahead.

jhamman:
I'm working on tests now. I've also created a binder with this branch on it that can be run on binder.pangeo.io: [Binder](https://binder.pangeo.io/v2/gh/jhamman/pangeo_test_binder/dask_k8s_affinity)

dask_kubernetes/objects.py
```diff
@@ -222,4 +206,49 @@ def clean_pod_template(pod_template):
...
+    )
+    
+    if pod_template.spec.affinity is None:
+        pod_template.spec.affinity = client.V1Affinity(
+                node_affinity=client.V1PodAntiAffinity(
+                    preferred_during_scheduling_ignored_during_execution=[affinity]
+                )
+            )
+    else:
+        print('TODO: need to handle the various partial states here')
```
consideRatio: I've done partial state handling in #148, doing this was really messy sadly.

consideRatio: You are very welcome to copy paste etc to accomplish the desired outcome.

jhamman:
@consideRatio - I merged our two efforts. Let me know what you think.

dask_kubernetes/objects.py
```diff
@@ -229,4 +229,56 @@ def clean_pod_template(pod_template):
...
+        if affinity is None:
+            affinity = client.V1Affinity()
+        if affinity.node_affinity is None:
+            affinity.node_affinity = client.V1NodeAffinity()
+
+        # a common object for both a preferred and a required node affinity
+        node_selector_term = client.V1NodeSelectorTerm(
+            match_expressions=[
+                client.V1NodeSelectorRequirement(
+                    key="k8s.dask.org/node-purpose", operator="In", values=["worker"]
```
jhamman: are we okay with In vs. Equals here?

consideRatio: Logically right now they are exactly the same, I think it is possible to make it `"Equal"` (or was it `"Equals"`?) instead assuming we remove the brackets `[]` around `"worker"`.

dask_kubernetes/objects.py
```diff
@@ -229,26 +229,56 @@ def clean_pod_template(pod_template):
...
-    else:
-        print('TODO: need to handle the various partial states here')
+        elif match_node_purpose == "prefer":
+            if (
+                affinity.node_affinity.preferred_during_scheduling_ignored_during_execution
+                is None
+            ):
+                affinity.node_affinity.preferred_during_scheduling_ignored_during_execution = (
+                    []
+                )
```
consideRatio: The assignment above should be an array I think. If we could say the type of elements it should be an array of V1PreferredSChedulingTerm objects, but as we cannot type the arrays element it should simply be an array.

I understand it currently as it is a tuple with one element: an array. I'm not confident on how the parenthesis are interpreted though.

dask_kubernetes/objects.py
jhamman: Looking at this a little more, it doesn't seem like Equals is allowed here: https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1NodeSelectorRequirement.md

consideRatio: Ah

dask_kubernetes/objects.py
jhamman: This evaluates to an empty list.

```ipython
In [1]: x = ([])

In [2]: x
Out[2]: []
```

As you say, we can't type this further. AFAIK, there isn't a way to instantiate an empty list of `V1PreferredSchedulingTerm` objects, like we do for the `V1NodeSelector`s above.

consideRatio: Ah okay! Technically then all good!!

I just learned that `()` is a 0-tuple or a tuple with 0 elements, and ("test",) is a tuple with 1 element, and ("test", "more") is a tuple with two elements, but ("something") is just "something".

jhamman:
@yuvipanda / @jacobtomlinson - I think we're all done here. Ready to merge?

PR #148, opened 2019-05-12
[CLOSE] Add default node affinity (Duplicate of #147)

### LINKED ISSUE #146: Feature Request: Default pod affinity

consideRatio:
## UPDATE
I failed to notice the PR #147 attempting the same thing when I got working on this. Feel free to straight up close or copy/paste/edit parts.

---

This PR is meant to close #146 by adding a default preference for nodes with a `k8s.dask.org/node-purpose=worker` label, and optionally make the worker pods require scheduling there.

I struggled with getting my tests to run locally, so I figured I'll start by submitting this PR as a WIP.

/cc: @jhamman

### DIFF
--- a/dask_kubernetes/objects.py
+++ b/dask_kubernetes/objects.py
@@ -180,7 +180,7 @@ def make_pod_from_dict(dict_):
     )
 
 
-def clean_pod_template(pod_template):
+def clean_pod_template(pod_template, match_node_purpose="prefer"):
     """ Normalize pod template and check for type errors """
     if isinstance(pod_template, str):
         msg = ('Expected a kubernetes.client.V1Pod object, got %s'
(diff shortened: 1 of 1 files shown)


jhamman:
I've copied the meat of this PR over to #147. This can be closed.

PR #149, opened 2019-05-17
Use yaml.safe_load

mrocklin:
this avoids an annoying warning


mrocklin:
@jhamman you might want to sneak this in before the upcoming release

PR #158, opened 2019-06-26
CI: List env name

TomAugspurger:


TomAugspurger:
It looks like we're still pulling in older versions of dask / distributed.

```
# packages in environment at /home/circleci/miniconda/envs/dask-kubernetes-test:
#
# Name                    Version                   Build  Channel
...
dask                      1.2.2+76.g84ff7375          pypi_0    pypi
...
distributed               1.28.1+57.ga511f0ea          pypi_0    pypi
...
```

I think this is being cached.

```
Found a cache from build 355 at miniconda-v1-Qn_UkmIr036cNlt57c1x3bkQ_sORoIyFhXRAwqUbazs=
Size: 432 MB
Cached paths:
  * /home/circleci/miniconda

```

mrocklin:
Right, we do cache the environment keyed off of the text of the environment.yml file.  Using `@master` doesn't help us much here.  I've pushed a `dask=2` line to environment.yml, which should help?

TomAugspurger:
I suppose the best thing to do is to make two jobs, 1 against some released version of dask / distributed (2.x) and one against master.

jsignell:
I think this is no longer an issue since we use the latest docker image.

PR #162, opened 2019-07-10
WIP Use SpecCluster

jacobtomlinson:
This PR contains a rather large overhaul of the library to use the new `SpecCluster`. I've taken this opportunity to wrap up changes from other PRs adding asyncio support.

Changes:
- Add classes for `Pod`, `Worker` and `Scheduler`
- Make the scheduler a pod running on the kubernetes cluster instead of a `LocalCluster`
- Base `KubeCluster` on `SpecCluster` instead of `Cluster`
- Remove unnecessary scaling code from `KubeCluster`

Currently this relies on changes to `SpecCluster` in dask/distributed#2827.

Still to do:
- [x] Pull in all async changes
- [x] Refactor pod spec creation
- [x] Tests
- [x] Docs


jacobtomlinson:
I'm currently facing issues with the way distributed handles asyncio. 

```python
from dask_kubernetes import KubeCluster
cluster = KubeCluster()     
```

```python-traceback                                                                                                                      
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-4-3751171c0c19> in <module>
----> 1 cluster = KubeCluster()

~/Projects/dask/dask-kubernetes/dask_kubernetes/core.py in __init__(self, pod_template, name, namespace, n_workers, host, port, env, auth, **kwargs)
    334         }
    335 
--> 336         super().__init__({}, scheduler, worker, **kwargs)
    337 
    338     @classmethod

~/miniconda3/lib/python3.7/site-packages/distributed/deploy/spec.py in __init__(self, workers, scheduler, worker, asynchronous, loop, silence_logs)
    142         if not self.asynchronous:
    143             self._loop_runner.start()
--> 144             self.sync(self._start)
    145             self.sync(self._correct_state)
    146             self.sync(self._wait_for_workers)

~/miniconda3/lib/python3.7/site-packages/distributed/deploy/cluster.py in sync(self, func, asynchronous, callback_timeout, *args, **kwargs)
    243             return future
    244         else:
--> 245             return sync(self.loop, func, *args, **kwargs)

~/miniconda3/lib/python3.7/site-packages/distributed/utils.py in sync(loop, func, callback_timeout, *args, **kwargs)
    330             e.wait(10)
    331     if error[0]:
--> 332         six.reraise(*error[0])
    333     else:
    334         return result[0]

~/miniconda3/lib/python3.7/site-packages/six.py in reraise(tp, value, tb)
    691             if value.__traceback__ is not tb:
    692                 raise value.with_traceback(tb)
--> 693             raise value
    694         finally:
    695             value = None

~/miniconda3/lib/python3.7/site-packages/distributed/utils.py in f()
    315             if callback_timeout is not None:
    316                 future = gen.with_timeout(timedelta(seconds=callback_timeout), future)
--> 317             result[0] = yield future
    318         except Exception as exc:
    319             error[0] = sys.exc_info()

~/miniconda3/lib/python3.7/site-packages/tornado/gen.py in run(self)
    733 
    734                     try:
--> 735                         value = future.result()
    736                     except Exception:
    737                         exc_info = sys.exc_info()

~/miniconda3/lib/python3.7/site-packages/distributed/deploy/spec.py in _start(self)
    156         self._lock = asyncio.Lock()
    157         self.status = "starting"
--> 158         self.scheduler = await self.scheduler
    159         self.status = "running"
    160 

~/Projects/dask/dask-kubernetes/dask_kubernetes/core.py in _()
     46         async def _():
     47             async with self.lock:
---> 48                 await self.start()
     49             return self
     50 

~/Projects/dask/dask-kubernetes/dask_kubernetes/core.py in start(self)
     54         async with async_timeout.timeout(1):
     55             self.pod = await self.core_api.create_namespaced_pod(
---> 56                 self.namespace, self.pod_template
     57             )   
     58         self.status = "running"

~/miniconda3/lib/python3.7/site-packages/kubernetes_asyncio/client/api_client.py in __call_api(self, resource_path, method, path_params, query_params, header_params, body, post_params, files, response_type, auth_settings, _return_http_data_only, collection_formats, _preload_content, _request_timeout)
    164             post_params=post_params, body=body,
    165             _preload_content=_preload_content,
--> 166             _request_timeout=_request_timeout)
    167 
    168         self.last_response = response_data

~/miniconda3/lib/python3.7/site-packages/kubernetes_asyncio/client/rest.py in POST(self, url, headers, query_params, post_params, body, _preload_content, _request_timeout)
    228                                    _preload_content=_preload_content,
    229                                    _request_timeout=_request_timeout,
--> 230                                    body=body))
    231 
    232     async def PUT(self, url, headers=None, query_params=None, post_params=None,

~/miniconda3/lib/python3.7/site-packages/kubernetes_asyncio/client/rest.py in request(self, method, url, query_params, headers, body, post_params, _preload_content, _request_timeout)
    169                 raise ApiException(status=0, reason=msg)
    170 
--> 171         r = await self.pool_manager.request(**args)
    172         if _preload_content:
    173 

~/miniconda3/lib/python3.7/site-packages/aiohttp/client.py in _request(self, method, str_or_url, params, data, json, cookies, headers, skip_auto_headers, auth, allow_redirects, max_redirects, compress, chunked, expect100, raise_for_status, read_until_eof, proxy, proxy_auth, timeout, verify_ssl, fingerprint, ssl_context, ssl, proxy_headers, trace_request_ctx)
    415         timer = tm.timer()
    416         try:
--> 417             with timer:
    418                 while True:
    419                     url, auth_from_url = strip_auth_from_url(url)

~/miniconda3/lib/python3.7/site-packages/aiohttp/helpers.py in __enter__(self)
    566 
    567         if task is None:
--> 568             raise RuntimeError('Timeout context manager should be used '
    569                                'inside a task')
    570 

RuntimeError: Timeout context manager should be used inside a task
```

Any support would be appreciated. Ping @mrocklin @quasiben

quasiben:
I'm not very familiar with `async_timeout` perhaps you could use another timeout call `asyncio.wait_for` ? However, that's not usable with a context.  I recall there being issues with how  kubernetes_asyncio uses aiohttp ?

jacobtomlinson:
Hmm possibly. It's a bit of a pain that the library is swagger generated, which makes it harder to make changes to.

I think the error I'm getting could be a bit misleading. It's within `if task is None:`. So the error `Timeout context manager should be used inside a task` possibly suggests that it is using a timeout when it shouldn't. As opposed to my first assumption which was that a timeout was missing.

retrry:
Hey,
what is still needed to be done with this pull request? As I understand this and async changes are required to use dask-kubernetes with dask-labextension. 

I should be able to help test it, if there would be a need for that.

jacobtomlinson:
@retrry see the checklist in the original comment here. There are also a bunch of changes in `SpecCluster` since this was opened which need to be factored in.

Thanks for your offer to test this out. I'm aiming to finish this in the next few weeks.

dask_kubernetes/core.py
```diff
@@ -14,19 +17,186 @@
...
+    async def start(self, **kwargs):
+        self.pod = await self.core_api.create_namespaced_pod(
+            self.namespace, self.pod_template
+        )
+
+        while self.pod.status.phase == "Pending":
+            self.pod = await self.core_api.read_namespaced_pod(
+                self.pod.metadata.name, self.namespace
+            )
+            await asyncio.sleep(0.1)
```
mrocklin: I don't suppose `kubernetes_asyncio` provides any sort of event that triggers when the status changes that we could use to avoid polling?

dask_kubernetes/core.py
```diff
@@ -14,19 +17,186 @@
...
+
+        while self.address is None:
+            logs = await self.logs()
+            for line in logs.splitlines():
+                for query_string in ["worker at:", "Scheduler at:"]:
+                    if query_string in line:
+                        self.address = line.split(query_string)[1].strip()
+            await asyncio.sleep(0.1)
```
mrocklin: In other SpecCluster we treat a worker Job as "started" once it has been successfully handed off to the resource manager.  We don't actually wait for it to switch from a pending to an active state.  This might not be correct, but currently I think that various other systems, like adaptive, expect this sort of behavior.  The Cluster class will separately poll the scheduler to find out who has arrived.  

Note that this isn't true for the scheduler, for whom we do actually need the address.

dask_kubernetes/core.py
```diff
@@ -14,19 +17,186 @@
...
+    kwargs:
+        TODO Document Worker kwargs
+    """
+
+    def __init__(self, scheduler: str, **kwargs):
+        super().__init__(**kwargs)
+
+        self.scheduler = scheduler
+
+        self.pod_template.metadata.labels["dask.org/component"] = "worker"
```
mrocklin: We used to have more labels here, didn't we?  I suspect that, for example, Pangeo clusters are expecting certain labels today.

dask_kubernetes/core.py
```diff
@@ -14,19 +17,186 @@
...
+class Scheduler(Pod):
+    """ A Remote Dask Scheduler controled by SSH
+    Parameters
+    ----------
+    address: str
+        The hostname where we should run this worker
+    connect_kwargs: dict
+        kwargs to be passed to asyncssh connections
+    kwargs:
+        TODO Document Scheduler kwargs
```
mrocklin: Can we defer users to the `dask.distributed.Scheduler` docstring?

dask_kubernetes/core.py
```diff
@@ -317,226 +502,41 @@ def namespace(self):
...
-            return
-        self._delete_pods(to_delete)
+        if scheduler:
+            logs["Scheduler"] = await self.scheduler.logs()
 
-    def __enter__(self):
-        return self
+        if workers:
+            for key, worker in self.workers.items():
+                logs[key] = await worker.logs()
```
mrocklin: I suspect that we will want to await many of these at once, maybe something like this:

```python
worker_logs = await asyncio.gather(*[w.logs() for w in self.workers.values()])
for key, log in zip(self.workers, worker_logs):
    logs[key] = log
```

dask_kubernetes/core.py
```diff
@@ -181,67 +369,64 @@ def __init__(
...
+                    for k, v in self.env.items()
                 ]
             )
-        self.pod_template.metadata.generate_name = name
+        self.pod_template.metadata.generate_name = self._generate_name
+
+        finalize(
+            self,
+            self.sync,
+            _cleanup_pods,
```
mrocklin: Now that this is an async def function I suspect that `finalize` won't work well.  Also, there is a chance that the event loop is now closed.  We might have to get a little creative here.  We probably have to have a function that uses the current event loop, if it's still active, or makes a new event loop if it isn't.  Alternatively, we could do this bit synchronously with the vanilla `kubernetes` API.

I may have handled this in the async dev branch.  I don't recall.

dask_kubernetes/core.py
jacobtomlinson: That sounds reasonable. We must poll for the address before leaving the function here though. Do you propose I call the super start (which switches the status) higher up?

dask_kubernetes/tests/test_core.py
```diff
@@ -91,25 +93,28 @@ def test_basic(cluster, client):
     assert all(client.has_what().values())
 
 
-def test_logs(cluster):
+@pytest.mark.asyncio
+async def test_logs(cluster):
+    logs = cluster.logs()
```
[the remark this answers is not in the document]
jacobtomlinson: Would you? The `logs` method is sync and the `_logs` method is async.

dask_kubernetes/core.py
jacobtomlinson: I don't think so.

dask_kubernetes/core.py
jacobtomlinson: Those still exist, common labels are set all together when the pod spec is cleaned. This is just setting a component specific label once we are into differentiated `Scheduler` and `Worker` classes.

dask_kubernetes/core.py
jacobtomlinson: Yeah I was noticing some warnings related to this. Vanilla k8s may be my preference.

dask_kubernetes/core.py
mrocklin: We might have to treat schedulers differently than workers in this respect.  I don't think we need the worker address any more.

dask_kubernetes/tests/test_core.py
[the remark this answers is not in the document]
mrocklin: The `logs` definition actually uses `client.sync`, which magically switches to asynchronous if it detects that it is being used asynchronously.  If you look at `type(logs)` I suspect that you will find that it is an `asyncio.Future`.

dask_kubernetes/core.py
jacobtomlinson: Ah ok!

dask_kubernetes/tests/test_core.py
```diff
@@ -29,59 +29,61 @@
...
-def pod_spec(image_name):
+async def pod_spec(image_name):
     yield make_pod_spec(
         image=image_name, extra_container_config={"imagePullPolicy": "IfNotPresent"}
     )
 
 
 @pytest.fixture
-def clean_pod_spec(pod_spec):
+async def clean_pod_spec(pod_spec):
```
mrocklin: Given that there are no `await` statements in this function you can probably drop the `async` as well.

dask_kubernetes/tests/test_core.py
```diff
@@ -29,59 +29,61 @@
...
+@pytest.mark.asyncio
+async def test_basic(cluster, client):
     cluster.scale(2)
     future = client.submit(lambda x: x + 1, 10)
     result = future.result()
     assert result == 11
 
-    while len(cluster.scheduler.workers) < 2:
+    while len(cluster.workers) < 2:
         sleep(0.1)
```
mrocklin: These two checks are actually different.

If you want to wait for two workers on the cluster, then you can just do `await cluster`

However if you want to wait until the scheduler has seen two workers, then you might do `cluster.scheduler_info["workers"]`.  This value is updated by a communication channel regularly.

dask_kubernetes/tests/test_core.py
```diff
@@ -124,13 +128,15 @@ def test_ipython_display(cluster):
         sleep(0.5)
 
 
-def test_dask_worker_name_env_variable(pod_spec, loop, ns):
+@pytest.mark.asyncio
+async def test_dask_worker_name_env_variable(pod_spec, loop, ns):
```
mrocklin: You have `async def` here, but `with KubeCluster` below.  Mixing async and non-async code will probably result in unpleasant hangs.

mrocklin:
I see that you're exploring DNS config quite a bit these days :)

jacobtomlinson:
I've definitely had a fun morning with getting DNS in minikube working. But I think I'm there now and can actually get back to my unit tests 😆.

dask_kubernetes/core.py
```diff
@@ -203,7 +203,7 @@ def __init__(self, **kwargs):
         # TODO Resolve service race condidion
         # There is some race condition happening here where the service isn't actually ready by the time
         # this function exits.
-        await asyncio.sleep(2)
+        await asyncio.sleep(5)
```
mrocklin: Oof, that's a long sleep

jacobtomlinson: It sure is! I've been cranking it up from `0.1` and I've only now managed to get the tests to pass. 

I've been digging into the k8s API to try and find something that I can watch for a state change to avoid this sleep, but so far no luck.

mrocklin: Is this the problem that we were running into over screenshare?  

if it's a major pain we can always try https://github.com/dask/distributed/issues/3084

jacobtomlinson: Yeah. I'm trying to solve it cleanly, but 3084 may be a good workaround.

mrocklin:
Oooh, green check marks :)

jacobtomlinson:
After further testing on the [Pangeo Binder](https://hub.aws-uswest2-binder.pangeo.io) it seems that we do not have enough permissions there to create the service.

dask_kubernetes/core.py
jacobtomlinson: Is there a neat way of doing this in Python? Or should I just provide a link to the dask docs?

mrocklin: Probably, but I would just link, and maybe add an entry in `See Also`

mrocklin:
> After further testing on the Pangeo Binder it seems that we do not have enough permissions there to create the service.

Hrm, is the Pangeo binder deployment particularly strict, or is this likely to be common?

mrocklin:
Also, it would be interesting to maintain a local scheduler option.  My hope is that all we have to do here is to swap out the `Scheduler(Pod)` class with the `dask.distributed.Scheduler` class.  My guess is that this will come in a future PR, but I thought I'd mention it now, just in case that option makes it easier to get this out the door.

jacobtomlinson:
> Hrm, is the Pangeo binder deployment particularly strict, or is this likely to be common?

It is quite strict, I wrote the original permissions as they are set in the pangeo helm chart. I've raised a PR to expand them, it just may take some time for the change to trickle through all the deployments.

> Also, it would be interesting to maintain a local scheduler option. 

Yes that would definitely be easy to do! Provided the default local scheduler address is routable from the workers.

jacobtomlinson:
Outstanding issues:
- [ ] Dashboard url is broken due to it no longer being local to the notebook
- [ ] There is an unpleasant sleep after the service creation to avoid race conditions
- [ ] Calling the repr of the distributed client gives `AttributeError: 'Scheduler' object has no attribute 'identity'`

mrocklin:
Tests pass!  Do we get to check another checkbox ?

jacobtomlinson:
Yep! Things are looking good here now. However we will need to wait for another distributed release for this now.

We are still unable to proxy to the dashboard though.

mrocklin:
> However we will need to wait for another distributed release for this now.

I can push out something soonish.  Let's aim for the end of this week.

mrocklin:
> We are still unable to proxy to the dashboard though.

Do we (you) have a plan?

jacobtomlinson:
If folks are happy with jupyterhub/jupyter-server-proxy#154 then we could tweak the dashboard URL to work with that.

athornton:
I wrote a convenience class to do dashboard (and worker) proxying (specifically for the JupyterHub/k8s case, where the workers are on a private network inaccessible to the user, whose only entry point into the system is an HTTPS endpoint).  Take a gander at https://github.com/lsst-sqre/jupyterlabutils ; the `sync()` override in LSSTDaskClient seems to have been a bad idea (and the problem I thought I was solving was elsewhere), but the rest of the client wrapper and the ClusterProxy object seem to work fine.

athornton:
`repr_html()` doesn't work:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/opt/lsst/software/stack/python/miniconda3-4.5.12/envs/lsst-scipipe-1172c30/lib/python3.7/site-packages/IPython/core/formatters.py in __call__(self, obj)
    343             method = get_real_method(obj, self.print_method)
    344             if method is not None:
--> 345                 return method()
    346             return None
    347         else:

/opt/lsst/software/stack/python/miniconda3-4.5.12/envs/lsst-scipipe-1172c30/lib/python3.7/site-packages/distributed/client.py in _repr_html_(self)
    792             and self.cluster.scheduler
    793         ):
--> 794             info = self.cluster.scheduler.identity()
    795             scheduler = self.cluster.scheduler
    796         elif (

AttributeError: 'Scheduler' object has no attribute 'identity'
```

....and the started pods seem to die after a while:
```
distributed.core - INFO - Starting established connection
distributed.scheduler - INFO - Scheduler closing...
distributed.scheduler - INFO - Scheduler closing all comms
distributed.scheduler - INFO - Remove worker tcp://10.24.42.236:38505
distributed.core - INFO - Removing comms to tcp://10.24.42.236:38505
distributed.scheduler - INFO - End scheduler at 'tcp://10.24.38.233:8786'
```

mrocklin:
Try updating distributed to 2.5.1

athornton:
I already am at 2.5.1:

```
[athornton@nb-athornton-exp-d-2019-10-01-daskb ~]$ pip freeze | grep '^dask'
dask==2.1.0+9.g76af7b7
dask-kubernetes==0.9.2+52.ga18a2b2
dask-labextension==1.0.3
[athornton@nb-athornton-exp-d-2019-10-01-daskb ~]$ pip freeze | grep '^distributed'
distributed==2.5.1
```

jacobtomlinson:
Many thanks for testing this and giving pointers!

> repr_html() doesn't work:

I am surprised as that was fixed upstream.

> the started pods seem to die after a while

The scheduler [will timeout](https://github.com/dask/dask-kubernetes/pull/162/files#diff-6e124cd9e896a4e40493db8ca10fa968R146) after 5 mins of inactivity.

athornton:
I'll grab distributed from master.  Good to know that's a configurable timeout.  In practice, once I have a build that has settled down long enough for the prepuller to pull it, that won't be a problem.

athornton:
Oh hey this very issue has a question I need the answer to.  @jacobtomlinson what is the new minimal set of permissions we need?

jacobtomlinson:
See https://github.com/pangeo-data/helm-chart/pull/99. Adding documentation to this PR is on my list to do before it gets merged.

mrocklin:
Distributed was released, so the repr thing should be safe

jhamman:
@jacobtomlinson - are we just waiting on https://github.com/jupyterhub/jupyter-server-proxy/issues/152 or is there more to do here? Are you waiting on additional testing. I think @scottyhq had given this a bit of exercise a few weeks ago.

jacobtomlinson:
@jhamman yeah I'm waiting on that issue. I also need to a bit of docs updating.

More testing is very welcome, but @scottyhq has been awesome so far.

mrocklin:
I would very much like to get this in.  If it's unlikely that we can merge-and-releaase the jupyter-server-proxy work soon, then can we modify this work to default to a local Scheduler until that finishes up?  

We're in an awkward position right now with the latest release of distributed being somewhat broken with the latest release of dask-kubernetes.

jacobtomlinson:
> I would very much like to get this in. If it's unlikely that we can merge-and-releaase the jupyter-server-proxy work soon, then can we modify this work to default to a local Scheduler until that finishes up?

I think setting the local scheduler as the default and getting this merged would be my preference regardless. 

The issue we are blocked on isn't actually in a hard dependency of this library. It's just a library that is commonly used with it. Therefore people are likely to have issues unless we make it a hard dependency which we aren't going to do.

I'll get this done now.

jacobtomlinson:
I've changed things so that the default behaviour is for the scheduler to be local to the cluster manager. There is also a kwarg called `local_scheduler` which you can set to `False` for the scheduler to run in a pod on the cluster. 

I would really apprciate more testing from @scottyhq, @jhamman, etc.

dask_kubernetes/core.py
```diff
@@ -14,20 +17,207 @@
...
+            self._pod.metadata.name, self.namespace
+        )
+        return self._pod
+
+    def __repr__(self):
+        return "<Pod %s: status=%s>" % (type(self).__name__, self.status)
+
+
+class Worker(Pod):
+    """ A Remote Dask Worker controled by SSH
```
mrocklin:
```suggestion
    """ A Remote Dask Worker controled by Kubernetes
```

dask_kubernetes/core.py
```diff
@@ -14,20 +17,207 @@
...
+    Parameters
+    ----------
+    scheduler: str
+        The address of the scheduler
+    address: str
+        The hostname where we should run this worker
+    connect_kwargs: dict
+        kwargs to be passed to asyncssh connections
+    kwargs:
+        TODO Document Worker kwargs
```
mrocklin: Lingering TODO

dask_kubernetes/core.py
```diff
@@ -14,20 +17,207 @@
...
+        self.pod_template.metadata.labels["dask.org/component"] = "worker"
+        self.pod_template.spec.containers[0].env.append(
+            kubernetes.client.V1EnvVar(
+                name="DASK_SCHEDULER_ADDRESS", value=self.scheduler
+            )
+        )
+
+
+class Scheduler(Pod):
+    """ A Remote Dask Scheduler controled by SSH
```
mrocklin:
```suggestion
    """ A Remote Dask Scheduler controled by Kubernetes
    
```

dask_kubernetes/core.py
```diff
@@ -14,20 +17,207 @@
...
+        while self.address is None:
+            logs = await self.logs()
+            for line in logs.splitlines():
+                if "Scheduler at:" in line:
+                    self.address = line.split("Scheduler at:")[1].strip()
+            await asyncio.sleep(0.1)
+
+        self.service = await self._create_service()
+        self.address = "tcp://{name}.{namespace}:{port}".format(
+            name=self.service.metadata.name, namespace=self.namespace, port=8786
```
mrocklin: Is this port intended to be hard-coded?  Or does it not matter because of Kubernetes magic?

dask_kubernetes/core.py
```diff
@@ -14,20 +17,207 @@
     yaml = False
 
 import dask
-from distributed.deploy import LocalCluster, Cluster
+from distributed.deploy import SpecCluster, ProcessInterface
 from distributed.comm.utils import offload
-import kubernetes
+from distributed.utils import Log, Logs
+from distributed.scheduler import Scheduler as LocalScheduler
```
mrocklin:
```suggestion
from distributed import Scheduler as LocalScheduler
```

mrocklin: We might also `import dask.distrtibuted` and then use the fully namespaced `dask.distributed.Scheduler` term where we would use `LocalScheduler`

dask_kubernetes/core.py
```diff
@@ -182,67 +410,69 @@ def __init__(
...
-                raise
+        common_options = {
+            "core_api": self.core_api,
+            "pod_template": self.pod_template,
+            "namespace": self._namespace,
+            "loop": self.loop,
+        }
+
+        if self._local_scheduler:
+            self.scheduler_spec = {"cls": LocalScheduler, "options": {}}
```
mrocklin: It might be important to pass through some minimal set of options here.  This is what we use for dask-jobqueue:

```python
        scheduler = {
            "cls": Scheduler,  # Use local scheduler for now
            "options": {
                "protocol": protocol,
                "interface": interface,
                "host": host,
                "dashboard_address": dashboard_address,
                "security": security,
            },
        }
```

dask_kubernetes/core.py
jacobtomlinson: Yeah it is kubernetes magic. The scheduler runs on the default ports and we need to tell kubernetes what those are. I agree that having magic numbers floating about is unpleasant. 

Would it be preferable to move to a constant?

jacobtomlinson:
Ok I think this is ready for final review and hopefully a merge.

I've done some more manual testing on the Pangeo Binder with the switch to a local scheduler by default and everything seems to be working ok.

dask_kubernetes/core.py
```diff
@@ -154,25 +344,69 @@ def __init__(
...
+        self.host = self.host or dask.config.get("kubernetes.host")
+        self.port = (
+            self.port if self.port is not None else dask.config.get("kubernetes.port")
+        )
+        self._protocol = self._protocol or dask.config.get("kubernetes.protocol")
+        self._interface = self._interface or dask.config.get("kubernetes.interface")
+        self._dashboard_address = self._dashboard_address or dask.config.get(
+            "kubernetes.dashboard_address"
+        )
+        self._security = self._security or dask.config.get("kubernetes.security")
```
mrocklin: I think that we might want to verify that this obect is a [`distributed.Security`](https://github.com/dask/distributed/blob/master/distributed/security.py) object.  Although @jcrist would know more about conventions here.

requirements.txt
```diff
@@ -1,3 +1,4 @@
-dask>=0.19.0
-distributed>=1.22.0
+dask>=2.3.0
+distributed>=2.3.0
```
mrocklin: Do these need to be bumped to 2.5.2?

dask_kubernetes/core.py
jacobtomlinson: Ah ok, so I guess grabbing this from the config might not make sense in this case?

jcrist:
> There is also a kwarg called local_scheduler which you can set to False for the scheduler to run in a pod on the cluster.

In dask-yarn we call this `deploy_mode="local"` and `deploy_mode="remote"`, which was chosen since users were familiar with spark which also has a "deploy-mode" flag. I don't think we need to be uniform here in kwarg name, just thought I'd mention it.

dask_kubernetes/core.py
jcrist: You might store a full security object spec as a dictionary under `kubernetes.security`, but I'd guess this would most often be a runtime parameter. I'd leave it out of config for now.

jacobtomlinson: Ok thanks! Should I be type checking here too?

jacobtomlinson:
> I don't think we need to be uniform here in kwarg name, just thought I'd mention it.

I'm more than happy to be uniform. Consistency is good! I wasn't 100% happy with `local_scheduler` anyway.

dask_kubernetes/core.py
jcrist: Probably? Should either be `None` or a `distributed.security.Security` object.

mrocklin:
@jacobtomlinson thoughts on merging?

jacobtomlinson:
Yep I think I'm ready!

mrocklin:
Woo!

jhamman:
Nice work @jacobtomlinson. Glad to see this come together!

PR #171, opened 2019-08-15
Prepare for Dask 2.3.0 release

jrbourbeau:
This is just to check that things work with the current `master` branch of `dask/dask` and `dask/distributed` prior to releasing `2.3.0` upstream. 

I think the CI already runs against upstream `master` branches, so I've included a trivial change to the README to trigger CI

### DIFF
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -538,8 +538,8 @@ class KubeCluster(Cluster):
         return self.cluster.scheduler_comm
 
     @property
-    def worker_spec(self):
-        return self.cluster.worker_spec
+    def scheduler_info(self):
+        return self.cluster.scheduler_info
 
 
 def _cleanup_pods(namespace, labels):
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -545,6 +545,10 @@ class KubeCluster(Cluster):
     def scheduler_info(self):
         return self.cluster.scheduler_info
 
+    @property
+    def periodic_callbacks(self):
+        return self.cluster.periodic_callbacks
+
 
 def _cleanup_pods(namespace, labels):
     """ Remove all pods with these labels in this namespace """
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -537,6 +537,10 @@ class KubeCluster(Cluster):
     def scheduler_comm(self):
         return self.cluster.scheduler_comm
 
+    @property
+    def worker_spec(self):
+        return self.cluster.worker_spec
+
 
 def _cleanup_pods(namespace, labels):
     """ Remove all pods with these labels in this namespace """
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -22,6 +22,7 @@ from tornado import gen
 from .objects import make_pod_from_dict, clean_pod_template
 from .auth import ClusterAuth
 from .logs import Log, Logs
+from .adaptive import Adaptive
 
 logger = logging.getLogger(__name__)
 
@@ -377,6 +378,9 @@ class KubeCluster(Cluster):
                 pod.metadata.name, pod.metadata.namespace
             )
         )
+
+    def adapt(self, Adaptive=Adaptive, **kwargs):
+        return super().adapt(Adaptive=Adaptive, **kwargs)
 
     def scale(self, n):
         """ Scale cluster to n workers
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,9 +39,6 @@ jobs:
       - run:
           command: |
             sudo kubectl get deployment
-      - restore_cache:
-          keys:
-            - miniconda-v1-{{ checksum "ci/environment-3.7.yml" }}
       - run:
           name: install miniconda
           command: |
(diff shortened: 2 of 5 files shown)


dask_kubernetes/core.py
```diff
@@ -537,6 +537,10 @@ def __exit__(self, type, value, traceback):
     def scheduler_comm(self):
         return self.cluster.scheduler_comm
 
+    @property
+    def worker_spec(self):
+        return self.cluster.worker_spec
```
mrocklin: This shouldn't be necessary.  The distributed package shouldn't require this attribute.

mrocklin: We use it a couple times in case we're dealing with a SpecCluster, but I had hoped that we weren't strict in our dependence.  In particular, dask-yarn doesn't work like this, and there aren't plans to make it do so in the future.

jrbourbeau:
It appears there's some interesting caching-related behavior happening with the CI. The normal build triggered when I push a commit times out (ref https://circleci.com/gh/dask/dask-kubernetes/394) while if I rebuild the job without a cache the CI completes (ref https://circleci.com/gh/dask/dask-kubernetes/395). Will look into this more tomorrow

dask_kubernetes/core.py
jrbourbeau: Thanks for pointing that out. I've removed it

TomAugspurger:
I wonder if your cached builds are ending up with different versions of dask / distributed on the workers vs. the scheduler. I think the scheduler image is built against master, while the workers are from the cached conda env.

mrocklin:
The errors in the most recent CI build are definitely due to a mismatch in versions between scheduler and worker.  We added the `extra` attribute recently.

mrocklin:
This failure seems possibly more concerning

```python-traceback
  File "/home/circleci/miniconda/envs/dask-kubernetes-test/lib/python3.7/site-packages/tornado/ioloop.py", line 743, in _run_callback
    ret = callback()
  File "/home/circleci/miniconda/envs/dask-kubernetes-test/lib/python3.7/site-packages/tornado/ioloop.py", line 767, in _discard_future_result
    future.result()
  File "/home/circleci/miniconda/envs/dask-kubernetes-test/lib/python3.7/site-packages/distributed/deploy/adaptive_core.py", line 182, in adapt
    recommendations = await self.recommendations(target)
  File "/home/circleci/miniconda/envs/dask-kubernetes-test/lib/python3.7/site-packages/distributed/deploy/adaptive_core.py", line 133, in recommendations
    plan = self.plan
  File "/home/circleci/miniconda/envs/dask-kubernetes-test/lib/python3.7/site-packages/distributed/deploy/adaptive.py", line 104, in plan
    return set(self.cluster.workers)
TypeError: unhashable type: 'V1Pod'
```

We recently changed adaptive to track sets of launched workers (`plan/requested`) and a set of arrived workers (`observed`).  We implement those for most clusters here:

https://github.com/dask/distributed/blob/6302175056742d101b8af52336bc6bbf4227da35/distributed/deploy/adaptive.py#L99-L112

Currently these are expected to get a collection that is keyed by worker name in order to handle comparisons.  One solution would be to specify the worker name as we launch a pod, and keep that name->pod mapping around.

mrocklin:
As an alternative, we might also ...

1.   change requirements in dask-kubernetes to avoid the newest release for now
2.  bring back the old Adaptive code into this library

mrocklin:
In case we want to go with option 2, the PR that made this change is here: https://github.com/dask/distributed/pull/2904

We would probably walk one commit back from there, copy the `adaptive.py` file into here, and then overwrite the `adapt` method to use that class rather than the one in distributed (there is an `Adaptive=` keyword we can use).

jrbourbeau:
@mrocklin thanks for all the feedback. Do you have a preference for any of the three options? You're more familiar with the current development practices of `dask-kubernetes`. 

For context, is option 2 intended to be a workaround until a more long term fix for `plan/requested/observed` is added?

mrocklin:
They're all workarounds.  I think that option 2 is probably the most straightforward for someone not familiar with Kubernetes to handle quickly (I hope).

jrbourbeau:
Alright, I'll give option 2 a go

mrocklin:
I can also do it if things get too involved.  From my perspective it's also
nice to increase exposure to these parts of the codebase to others.

jrbourbeau:
Agreed, knowledge transfer is definitely a good thing. Thanks for guiding here.

jrbourbeau:
Came across a configuration setting change in `dask/dask` that impacted `dask_kubernetes/tests/test_core.py::test_maximum`. I fixed the issue here and raised an issue upstream (ref https://github.com/dask/dask/issues/5282)

jrbourbeau:
CI is passing now

FWIW I'm in favor of keeping the removal of CI caching. The `conda` environment setup doesn't seem to take that long (~90 seconds) and caching is causing some issues with this and other PRs (ref https://circleci.com/gh/dask/dask-kubernetes/396). That said, I will add it back if other feel differently.

TomAugspurger:
Agreed w.r.t. caching.

jacobtomlinson:
This all looks reasonable to me. Thanks for picking this up.

Are we ready to go with this?

jrbourbeau:
I think so. @mrocklin any comments or concerns?

mrocklin:
Yes, this seems fine to me

jrbourbeau:
Great, merging then. Thanks all for reviewing!

Issue #170, opened 2019-08-13
SSL error when cleaning up pods


giordyb:
Hi,
I'm having a small issue running dask-kubernetes on a local Kubernetes 1.14.3 cluster (the one provided by the latest docker desktop: the job runs fine and I get back the correct results but it looks like there is an SSL issue when it tries to clean up the pods.
`  File "/usr/local/lib/python3.7/site-packages/dask_kubernetes/core.py", line 544, in _cleanup_pods`
This is the error that I get when I run my script job:

```
root@69feaef9d947:/app# python process_receipts_dask2_kubernetes.py --testrun
/usr/local/lib/python3.7/site-packages/distributed/client.py:2: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  from collections import defaultdict, Iterator
/usr/local/lib/python3.7/site-packages/distributed/publish.py:1: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  from collections import MutableMapping
/usr/local/lib/python3.7/site-packages/distributed/scheduler.py:2: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  from collections import defaultdict, deque, OrderedDict, Mapping, Set
dasboard: {'dashboard': 8787}
                               file           data_gt  ... partita_iva_ocr partita_iva_match
0  MV_trfktqxxci_20180830070237.jpg               nan  ...     xxxxxxxxxxx              True
0  MV_tentyvsblx_20180903070230.jpg  29/08/2018 12.08  ...            None             False
0  MV_qrmiwvlyio_20180908100205.jpg               nan  ...            None             False
0  MV_jxavynnpmd_20180903070227.jpg               nan  ...            None             False
0  MV_dpaztptfio_20180908100214.jpg        06/09/2018  ...            None             False

[5 rows x 10 columns]
[...]

PR #172, opened 2019-08-15
fixed _cleanup_pods auth using kubectl config

giordyb:
Hi, as requested here is the pull request to fix issue #170.

Since _cleanup_pods is only called from the Cluster object I thought it would be cleaner to just avoid creating a new kubernetes.client.CoreV1Api() object and just re-use the one created on _init by passing it as a argument.

thanks,

Giordano

### DIFF
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -234,7 +234,7 @@ class KubeCluster(Cluster):
             )
         self.pod_template.metadata.generate_name = name
 
-        finalize(self, _cleanup_pods, self.namespace, self.pod_template.metadata.labels)
+        finalize(self, _cleanup_pods, self.namespace, self.pod_template.metadata.labels, self.core_api)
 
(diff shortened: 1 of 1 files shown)


jakirkham:
I think Jacob's out today, but should be back in a day. Hopefully he can take a look then 🙂

jacobtomlinson:
Thanks for this! These CI issues should be fixed after #171. Holding until then.

jacobtomlinson:
Since #162 has been merged this now has conflicts.

Also one thing we discussed as part of that PR was adding a change like this. Sadly due to switching to asyncio for the kubernetes API we can't guarantee that the loop is still running at finalize time. This means passing the now async API object to the finalize call isn't suitable. In the rewrite, we actually create a new sync version of the API object during cleanup.

Thanks anyway for making this change! If you have thoughts or comments we would be keen to hear them.

PR #179, opened 2019-09-30
WIP Support multiple worker types DO NOT MERGE

joeschmid:
This is currently just a prototype to try out supporting multiple worker types, each with different pod specs. The approach is roughly:

• Leave the existing worker spec as a special "default" worker type
• Adaptive mode would only use the default worker type
• cluster.scale(n) would use the default worker type
• Provide the ability to add any number of worker types, stored as dict entries of str -> pod spec
• Provide the ability to call cluster.scale(n, worker_type='gpu') to manually scale specify worker types

### DIFF
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -235,6 +235,8 @@ class KubeCluster(Cluster):
             )
         self.pod_template.metadata.generate_name = name
 
+        self.worker_type_pod_templates = {}
+
         finalize(self, _cleanup_pods, self.namespace, self.pod_template.metadata.labels)
 
         if n_workers:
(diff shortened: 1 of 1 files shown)


jacobtomlinson:
This is definitely an interesting use case. I would like to know more about how you are using dask-kubernetes. 

It sounds like you are creating a large long living Dask cluster with a variety of worker types in order to back your Prefect pipeline.

This libraries primary focus is on short lived ad-hoc clusters.

> Dask Kubernetes deploys Dask workers on Kubernetes clusters using native Kubernetes APIs. It is designed to dynamically launch short-lived deployments of workers during the lifetime of a Python process.

I wonder if this functionality would also be suited for the [Dask helm chart](https://github.com/helm/charts/tree/master/stable/dask) which targets long lived clusters.

mrocklin:
@jacobtomlinson are long-lived clusters in-scope, or would we want to use entirely different Kubernetes primitives than we're using here today.

jacobtomlinson:
My preference would be that long-lived clusters should use resources with better lifecycle support. For example, the scheduler should really be a deployment so that it would be restarted if it failed.

This library also assumes that the cluster should be garbage collected when the Python session is closed or the cluster object is deleted. This may not be true for long-lived clusters.

I'm always keen to see if we could align the two projects more, but they do seem to be tageting different use cases.

mrocklin:
> I'm always keen to see if we could align the two projects more, but they do seem to be tageting different use cases.

Agreed.  I'm mostly thinking about longer-term work.  @jcrist seems to have done this with Dask-Yarn, for example.  It might be good to have him write up what semantics might be for optional garbage collection with the Python processes.

Both the scheduler-deployment and pythhon-GC topics seem like they might be optional in the future?

jacobtomlinson:
> maybe this isn't a great use of KubeCluster or at least not how it was intended?

It's not how it was intended (at least from me), but I'm glad it is useful to you! 

> It may be possible that we could move to a model where at the start of each Prefect Flow run we create a new, dynamic Dask cluster and specify the types of workers (and maybe counts if it were non-adaptive) that we want for that Flow

That would align more with this library. However I'm not suggesting that your workflow is wrong, rather that perhaps there are better tools out there. 

I would imagine a version of the [Dask helm chart](https://github.com/helm/charts/tree/master/stable/dask) may give you more stability for longer running clusters. It uses Deployment resources which have built in lifecycle policies at the Kubernetes level. You would also manage scaling in Kubernetes too, which means no adaptivity.

joeschmid:
Thanks @jacobtomlinson. We started with the Dask helm chart (and that was a good way to begin), but we wanted to make a bunch of customizations (run multiple instances of JupyterLab on different AWS instance sizes, etc.) so we ended up doing our own k8s yaml files for Deployments, Services, etc., but starting from what the Dask helm chart produced.

It turns out that Prefect does have a mechanism for running Flows on a dynamically created Dask cluster using dask-kubernetes. FYI, I opened an issue in Prefect's repo about eventually supporting multiple worker types: https://github.com/PrefectHQ/prefect/issues/1586 I referenced for PR for using SpecCluster and pointed them to this discussion.

@jacobtomlinson probably nothing else to do on this for now, but definitely interested in future support for multiple worker types in a Dask cluster dynamically created with dask-kubernetes. That can definitely wait as I know you're still getting the SpecCluster PR merged. Thanks!

jacobtomlinson:
Ok great! I'll close this out for now but let's keep track of this.

PR #189, opened 2019-10-15
Remove Adaptive class

mrocklin:
The upstream class should work fine?


jacobtomlinson:
Yeah, this is dead code now. Thanks for clearing it up.

PR #203, opened 2019-10-28
adding gcc to EXTRA_APT_PACKAGES to fix error

ksaur:
This pull request fixes a bug in the documentation. It seems the fastparquet packet requires gcc, so without `gcc` in `EXTRA_APT_PACKAGES` this error will occur:
```
  copying fastparquet/test/test_util.py -> build/lib.linux-x86_64-3.7/fastparquet/test
  copying fastparquet/test/test_with_n.py -> build/lib.linux-x86_64-3.7/fastparquet/test
  .....
  building 'fastparquet.speedups' extension
  creating build/temp.linux-x86_64-3.7
  creating build/temp.linux-x86_64-3.7/fastparquet
  gcc -pthread -B /opt/conda/compiler_compat -Wl,--sysroot=/ -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -fPIC -I/opt/conda/include/python3.7m -I/opt/conda/lib/python3.7/site-packages/numpy/core/include -c fastparquet/speedups.c -o build/temp.linux-x86_64-3.7/fastparquet/speedups.o
  unable to execute 'gcc': No such file or directory
  error: command 'gcc' failed with exit status 1
```

It appears that gcc is not in the base image specified in this example:
```
$ docker run -i -t daskdev/dask-notebook /bin/bash
jovyan@d011bb568da8:~$ gcc
bash: gcc: command not found
```

Adding gcc` in `EXTRA_APT_PACKAGES` fixes this documentation bug

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -45,13 +45,11 @@ Quickstart
         containers:
         - image: daskdev/dask:latest
           imagePullPolicy: IfNotPresent
-          args: [dask-worker, --nthreads, '2', --no-bokeh, --memory-limit, 6GB, --death-timeout, '60']
+          args: [dask-worker, $(DASK_SCHEDULER_ADDRESS), --nthreads, '2', --no-dashboard, --memory-limit, 6GB, --death-timeout, '60']
           name: dask
           env:
             - name: EXTRA_PIP_PACKAGES
-              value: fastparquet git+https://github.com/dask/distributed
-            - name: EXTRA_APT_PACKAGES
-              value: gcc
+              value: git+https://github.com/dask/distributed
           resources:
             limits:
               cpu: "2"
@@ -188,6 +186,10 @@ are available on https://hub.docker.com/r/daskdev .
 More information about these images is available at the
 `Dask documentation <https://docs.dask.org/en/latest/setup/docker.html>`_.
 
+Note that these images can be further customized with extra packages using
+``EXTRA_PIP_PACKAGES``, ``EXTRA_APT_PACKAGES``, and ``EXTRA_CONDA_PACKAGES``
+as described in the
+`Extensibility section <https://docs.dask.org/en/latest/setup/docker.html#extensibility>`_.
 
 Deployment Details
 ------------------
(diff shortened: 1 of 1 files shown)


TomAugspurger:
I'd prefer changing the example `EXTRA_PIP_PACKAGES` to a pure python package. I worry a bit about users copying these examples without understanding that bringing in `gcc` will increase the size of their docker image substantially.

We could certainly have a followup note saying that packages with C extensions may require a compiler.

ksaur:
Ok, i'm not familiar enough with what your team wants for an example to make this change then. I just wanted to help users that will inevitably run into this same issue.

TomAugspurger:
So, concretely, we can change thet main `worker-spec.yml` example to have

```
...
  - name: EXTRA_PIP_PACKAGES
    value: git+https://github.com/dask/distributed
```

And have a section below, probably in the "Docker Images" showing how to further customize the environment. I'd recommend just mentioning `EXTRA_PIP_PACKAGES`, `EXTRA_APT_PACKAGES`, and `EXTRA_CONDA_PACKAGES`, since those are more fully documented in https://docs.dask.org/en/latest/setup/docker.html#extensibility.

ksaur:
Ok as you suggested I simplified the example and then just mentioned the `EXTRA`'s and put a link as specified. I also updated the args to work with the current setup.

doc/source/index.rst
```diff
@@ -45,11 +45,11 @@ Quickstart
         containers:
         - image: daskdev/dask:latest
           imagePullPolicy: IfNotPresent
-          args: [dask-worker, --nthreads, '2', --no-bokeh, --memory-limit, 6GB, --death-timeout, '60']
+          args: [dask-worker, $(DASK_SCHEDULER_ADDRESS), --nthreads, '2', --no-dashboard, --memory-limit, 6GB, --death-timeout, '60']
```
jacobtomlinson: I don't think the `$(DASK_SCHEDULER_ADDRESS)` syntax is correct here, the brackets mean the value of the variable will be executed. 

Also the worker should be able to pick up this environment variable itself and shouldn't need specifying in the CLI if set. 

Can we drop this again or are you have specific issues with this?

PR #207, opened 2019-11-27
WIP: change scheduler_timeout to idle_timeout

### LINKED ISSUE #206: scheduler closing with deploy_mode='remote'
Using `KubeCluster(deploy_mode='remote')`, the `scheduler_timeout` kwarg seems to define the amount of time following any active requests from the client before which the scheduler closes.

For example, if I call:
```
cluster = KubeCluster(deploy_mode='remote', scheduler_timeout='10s')
client = Client(cluster)
future = client.submit(some_function,some_args)
```

The scheduler will close (and the client will close) 10s after that call to `submit`. If I keep making calls to `submit`, however, then the scheduler and client will keep staying alive.

If I understand this kwarg correctly, it should not have this behavior (and, indeed it doesn't with a LocalCluster). My expected behavior is that unless the client dies (e.g. from restarting the python kernel), it would still be connected to the scheduler and this timeout should not trigger.

bolliger32:
(partially) fixes #206 

- changes the kwarg `scheduler_timeout` to `idle_timeout` to match the actual behavior of this kwarg
- passes the configuration of this setting to `distributed.yaml`, rather than having a separate value in `kubernetes.yaml`

The question still remains of whether there should be an implementation of the original intended behavior of `scheduler_timeout` (i.e. to timeout after a certain amount of time not having been connected to a client). And, if so, should the pod containing the scheduler also be killed if the scheduler dies.

### DIFF
--- a/dask_kubernetes/kubernetes.yaml
+++ b/dask_kubernetes/kubernetes.yaml
@@ -7,7 +7,7 @@ kubernetes:
   host: "0.0.0.0"
   port: 0
   env: {}
-  scheduler-timeout: "5 minutes" # Length of inactivity to wait before closing the cluster
+  idle-timeout: null # Length of inactivity to wait before closing the cluster
   deploy-mode: "local"
   interface: null
   protocol: "tcp://"
--- a/dask_kubernetes/kubernetes.yaml
+++ b/dask_kubernetes/kubernetes.yaml
@@ -7,7 +7,6 @@ kubernetes:
   host: "0.0.0.0"
   port: 0
   env: {}
-  idle-timeout: null # Length of inactivity to wait before closing the cluster
   deploy-mode: "local"
   interface: null
   protocol: "tcp://"
--- a/dask_kubernetes/kubernetes.yaml
+++ b/dask_kubernetes/kubernetes.yaml
@@ -7,6 +7,7 @@ kubernetes:
   host: "0.0.0.0"
   port: 0
   env: {}
+  idle-timeout: null
   deploy-mode: "local"
   interface: null
   protocol: "tcp://"
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -141,10 +141,7 @@ class Scheduler(Pod):
         self.pod_template.metadata.labels["dask.org/component"] = "scheduler"
         cli_args = ["dask-scheduler"]
         if self._idle_timeout is not None:
-            cli_args += [
-                "--idle-timeout",
-                self.__idle_timeout
-            ]
+            cli_args += ["--idle-timeout", self.__idle_timeout]
         self.pod_template.spec.containers[0].args = cli_args
 
     async def start(self, **kwargs):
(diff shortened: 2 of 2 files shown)


dask_kubernetes/kubernetes.yaml
```diff
@@ -7,7 +7,6 @@ kubernetes:
   host: "0.0.0.0"
   port: 0
   env: {}
-  scheduler-timeout: "5 minutes" # Length of inactivity to wait before closing the cluster
```
jacobtomlinson: Could you replace this with an `idle-timeout` option?

jacobtomlinson: I'm unsure about grabbing the config option from the distributed settings. It may cause confusion not having it in the kubernetes config.

bolliger32: sounds good. I hadn't realized that the `dask_kubernetes.Scheduler` class bypasses `distributed.Scheduler` completely (I think?), so agree this makes sense.

bolliger32:
> Thanks for adding this. It should defintely clear things up.
> 
> Would you be happy to also add the `scheduler_timeout` option and hook it up to the `scheduler_timeout` in distributed so we get the correct functionality?
> 
> The problem was that I had wired `scheduler_timeout` in dask-kubernetes to `idle_timout` in distributed. But both options exist and it would be great to expose both to the user.

My understanding from @mrocklin 's comment was that `scheduler_timeout` actually isn't implemented in `distributed` at the moment. Is that correct? If it is, would be easy to wire to that command. If not, might take a little more thinking on where exactly to put this flag (`dask-kubernetes` vs. `distributed`)

jacobtomlinson:
Ah you're right dask/distributed#2346 is still open. Sorry about the confusion.

I guess this helps the use case for scheduler timeouts when clients are not connected though.

bolliger32:
true. One other minor challenge is that the string parsing of the original default `scheduler_timeout` (what is now `idle_timeout`) meant that you couldn't pass 2-word strings. So the default "5 minutes" caused an error because the quotes didn't make it to the CLI `dask-scheduler` call. I'm not sure if this is super important to fix - maybe in the docstring we just tell users that they must specify it as a "number of seconds"? Although it probably wouldn't be too hard to figure out what combination of escape characters would make the string quotes get all the way to the CLI call

jacobtomlinson:
It might be nice to fix that. But we can do that in a different PR if you like.

jacobtomlinson:
It looks like CI is just failing on Black now. Can you please run `black` on the module and push again?

bolliger32:
i did and it adjusted a couple things, but its still failing. Maybe the black config is different on my machine? I could try setting --line-length=79? That would reformat parts that I didn't touch.

jacobtomlinson:
Strange! I've just run it myself with default settings and it reformatted one file. I've pushed it into this PR for you.

bolliger32:
hmm. maybe it was a different black version and they changed their defaults? thanks!

jacobtomlinson:
Thanks for your help!

PR #209, opened 2019-12-03
Add timeout settings to wait for the remote scheduler service

Timost:
When using cloud providers load balancers (in my case Azure) for the remote scheduler service it takes more than 30 secs for the service to get an IP accessible from outside the cluster.

This PR adds a new setting in the dask-kubernetes yml config file to set this timeout to a given value in seconds.

This is my first PR on the project so this will likely need some extra work.

### DIFF
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -129,13 +129,14 @@ class Scheduler(Pod):
     ----------
     idle_timeout: str, optional
         The scheduler task will exit after this amount of time
-        if there are no requests from the client. Default is to 
+        if there are no requests from the client. Default is to
         never timeout.
     """
 
     def __init__(self, idle_timeout: str, **kwargs):
         super().__init__(**kwargs)
         self.service = None
+        self._service_wait_timeout_s = None
         self._idle_timeout = idle_timeout
 
         self.pod_template.metadata.labels["dask.org/component"] = "scheduler"
@@ -167,7 +168,7 @@ class Scheduler(Pod):
             # Wait for load balancer to be assigned
             start = time.time()
             while self.service.status.load_balancer.ingress is None:
-                if time.time() > start + 30:
+                if time.time() > start + self._service_wait_timeout_s:
                     raise TimeoutError(
                         "Timed out waiting for Load Balancer to be provisioned."
                     )
@@ -205,6 +206,9 @@ class Scheduler(Pod):
             self.service_template.spec.type = dask.config.get(
                 "kubernetes.scheduler-service-type"
             )
+        self._service_wait_timeout_s = (
+            dask.config.get("kubernetes.scheduler-service-wait-timeout", 30) or 30
+        )
         await self.core_api.create_namespaced_service(
             self.namespace, self.service_template
         )
@@ -265,7 +269,7 @@ class KubeCluster(SpecCluster):
         ``[InCluster(), KubeConfig()]``.
     idle_timeout: str (optional)
         The scheduler task will exit after this amount of time
-        if there are no requests from the client. Default is to 
+        if there are no requests from the client. Default is to
         never timeout.
     deploy_mode: str (optional)
         Run the scheduler as "local" or "remote".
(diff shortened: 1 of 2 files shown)


dask_kubernetes/core.py
```diff
@@ -205,6 +206,9 @@ def __init__(self, idle_timeout: str, **kwargs):
             self.service_template.spec.type = dask.config.get(
                 "kubernetes.scheduler-service-type"
             )
+        self._service_wait_timeout_s = (
+            dask.config.get("kubernetes.scheduler-service-wait-timeout", 30) or 30
+        )
```
jacobtomlinson: We should move this to the [`KubeCluster`](https://github.com/dask/dask-kubernetes/blob/master/dask_kubernetes/core.py#L381) object and allow folks to pass it as a kwarg there too.

You would then add it to the scheduler spec [here](https://github.com/dask/dask-kubernetes/blob/master/dask_kubernetes/core.py#L489) which would then get passed into the `Scheduler` object [here](https://github.com/dask/dask-kubernetes/blob/master/dask_kubernetes/core.py#L136).

Timost: I actually started by doing what you propose, but seeing the already numerous kwargs in the `KubeCluster` constructor, I ended up adding it only in the configuration since I wasn't sure about adding yet another argument.

To be sure I understood correctly. Should I move everything entirely or should I add the "constructor way" to set it in addition to the possibility of using the config ?

jacobtomlinson: Yes we should use the constructor and config for all options. This class has a lot of kwargs and we will certainly add more in the future. It's better to be consistent.

Timost: Ok sounds good to me. I'll make the modifications.

Timost:
PTAL

dask_kubernetes/core.py
```diff
@@ -131,18 +131,26 @@ class Scheduler(Pod):
...
         self.pod_template.metadata.labels["dask.org/component"] = "scheduler"
         cli_args = ["dask-scheduler"]
         if self._idle_timeout is not None:
-            cli_args += ["--idle-timeout", self.__idle_timeout]
+            cli_args += ["--idle-timeout", self._idle_timeout]
+
+        if self._service_wait_timeout_s is None:
+            self._service_wait_timeout_s = (
+                dask.config.get("kubernetes.scheduler-service-wait-timeout", 30) or 30
+            )
```
jacobtomlinson: I would prefer this to be consistent with other options and also moved to the `_start` method.

I would expect to see something there like:

```python
self._scheduler_service_wait_timeout = self._scheduler_service_wait_timeout or dask.config.get("kubernetes.scheduler-service-wait-timeout")
```

We also do not set our default values in code as the Dask config library will draw default values from the [example config file](https://github.com/dask/dask-kubernetes/blob/master/dask_kubernetes/kubernetes.yaml). So if you could add the default of `30` there I would appreciate it.

Timost: I think I already set the default value in the example config file. I'll move the parsing to `KubeCluster._start` though.

If someone sets the `scheduler-service-wait-timeout` to null in the config, it will crash as it is though.

jacobtomlinson: Thanks for making the change.

I feel like if someone sets the timeout to `null` then it should wait forever. Perhaps we should implement this and then document the fact that it's probably not a great idea to do that.

Timost: Sure. Where would that go in terms of documentation ? In the `Scheduler` docstring or somewhere else ?

jacobtomlinson: I would put it in the `KubeCluster` docstring for now with the new kwarg.

Timost: Hum okay but you can't really set it to None from the KubeCluster constructor since it already defaults to None. This "option" will only be available when using the config file.

Timost: Maybe we can say that the "no-timeout" value is `-1` and not `None` this way we can pass it also in the `KubecCluster` constructor.

jacobtomlinson: I think `0` is a common value for disabling timeouts like this.

Timost: Ok I'll go with `0` then !

Timost:
PTAL

PR #211, opened 2019-12-04
Add optional separate scheduler deployment

### LINKED ISSUE #208: Configure scheduler pod separately from worker pods
Hi,
First of all thank you for working on dask integration with kubernetes !

I have a need that I think I can't fulfil at the moment with dask-kubernetes. My use case is the following:

I want to save money by using "pre-emptible" nodes (or low-priority in azure) for my worker pods.
However since these nodes can be shot at any time I want to put the scheduler pod on another node pool which is guaranteed to be stable, this can easily be achieved in kubernetes with taints/tolerations/affinities and I've been able to do it by working with kubernetes deployments directly. Using dask-kubernetes would be nice though because I could potentially have auto-scaling from dask directly and also easily submit jobs from dev computers (with dynamic dependencies) etc...

However, looking at the code, it seems I can't pass one deployment definition (aka `pod_template`) for the scheduler and another one for the worker pods.

Could you confirm this is not possible yet ? I'm willing to spend some time to implement this if you consider this is important/needed. Maybe adding an optional `scheduler_template` arg to the `KubeCluster` constructor, although there are already a lot of arguments...

Timost:
This can be usefull when one wants to set different affinities for the
scheduler and the worker pods.

My personal use case is to take advantage of premptible nodes for dask workers but keep the scheduler on a guaranteed node.

Attempts to fix #208.

### DIFF
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -142,13 +142,11 @@ class Scheduler(Pod):
         self.service = None
         self._idle_timeout = idle_timeout
         self._service_wait_timeout_s = service_wait_timeout_s
-
-        self.pod_template.metadata.labels["dask.org/component"] = "scheduler"
-        cli_args = ["dask-scheduler"]
         if self._idle_timeout is not None:
-            cli_args += ["--idle-timeout", self._idle_timeout]
-
-        self.pod_template.spec.containers[0].args = cli_args
+            self.pod_template.spec.containers[0].args += [
+                "--idle-timeout",
+                self._idle_timeout,
+            ]
 
     async def start(self, **kwargs):
         await super().start(**kwargs)
@@ -255,6 +253,9 @@ class KubeCluster(SpecCluster):
     ----------
     pod_template: kubernetes.client.V1Pod
         A Kubernetes specification for a Pod for a dask worker.
+    scheduler_pod_template: kubernetes.client.V1Pod (optional)
+        A Kubernetes specification for a Pod for a dask scheduler.
+        Defaults to the pod_template.
     name: str (optional)
         Name given to the pods.  Defaults to ``dask-$USER-random``
     namespace: str (optional)
@@ -367,9 +368,11 @@ class KubeCluster(SpecCluster):
         dashboard_address=None,
         security=None,
         scheduler_service_wait_timeout=None,
+        scheduler_pod_template=None,
         **kwargs
     ):
         self.pod_template = pod_template
+        self.scheduler_pod_template = scheduler_pod_template
         self._generate_name = name
         self._namespace = namespace
         self._n_workers = n_workers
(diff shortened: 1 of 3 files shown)


dask_kubernetes/core.py
```diff
@@ -393,6 +396,73 @@ def __init__(
         self.kwargs = kwargs
         super().__init__(**self.kwargs)
 
+    def _clean_pod_templates(self):
+        for pod_template, pod_type in [
+            (self.pod_template, "worker"),
+            (self.scheduler_pod_template, "scheduler"),
+        ]:
```
jacobtomlinson: I wonder if you could call this function twice and pass these in as args. Would make it more readable.

dask_kubernetes/core.py
```diff
@@ -393,6 +396,73 @@ def __init__(
...
+
+        if not self.scheduler_pod_template:
+            self.scheduler_pod_template = copy.deepcopy(self.pod_template)
+            self.scheduler_pod_template.spec.containers[0].args = ["dask-scheduler"]
+
+        self.pod_template = clean_pod_template(self.pod_template, pod_type="worker")
+
+        self.scheduler_pod_template = clean_pod_template(
+            self.scheduler_pod_template, pod_type="scheduler"
+        )
```
jacobtomlinson: As we've broken out into a helper function I would prefer to return these values rather than update the object.

dask_kubernetes/core.py
```diff
@@ -393,6 +396,73 @@ def __init__(
...
+            pod_template.metadata.namespace = self._namespace
+
+            if self.env:
+                pod_template.spec.containers[0].env.extend(
+                    [
+                        kubernetes.client.V1EnvVar(name=k, value=str(v))
+                        for k, v in self.env.items()
+                    ]
+                )
+            pod_template.metadata.generate_name = self._generate_name
```
jacobtomlinson: Again I would prefer to return the template rather than overwrite values.

dask_kubernetes/core.py
```diff
@@ -425,33 +495,8 @@ def __init__(
...
-            d = dask.config.expand_environment_variables(d)
-            self.pod_template = make_pod_from_dict(d)
-
-        if not self.pod_template:
-            msg = (
-                "Worker pod specification not provided. See KubeCluster "
-                "docstring for ways to specify workers"
-            )
-            raise ValueError(msg)
+        self._clean_pod_templates()
```
jacobtomlinson: This function is a bit **magic**. It's not obvious what the result of calling it is. I would prefer a direct assignment for clarity.

dask_kubernetes/core.py
```diff
@@ -464,28 +509,14 @@ def __init__(
...
-
-        if self.env:
-            self.pod_template.spec.containers[0].env.extend(
-                [
-                    kubernetes.client.V1EnvVar(name=k, value=str(v))
-                    for k, v in self.env.items()
-                ]
-            )
-        self.pod_template.metadata.generate_name = self._generate_name
+        self._fill_pod_templates()
```
jacobtomlinson: Again I see the need for breaking this out into a function but it's a bit magic.

dask_kubernetes/core.py
```diff
@@ -393,6 +396,73 @@ def __init__(
         self.kwargs = kwargs
         super().__init__(**self.kwargs)
 
+    def _clean_pod_templates(self):
```
jacobtomlinson: Can we have a docstring here?

dask_kubernetes/core.py
```diff
@@ -393,6 +396,73 @@ def __init__(
...
+            self.scheduler_pod_template = copy.deepcopy(self.pod_template)
+            self.scheduler_pod_template.spec.containers[0].args = ["dask-scheduler"]
+
+        self.pod_template = clean_pod_template(self.pod_template, pod_type="worker")
+
+        self.scheduler_pod_template = clean_pod_template(
+            self.scheduler_pod_template, pod_type="scheduler"
+        )
+
+    def _fill_pod_templates(self):
```
jacobtomlinson: Can we have a docstring here?

dask_kubernetes/core.py
Timost: Yeah I see what you mean ! I'll make the modifications.

dask_kubernetes/core.py
Timost: As far as I can tell this requires to `copy.deepcopy` the template. Is that okay with you ?

dask_kubernetes/core.py
Timost: The thing though is that there is a bit of logic to assign the scheduler template to the pod template by default. And this requires having knowledge of both templates. I can move this part out of the function, but I fell it belongs more there than in the main `_start` method.

jacobtomlinson: I think my preference would be to have it in the main `_start` function. It makes the start function clearer to read.

dask_kubernetes/core.py
jacobtomlinson: Sure.

dask_kubernetes/core.py
Timost: Ok I see what you mean. I'll change it.

dask_kubernetes/core.py
Timost: Maybe `_fill_pod_templates` and `_clean_pod_templates` (or there equivalent) should go in `objects.py` what do you think ?

jacobtomlinson: I'm not sure if that would add any benefit. It just adds further indirection.

Timost: Ok I'll leave them there then.

Timost:
@jacobtomlinson Could you give me a few pointers on how to run the tests on my computer ? looking at `.circleci/config.yml` it looks like I need to install minikube and run the tests against a k8s cluster inside minikube.

I'm using Debian.

jacobtomlinson:
Check out the testing section of the docs.

https://kubernetes.dask.org/en/latest/testing.html

Timost:
Oh thanks. I missed that sorry !

Timost:
PTAL:
- Changed methods style (return new values for assignement) thanks for the nice link.
- Added doc strings (not sure if this is enough)
- Fixed the code so the tests pass (at least on my computer)

jacobtomlinson:
That's great many thanks for making these changes!

Timost:
Thank you for the very responsive review ! It's really nice !

PR #214, opened 2019-12-20
Raise exception if retries are exhausted

### LINKED ISSUE #212: Reraise ApiError when all retries fail
See the thread starting at https://github.com/pangeo-data/pangeo-stacks/issues/110#issuecomment-567590381.

In Pod.start we try to start the pod 10 times

https://github.com/dask/dask-kubernetes/blob/master/dask_kubernetes/core.py#L66-L74

If all 10 fail, we currently return nothing and return control to SpecCluster, which asserts that the pod state is set to running (which is hasn't) and gives an uninformative AssertionError. In that case it was an issue with the pod name (separate issue for that).

jacobtomlinson:
Fixes #212

### DIFF
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -64,14 +64,19 @@ class Pod(ProcessInterface):
         return self.pod_template.metadata.labels["dask.org/cluster-name"]
 
     async def start(self, **kwargs):
-        for _ in range(10):  # Retry 10 times
+        retry_count = 0  # Retry 10 times
+        while True:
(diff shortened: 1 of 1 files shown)


jacobtomlinson:
Hmm that might be a good idea.

I seem to remember seeing lots of spurious errors here, so I don't think surfacing them to the user unless it fails consistently would be useful. So I've added a debug line in case we need to investigate further in future.

TomAugspurger:
Sounds good!

Issue #216, opened 2020-01-07
Cluster's `worker_spec` not updated when a worker pod is killed


ig248:
Killing a worker pod (e.g. using `kubectl delete pod`) does not update the internal state of the cluster. Thus, (auto-)scaling behaviour gets out of sync.

This issue is likely to occur during a host node failure etc.

ig248:
This appears to be a deeper problem with `ProcessInterface` not providing an interface for a liveliness probe, and `SpecCluster.workers` holding on to potentially dead worker instances. Any thoughts much appreciated!

jacobtomlinson:
Looks like this was explored upstream in dask/distributed#2968. 

My understanding from that issue is that we need to implement a `_update_worker_status` method in the `KubeCluster` class which will be used to periodically check which workers are alive.

@mrocklin @guillaumeeb could you confirm this?

ig248:
@jacobtomlinson Thank you, it looks doable, though in the meantime I have created an alternative solution that makes use of k8s-native `Deployment`: https://github.com/octoenergy/dask-remote/pull/1

ig248:
@jacobtomlinson looking at `SpecCluster._update_worker_status()`, it appears that lost worker removal should be correctly inherited by `KubeCluster` - I may need to do additional testing.

jacobtomlinson:
@ig248 thanks for the info. For more information on why we don't use `Deployment` resources see https://github.com/dask/dask-kubernetes/issues/168#issuecomment-518587280.

PR #217, opened 2020-01-09
Set the name of each worker to allow tracking worker removals.

ig248:
This PR allows the `KubeCluster` to maintain the number of worker pods in adaptive mode, even if pods are lost/deleted externally.

This PR fixes https://github.com/dask/dask-kubernetes/issues/216

## Before this change
`KubeCluster._update_worker_status` was not working correctly. The name found in `KubeCluster.scheduler_info` defaulted to worker address, and was thus different from the corresponding key in `KubeCluster.workers`.

## After this change
The auto-generated worker name is handled correctly by the `Worker` and is appended to `args` in the Pod spec.

### Other improvements
- increas scope of slow test fixtures for faster tests
- remove `tornado` dependency in favour af native `asyncio` as not needed with `python>=3.5`

### DIFF
--- a/dask_kubernetes/tests/test_async.py
+++ b/dask_kubernetes/tests/test_async.py
@@ -440,11 +440,18 @@ async def test_reject_evicted_workers(cluster):
         ),
     )
 
-    # Wait until pod is evicted
+    # Wait until worker removal has been picked up by scheduler
     start = time()
     while len(cluster.scheduler_info["workers"]) != 0:
-        await gen.sleep(0.1)
-        assert time() < start + 60
+        delta = time() - start 
+        assert delta < 60, f"Scheduler failed to remove worker in {delta:.0f}s"
+        await asyncio.sleep(0.1)
+
+    # Wait until worker removal has been handled by cluster
+    while len(cluster.workers) != 0:
+        delta = time() - start 
+        assert delta < 60, f"Cluster failed to remove worker in {delta:.0f}s"
+        await asyncio.sleep(0.1)
 
 
 @pytest.mark.asyncio
--- a/dask_kubernetes/tests/test_async.py
+++ b/dask_kubernetes/tests/test_async.py
@@ -1,4 +1,5 @@
 import asyncio
+
 import base64
 import getpass
 import os
@@ -9,7 +10,6 @@ import yaml
 
 import kubernetes_asyncio as kubernetes
 import pytest
-from tornado import gen
 
 import dask
 from dask.distributed import Client, wait
@@ -113,7 +113,7 @@ async def test_basic(cluster, client):
     assert result == 11
 
     while len(cluster.scheduler_info["workers"]) < 2:
-        await gen.sleep(0.1)
+        await asyncio.sleep(0.1)
 
     # Ensure that inter-worker communication works well
     futures = client.map(lambda x: x + 1, range(10))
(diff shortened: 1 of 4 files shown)


jacobtomlinson:
This is great many thanks!

We currently have a test to check that workers are evicted correctly.

https://github.com/dask/dask-kubernetes/blob/master/dask_kubernetes/tests/test_async.py#L423-L448

Would you be interested in updating that test to also check that worker removals are tracked correctly?

ig248:
Yes will do, thanks for pointing out the relevant test.

ig248:
@jacobtomlinson IMHO this PR is ready to review/merge

dask_kubernetes/tests/test_sync.py
```diff
@@ -318,7 +316,7 @@ def test_scale_up_down(cluster, client):
 
     a, b = list(cluster.scheduler_info["workers"])
     x = client.submit(np.ones, 1, workers=a)
-    y = client.submit(np.ones, 50_000, workers=b)
+    y = client.submit(np.ones, 50000, workers=b)
```
jacobtomlinson: This was set intentionally using the underscore notation for readability. Is there a reason why you changed it?

dask_kubernetes/tests/test_async.py
```diff
@@ -456,12 +461,12 @@ def pod_spec(image_name):
...
     start = time()
     while len(cluster.scheduler_info["workers"]) != 2:
-        await gen.sleep(0.1)
+        await asyncio.sleep(0.1)
         assert time() < start + 20
 
     a, b = list(cluster.scheduler_info["workers"])
     x = client.submit(np.ones, 1, workers=a)
-    y = client.submit(np.ones, 50_000, workers=b)
+    y = client.submit(np.ones, 50000, workers=b)
```
jacobtomlinson: This was set intentionally using the underscore notation for readability. Is there a reason why you changed it?

dask_kubernetes/tests/test_async.py
```diff
@@ -40,31 +41,29 @@ def pod_spec(image_name):
...
     ns = kubernetes.client.V1Namespace(
         metadata=kubernetes.client.V1ObjectMeta(name=name)
     )
     await api.create_namespace(ns)
-    try:
-        yield name
-    finally:
-        await api.delete_namespace(name)
+    yield name
+    await api.delete_namespace(name)
```
jacobtomlinson: What would happen if we raise while yielding the name? Can that even happen? Would we still clean up?

dask_kubernetes/tests/test_sync.py
ig248: Oops, may have to do with my black settings - I will check!

dask_kubernetes/tests/test_async.py
ig248: ditto

dask_kubernetes/tests/test_async.py
ig248: Yes, pytest yield fixtures act as context managers and execute everything after `yield` as teardown - [docs here](https://docs.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code)

jacobtomlinson: TIL 😄

jacobtomlinson:
(And then I accidentally didn't squash and merge! My bad 😓)

PR #219, opened 2020-01-11
Run all tests dockerized *and* podified

ig248:
This PR address simplicity and reproducibility of the environment setup for both local development and CI testing. In particular, two aspects of the development environment caused me some difficulty initially:
  1. python environment: this could be addressed with `conda` or `pyenv` or...., but platform-specific issues with binary dependencies will always exist
  2. cluster and *networking* environment: [current instructions](https://kubernetes.dask.org/en/latest/testing.html) are not sufficiently reproducible as network setups vary wildly.

This PR addresses the two issues in a unified way:
  1. containerize the test environment and use `docker run` for **linting** and **unit tests**
  2. deploy a test pod **insude the K8S cluster** by using `kubectl run` to run **integration tests**

This approach works equally well in CI and is facilitated by reusing the same `make` commands.

### `conftest.py` and fixtures
- test no longer create/destroy a new namespace
- pytest will now accept `--namespace`, `--context`, and `--in-cluster` args, e.g.
```
$ py.test dask_kubernetes --context minikube --namespace test-namespace
```
- a new `Makefile` provides largerly self-documenting build scripts
- a new Docker image is provided with `make` as an ENTRYPOINT corresponding to the Makefile

### Running `make` commands
The new make commands can be run locally, in docker, or in a kubernetes pod, e.g.
```
$ make test  # run locally
$ make docker-make COMMAND=test  # run in docker
$ make k8s-make COMMAND=test  # run in a Pod
```
The transition is basically seamless.

## Backwards compatibility
The only potentially breaking change to existing workflow is the reliance on an existing namespace. `conftest.py` will default to the `default` namespace, so `py.test dask_kubernetes` still works.
 
## ToDo:
- [x] make sure CircleCI runs...
- [x] document changes in the PR description
- [x] document usage for local development, as in [existing instructions](https://kubernetes.dask.org/en/latest/testing.html)
- [x] remove [WIP] tag

## Future
A key issue with current test suite is that all tests are basically functional/integration tests. Even `test_objects.py` implicitly relies on the external environment via kube config. Refactoring some of the code will be necessary to allow test isolation and faster unit tests.


Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM daskdev/dask:latest
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    make \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /opt/data_apps
```
jacobtomlinson: My preference would be `/usr/local/src` as this is a more common source location.

ig248: > This looks like a great change and should make our testing environment more consistent.
Thank you!

>     * Switching to kind introduces Docker as a test dependency. Minikube does not have the same requirement. This should be noted somewhere.

My understanding is that `minikube start --vm-driver=none` implicitly uses host's docker - it comes pre-installed on CircleCI machine

- [ ] Though I have left the `minikube` make commands if you prefer to switch back?

>     * You've removed the CI conda environment in favour of the `daskdev/dask:latest` image. These are not equivalent. Specifically the CI environment installs `dask` and `distributed` from master, which allows `dask-kubernetes` to be tested against unreleased versions and for the libraries to vbe developed in parallel. Therefore `requirements-dev.txt` should probably include those.

- Thanks, I will make sure to update it accordingly
  - [ ] add dask@master and distributed@master to `requirements-dev.txt`

ig248:
> This looks like a great change and should make our testing environment more consistent.
Thank you!

>     * Switching to kind introduces Docker as a test dependency. Minikube does not have the same requirement. This should be noted somewhere.

My understanding is that `minikube start --vm-driver=none` implicitly uses host's docker - it comes pre-installed on CircleCI machine

- [ ] Though I have left the `minikube` make commands if you prefer to switch back?

>     * You've removed the CI conda environment in favour of the `daskdev/dask:latest` image. These are not equivalent. Specifically the CI environment installs `dask` and `distributed` from master, which allows `dask-kubernetes` to be tested against unreleased versions and for the libraries to vbe developed in parallel. Therefore `requirements-dev.txt` should probably include those.

- Thanks, I will make sure to update it accordingly
  - [ ] add dask@master and distributed@master to `requirements-dev.txt`

jacobtomlinson:
> My understanding is that minikube start --vm-driver=none implicitly uses host's docker - it comes pre-installed on CircleCI machine

Sure, but for local testing it is common to use minikube in VM mode and setting this up is covered in the minikube docs. Kind does not have this option, the docs state that docker is required before installing.

This was more a note for when you document local development rather than a change to the CI.

Dockerfile
ig248: :+1:

ig248:
I have extended "Testing" documentation, though trying to cover all possible scenarios made it a bit verbose..

doc/source/testing.rst
```diff
@@ -1,51 +1,124 @@
...
-    If you get an error message like the following::
+Set up a local Kubernetes cluster
+---------------------------------
+We can set up a kubernetes cluster on the local machine using either
+`minikube <https://minikube.sigs.k8s.io/>`_ or `kind <https://kind.sigs.k8s.io/>`_
 
-       RTNETLINK answers: File exists
+``minikube``
+^^^^^^^^^^^^
+1. Install ``minikube`` by following online instructions, or use::
```
jacobtomlinson: Can you link to instructions.

doc/source/testing.rst
```diff
@@ -1,51 +1,124 @@
...
 
-    it most likely means you have docker running on your host using the same
-    IP range minikube is using. You can fix this by editing your
-    ``/etc/docker/daemon.json`` file to add the following:
+      make kubectl-bootstrap
+2. Start ``minikube`` with::
 
-    .. code-block:: json
+      minikube start
+3. (recommended) to run containerized tests in a pod, make sure you have ``docker`` installed and run ``eval $(shell minikube docker-env)`` before building the Docker image
```
jacobtomlinson: Can you link to docker install instructions.

doc/source/testing.rst
```diff
@@ -1,51 +1,124 @@
...
+JSON.
+
+Once edited, restart docker with ``sudo systemctl restart docker``. It
+should come up using a different IP range, and you can run the
+``sudo ip route add`` command again. Note that restarting docker will
+restart all your running containers by default.
+
+``kind``
+^^^^^^^^
+1. Install ``kind`` by following online instructions, or use::
```
jacobtomlinson: Can you link to kind install instructions.

jacobtomlinson: Also note that kind requires docker.

doc/source/testing.rst
ig248: :+1:

doc/source/testing.rst
ig248: :+1:

doc/source/testing.rst
ig248: :+1: :+1:

ig248:
For some reason `kubectl run` started timing out on CircleCI...

jacobtomlinson:
`dask_kubernetes/tests/test_async.py::test_cluster_create` seems to just hang forever.

ig248:
> `dask_kubernetes/tests/test_async.py::test_cluster_create` seems to just hang forever.

I set up CI on the fork, hopefully it will help me debug and fix it. It definitely passes on my local (that's what they all say) and I have seen it pass in Circle...

ig248:
@jacobtomlinson I have done some bisecting and worryingly, upgrading from `daskdev/dask:latest` to `dask@master` and `distributed@master` in [this commit](daskdev/dask:latest) causes the worker to start timing out trying to connect to the scheduler. Has an issue been introduced upstream? Worrying indeed

jacobtomlinson:
Hmm that is frustrating. Looking at your table it seems to be `distributed@master` that is causing the trouble.

dask_kubernetes/tests/conftest.py
```diff
@@ -1,13 +1,52 @@
...
+import kubernetes
+
+from dask_kubernetes.auth import ClusterAuth, KubeConfig, InCluster
+
 
 def pytest_addoption(parser):
-    parser.addoption("--worker-image", help="Worker image to use for testing")
+    parser.addoption(
+        "--worker-image",
+        default="daskdev/dask:dev",
```
jacobtomlinson: This should be `daskdev/dask:latest`. And really it should be a custom image which installs `dask` and `distributed` from master.

Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM daskdev/dask:2.9.0
```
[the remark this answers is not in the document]
jacobtomlinson: Also this should probably be `daskdev/dask:latest`

jacobtomlinson:
FYI you can have the `daskdev/dask:latest` image install extra packages at runtime by setting the `EXTRA_PIP_PACKAGES` environment variable.

e.g

```
docker run --rm -e "EXTRA_PIP_PACKAGES=git+https://github.com/dask/dask@master git+https://github.com/dask/distributed@master" -p 8787:8787 -p 8786:8786 daskdev/dask:latest dask-scheduler
```

dask_kubernetes/tests/conftest.py
ig248: Good point, there is a mismatch between worker and "tester" image. I will address that - in fact, with layer sharing on the `kind` node it should be faster too

ig248:
Thanks, `EXTRA_PIP_PACKAGES` can be useful but slow down worker start-up siginificantly, I wouldn't want to add this in tests

ig248:
Right, the offending commit is this: https://github.com/dask/distributed/commit/32cb96effe9287fdc42dd73be36989be13abe99d

ig248:
Created issue here: https://github.com/dask/distributed/issues/3379

Will build worker image @ master for this PR

ig248:
using custom worker image now so everything is @ master. Is there a reason to keep the `docker/Dockerfile`?

jacobtomlinson:
@ig248 the commit you refer to is backward incompatible with previous versions on distributed. Which makes me think that not everything is running `distributed@master` in our tests, such as the workers.

ig248:
@jacobtomlinson see comment above. Everything is now at master and passing.

jacobtomlinson:
Ah whoops I didn't see that comment. Also as you keep force pushing I can't see the diff between commits, so it's hard to track what is changing here.

ig248:
@jacobtomlinson I culled the Dockerfiles. There's still the `docker/Dockerfile`- do you have a use for that?

jacobtomlinson:
No I think we can clear the whole `docker/` directory now.

ig248:
> No I think we can clear the whole `docker/` directory now.

Gone

jacobtomlinson:
Thanks for bearing with me on this, was a relatively long review.

ig248:
Thank you @jacobtomlinson ! Now on to replace some integration tests with unit tests...

PR #223, opened 2020-01-19
Towards unit test coverage

ig248:
This PR moves parts of the existing test suite to unit tests that can be run locally (or dockerized) without needing a live K8S cluster.

The motivation is that
- debugging async code inside integration tests is complex
- unit tests are much faster and speed up development/CI
- having unit tests in place should encourage new contributions to include specific unit tests

To enable synchronous unit testing, template rendering inside `KubeCluster` is refactored by creating properties `KubeCluster.rendered_worker_pod_template` and `KubeCluster.rendered_scheduler_pod_template`.  IMHO this makes the implementation cleaner, and could be applied similarly to other parts of the code in the future.

Unit tests can be run locally, dockerized, or in a pod:
```
make test TEST_KEYWORD=unit  # or simply `make test`
docker-make COMMAND=test TEST_KEYWORD=unit  # also aliased to `make ci-unit`
k8s-make COMMAND=test TEST_KEYWORD=unit
```

Integration tests are run in a pod as before:
```
k8s-make COMMAND=test TEST_KEYWORD=integration  # also aliased to `make ci-integration`
```

Thus, CI now runs three commands (each of which also works locally):
```
make ci-lint
make ci-unit
make ci-integration
```


dask_kubernetes/core.py
```diff
@@ -388,19 +386,19 @@ def __init__(
...
-        self._scheduler_service_wait_timeout = scheduler_service_wait_timeout
+        self.__generate_name = name
+        self.__namespace = namespace
+        self.n_workers = n_workers
+        self.idle_timeout = idle_timeout
+        self.deploy_mode = deploy_mode
+        self.protocol = protocol
+        self.interface = interface
+        self.dashboard_address = dashboard_address
+        self.scheduler_service_wait_timeout = scheduler_service_wait_timeout
```
jacobtomlinson: You've made a number of private attributes public here, what is the rationale behind this?

dask_kubernetes/core.py
```diff
@@ -415,7 +413,104 @@ def __init__(
...
+    def _n_workers(self):
+        return (
+            self.n_workers
+            if self.n_workers is not None
+            else dask.config.get("kubernetes.count.start")
+        )
+
+    @property
+    def _idle_timeout(self):
+        return self.idle_timeout or dask.config.get("kubernetes.idle-timeout")
```
jacobtomlinson: Moving these attributes to properties seems to add a large amount of [indirection](https://matthewrocklin.com/blog/work/2019/06/23/avoid-indirection), re-execution and complexity here. What is the reasoning for this?

dask_kubernetes/core.py
ig248: I find it is good practice to leave __init__ kwargs "as they are". Trying to hide anything away from the end user doesn't work in python anyway... Of course we can make these `_` or `__` if you prefer, but generally users expect `MyClass(key=value).key` to return the same value.

dask_kubernetes/core.py
ig248: - python properties effectively make (parts of) the code *stateless* and *functional*. Anything stateful is harder to unit test, debug and maintain than pure functional code. IMHO the *state* of the object should be described by the __init__ kwargs (or as close as possible) whenever possible.

- if re-execution performance becomes an issue, property getters can easily be cached with `lru_cache` (or an explicit internal `__idle_timeout` that's set on first access).

While it adds a little boilerplate, it allows us to (for example) add a unit test to check that `"kubernetes.idle-timeout"` is used as default *in isolation from other code paths*

ig248:
> However I would like to understand more about when you would expect folks to purely run unit tests without the integration tests?

Generally while working on a feature or bug locally, I like to run the *relevant unit test* only, before running the complete test suite - which will still be the case for any PR. Not being able to test an individual part in isolation makes development harder, and is often indicative of how maintainable the code will be in the future.

> I also see there has been a large amount of refactoring of the `_start` method here where config items have been pulled out into properties and many attributed have been made public. Is this for testing purposes?

I would say it is not purely for "testing purposes", but for *testability* and *maintainability*, which go hand-in-hand. Basically, being able to work on changing template behavior *in isolation* from even loops etc. can IMHO only make life easier.

dask_kubernetes/core.py
ig248: I would say that IMHO it falls under ["functions are a good idea"](https://matthewrocklin.com/blog/work/2019/06/23/avoid-indirection), maybe less relevant for the specific `_idle_timeout` getter, but *definitely* the case for the `rendered_worker_pod_template` property

mrocklin: That sounds like something that should be discussed as an issue and then implemented separately as a PR if the community agrees with your logic.  In the meantime I recommend that you revert these changes so that the main part of this PR isn't blocked on them.

ig248: @mrocklin having these attributes available as properties outside `_start()` is integral to the PR, so I am open to alternative suggestions but I don't see an easy way to revert a sub-set of changes?

jacobtomlinson:
Sorry for the delay in getting back to this. I've been out sick.

While I generally agree with your points around testability and maintainability there is a lot in this PR that I do not agree with and would've appreciated a discussion in an issue.

Isolating unit tests and integration tests to allow quick execution of individual unit tests while developing locally is a reasonable thing to do. However, given the fact that this codebase is so small and the tests run reasonably fast, I'm surprised this has come up. I'm also not sure of the benefit of being able to run tests without k8s as most local Docker installations ship with k8s out of the box and the integration tests are the important factor here as the whole purpose of the library is to integrate.

I completely agree that moving the rendering of the pod template to a property makes sense. However moving things like `idle_timout`, `name`, `namespace` etc to properties just feels unnecessary. Any property which is just checking the kwarg against the Dask config seems unnecessary and should probably be moved back to the `_start` method.

I also agree that `assert MyClass(key=value).key == value` makes sense. However, it would've been nice to discuss making these attributes public in a separate issue as it is a change to the public API and doesn't feel related to separating the tests. I'm also very confused about having public `key` attributes but also having a private `_key` property. For situations where properties make sense it feels like the internals required to create the property value should be kept private and the property itself should be public.

ig248:
@jacobtomlinson Thanks, I agree I may have changed a few things too many here in one PR. I will look into factoring out just the templating functions/properties while attempting to keep any changes to the public api to a minimum. *Some of the changes were inspired by ideas outside the scope of a single PR, like updating workers templates on the fly via a ConfigMap on the "cluster runner" pod, but that'd warrant a separate issue/discussion and is at the moment a bit specific to our deployment scenario.*

Issue #244, opened 2020-04-07
KubeCluster fails to adapt 


heavenmarshal:
I found the KubeCluster fails to adapt. After I call
```python
>>> cluster.adapt(minimum=2, maximum=2)
```
it is supposed to create a new worker pod if I delete one worker pod by kubectl delete pods (actually it is the behavior of the previous version, e.g. 0.8.0). However, there is no new worker created if I use the latest version 0.10.1. My code and configuration is provided blow.
```Dockerfile
# Dockerfile
FROM python
WORKDIR /app
RUN pip install dask==2.14.0 \
	&& pip install distributed==2.14.0 \
	&& pip install dask-kubernetes==0.10.1
COPY k8s_cluster.py /app
COPY worker/worker.yaml /app
```
```python
# k8s_cluster.py
import os
import yaml
from dask_kubernetes import KubeCluster

def build_k8s_cluster(n_workers):
    worker_config = yaml.safe_load(open('worker.yaml', 'r'))
    pod_name = worker_config['metadata'].pop('name', 'dask-worker') + '-{uuid}'
    namespace = os.getenv('MY_POD_NAMESPACE')
    cluster = KubeCluster.from_dict(worker_config, namespace=namespace, name=pod_name)
    cluster.adapt(minimum=n_workers, maximum=n_workers)
    return cluster
```
```yaml
# pod.yaml, manifest for the scheduler pod
apiVersion: v1
kind: Pod
metadata:
  name: example
  namespace: example
spec:
  serviceAccountName: example-serviceaccount
  restartPolicy: Never
  containers:
  - name: example
    image: heavenmarshal/dask-k8s-adapt
    imagePullPolicy: Always
    command: ["sleep"]
    args: ["2000"]
    env:
      - name: MY_POD_NAMESPACE
        valueFrom:
          fieldRef:
            fieldPath: metadata.namespace
```
```yaml
# worker/worker.yaml manifest file for the workers
kind: Pod
metadata:
  name: dask-worker
spec:
  restartPolicy: Never
  containers:
    - image: "heavenmarshal/dask-k8s-adapt"
      imagePullPolicy: Always
      command: ["/bin/bash"]
      args: ["-c", "dask-worker --nthreads '1' --no-bokeh --memory-limit 200Mi --death-timeout '60'"]
      name: dask-worker
      env:
        - name: EXTRA_PIP_PACKAGES
          value: fastparquet git+https://github.com/dask/distributed
        - name: PYTHONPATH
[...]

PR #246, opened 2020-05-07
test cluster.adapt when deleting pods

heavenmarshal:


jsignell:
Just for reference - these are tests for the issue described in #244

dask_kubernetes/tests/test_async.py
```diff
@@ -786,3 +786,29 @@ async def test_start_with_workers(pod_spec, ns, auth):
...
+        if x.metadata.name.startswith(cluster.name)
+    ]
+    assert len(worker_pods) == 2
+    # delete one worker pod
+    await core_api.delete_namespaced_pod(name=worker_pods[0], namespace=ns)
+    # test whether adapt will bring it back
+    start = time()
+    while len(cluster.scheduler_info["workers"]) != 2:
+        await asyncio.sleep(0.1)
+        assert time() < start + 20
```
jsignell, line 814: It might be nice to explicitly assert that there are two worker pods after you've waited

heavenmarshal: all fixed

jacobtomlinson:
This is in! Sorry things took so long here @heavenmarshal.

PR #251, opened 2020-05-22
Update core.py, added support of NodePort k8s service type

black2kanoi:
added support of NodePort k8s service type

### DIFF
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -209,7 +209,7 @@ class Scheduler(Pod):
             )
         if self.service.spec.type == "NodePort":
             start = time.time()
-            hostname = dask.config.get("kuberntetes.scheduler-service-nodeport-host")
+            hostname = dask.config.get("kubernetes.scheduler-service-nodeport-host")
             for port in self.service.spec.ports:
                 if port.name == "comm":
                     node_port = port.node_port
@@ -221,8 +221,6 @@ class Scheduler(Pod):
             self.dashboard_external_address = "http://{host}:{port}".format(
                 host=hostname, port=dashboard_node_port
                 )
-            print("External scheduler address: " + self.external_address)
-            print("External dashboard address: " + self.dashboard_external_address)
 
         # FIXME Create an optional Ingress just in case folks want to configure one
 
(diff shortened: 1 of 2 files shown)


dask_kubernetes/core.py
```diff
@@ -207,7 +207,22 @@ async def start(self, **kwargs):
...
+                if port.name == "dashboard":
+                    dashboard_node_port = port.node_port
+            self.external_address = "tcp://{host}:{port}".format(
+                host=hostname, port=node_port
+                )
+            self.dashboard_external_address = "http://{host}:{port}".format(
+                host=hostname, port=dashboard_node_port
+                )
+            print("External scheduler address: " + self.external_address)
+            print("External dashboard address: " + self.dashboard_external_address)
```
jacobtomlinson, line 225: Please remove these print lines.

dask_kubernetes/core.py
```diff
@@ -207,7 +207,22 @@ async def start(self, **kwargs):
             self.external_address = "tcp://{host}:{port}".format(
                 host=loadbalancer_host, port=SCHEDULER_PORT
             )
-        # FIXME Set external address when using nodeport service type
+        if self.service.spec.type == "NodePort":
+            start = time.time()
+            hostname = dask.config.get("kuberntetes.scheduler-service-nodeport-host")
```
jacobtomlinson, line 212: Could you please set a default for this in the `kubernetes.yaml` configuration file.

dask_kubernetes/core.py
black2kanoi: fixed in new commit

dask_kubernetes/core.py
black2kanoi: added to defaults, and fixed typo. in new commit

mr2dark:
Is it possible to integrate this pull request? It worked for my case with `NodePort`.
The only thing for this be consistent is that `dask.distributed` should be able to derive dashboard address (e.g. for the widget) from  `cluster.scheduler.dashboard_external_address` (if it's available) rather than from the metadata received for the scheduler service.

jacobtomlinson:
Superseded in #304

Issue #185, opened 2019-10-14
Connect to existing running cluster


jacobtomlinson:
It would be great to be able to attach a `KubeCluster` instance to an already running Dask cluster. This could be a cluster created with the [Dask helm chart](https://github.com/helm/charts/tree/master/stable/dask) for example.

We could then be able to take over scaling duties locally in our `KubeCluster` object and generally be able to interact with the cluster from our Python session.

raybellwaves:
I would be interested in working on this. I will need some hand-holding though.

jacobtomlinson:
Thanks @raybellwaves.

The main challenges here are going to be that the two approaches handle scaling different.y

In dask-kubernetes we are manually creating pods. However the Helm Chart uses a deployment to create them. There are pros and cons to each approach.

To support things here we would need to detect which mode of cluster we are connecting to and change our scaling technique accordingly.

griseau:
Hi guys, 

Do you have any update on this ? Or do you have any other solution to run the scheduler in kubernetes and be able to adapt the number of pods ? As far as I know, not possible today using helm deployment 

Thanks

jacobtomlinson:
This has been implemented in the operator in #460 and we have no plans to implement this in the classic `KubeCluster`. 

If you need this functionality I recommend switching to the [operator](https://kubernetes.dask.org/en/latest/operator.html).

Issue #186, opened 2019-10-14
Detach from running cluster


jacobtomlinson:
The current behaviour of dask-kubernetes is to create ephemeral clusters and use them to perform calculation. There is a bunch of logic here to clean up the cluster when the Python object is removed.

This may not always be the desired use case and we may want to support detaching from a cluster. This also fits with #185 where we may want to attach to an already running cluster.

mrocklin:
I think that @jcrist has conventions for this in dask-gateway and
dask-yarn.

jcrist:
We use the keyword `shutdown_on_close` in dask-gateway (defaults to True, except when connecting to a running cluster then the default is False). Typically a `cluster` object has a few things you'd want to cleanup:

- Remote cluster resources (e.g. kubernetes pods)
- Local async things (event loop, background tasks, comms, etc...)

The `.close()` method is always called on GC of an object, but the meaning of `.close()` differs based on the value of `shutdown_on_close` for the cluster.

```python
# On .close()/GC of cluster object, remote cluster is cleaned up
cluster = GatewayCluster()
cluster.close()

# On .close()/GC of cluster object, only local resources are cleaned up
cluster = GatewayCluster(shutdown_on_close=False)
cluster.close()

# You can override this behavior by passing `shutdown=True` to close
# or using the `shutdown()` method which is an alias to this
cluster.close(shutdown=True)
cluster.shutdown()
```

This option only makes sense if the scheduler is remote (which is always true for dask-gateway).

jacobtomlinson:
Thanks @jcrist this is the kind of thing I want to implement here. Now that we have the option of [creating remote schedulers](https://github.com/dask/dask-kubernetes/issues/84#issuecomment-541601780) it would be good to implement this functionality.

jacobtomlinson:
I'm going to close this out as being closely related to #185.

PR #255, opened 2020-06-02
Helm cluster

jacobtomlinson:
I seem to find myself continuously explaining to people the difference between the Dask Helm chart and dask-kubernetes, and the fact that they are incompatible because of their underlying designs. It is not obvious, and we should fix that.

In an attempt to resolve this I've added a `HelmCluster` cluster manager to dask-kubernetes which provides a thin wrapper around an existing Dask cluster which has been deployed via the Helm Chart.

```bash
# Install the chart using Helm
helm repo add dask https://helm.dask.org
helm repo update

helm install dask/dask --name myrelease
```

```python
# Connect to the cluster from a Python session and manually scale
from dask_kubernetes import HelmCluster

cluster = HelmCluster(release_name="myrelease")

cluster.scale(10)

client = Client(cluster)

# Do some Dask work

cluster.scale(1)
```

### Motivations

- Provide Helm Chart users with a cluster manager that has some useful features such as being able to call `cluster.scale(n)` and `cluster.get_logs()` from within their Python or Jupyter session.
- Update the documentation to provide a more clear explanation of the differences between `KubeCluster` and `HelmCluster`.
- Have a documentation page which can be referred to when explaining to users the differences.

### Challenges 

While there isn't much code to `HelmCluster` there are a few design challenges around how the cluster manager works.

**I have not made it possible to install the Helm Chart using `HelmCluster`**
Users can only to connect to an existing deployment. This is because I do not want to encourage people to install the Helm chart this way. Helm charts should be installed and managed natively using `helm`. Handling config, lifecycle, upgrades, etc would add bloat to the class.

**Scaling down the Helm Chart is not a safe operation**
Kubernetes assumes all pods are stateless and equal. When scaling down a Dask cluster the scheduler will instruct empty/idle workers to exit. Or rearrange work in order to make empty/idle workers. Kubernetes will reduce the deployment count to the desired number, killing containers or starting containers to satisfy this. This results in a race condition where kubernetes may kill pods before they are ready, or restart them if Dask closes them too soon. To work around this I've added documentation instructing users only to scale down when they know it is safe

**Adaptive scaling will be unstable**
As scaling down is not safe this means that adaptive scaling will likely be unstable and result in clusters thrashing as futures are lost and recalculated. I have left adaptivity in just in case there are valid use cases that I haven't thought of, however, if you call `cluster.adapt()` you will get a warning against doing that altogether. 

### Outstanding tasks

- [x] Fix the CI (I imagine it will not handle the new helm tests)
- [x] Update the [Kubernetes page in the Dask documentation](https://docs.dask.org/en/latest/setup/kubernetes.html) (dask/dask#6290)
- [x] Update the [Helm Chart repo](https://helm.dask.org/) with a link to the `HelmCluster` docs (dask/helm-chart#66)

### Future work

- Update the lab extension with the ability to configure and autocreate cluster object. This would be useful in the Jupyter that comes with the Helm Chart to already have the `HelmCluster` object listed.


raybellwaves:
Thanks for working on this. Am I right that this closes https://github.com/dask/dask-kubernetes/issues/185?

jacobtomlinson:
No problem @raybellwaves. Do you have any interest in testing this out and code reviewing?

> Am I right that this closes #185?

That's a tough one. I'm tempted to say yes, but there is a little nuance here. 

I had intended #185 to compliment #186. In #186 the idea would be to be able to detach from a running `KubeCluster` instance and connect to it again later. This PR is a little different in that you are connecting to a running Helm Cluster, which is not the same as a KubeCluster.

I think #186 would still be valuable, and we would need a way to attach too.

dask_kubernetes/helm.py
```diff
@@ -0,0 +1,244 @@
...
+    release_name: str
+        Name of the helm release to connect to.
+    namespace: str (optional)
+        Namespace in which to launch the workers.
+        Defaults to current namespace if available or "default"
+    port_forward_cluster_ip: bool (optional)
+        If the chart uses ClusterIP type services, forward the ports locally.
+        If you are using ``HelmCluster`` from the Jupyter session that was installed
+        by the helm chart this should be ``False``. If you are running it locally it should
+        be ``True``.
```
gforsyth, line 40: I don't think it's possible to query the status of a `helm` release if you're on a pod inside of that release, is it?

jacobtomlinson: If you copy your local `~/.kube/config` to the Jupyter pod you should be able to. Perhaps this should be documented?

Perhaps we should also consider adding a service role to the chart which allows modifying the worker deployment replica count and getting pod logs. Although that could have security implications.

jacobtomlinson:
Thanks for the review @gforsyth 

> The docstring for HelmCluster implies that a user might run this on the jupyter notebook that is spun up by the helm chart, but I'm not sure that can work, since that will be within the kubernetes cluster itself and I don't know that it can connect kubernetes from there, can it?

If you copy your local ~/.kube/config to the Jupyter pod you should be able to. Perhaps this should be documented?

> I tried to call cluster.adapt() per your warning (where cluster is a HelmCluster)

Hmm, that's frustrating. I'm hesitant to put the effort in here as adapting will not be a good experience with this cluster type. Perhaps instead of warning, it should raise a `NotImplementedError` and explicitly deny this behaviour.

gforsyth:
Ahh, of course. I'll give that a shot.

I think an explicit not implemented is the way to go for the adapt issue.

Timost:
Unless I'm wrong, this implicitly makes `helm` a new mandatory dependency of `dask-kubernetes`, if that's the case, I think the doc should be updated to mention that.

PR #257, opened 2020-07-03
Support passing dict/yaml paths directly to constructor

### LINKED ISSUE #254: Add support for passing in yaml template path in main constructor
In Prefect I'm trying to standardize some of our configuration on specifying a dask cluster class and kwargs for that class. For example:

```python
executor = Executor(
    cluster_class="dask_yarn.YarnCluster",
    cluster_kwargs={"environment": "..."},
)
```

This is proving a bit tricky with dask-kubernetes, since people commonly use the `KubeCluster.from_yaml` classmethod, whereas most other dask cluster implementations the class constructor is the common method. We can support this as well by taking any callable in as `cluster_class`, but it feels a bit wonky.

I'm wondering if y'all might be amenable to adding support for passing in either a dict or a yaml path as `pod_template` in the main constructor? This makes the `from_yaml` and `from_dict` constructors redundant, since all of this would be done in the main constructor.

jcrist:
Support passing in `dict` and `yaml` paths to `KubeCluster` constructor directly, rather than using classmethods like `from_yaml`. This makes `KubeCluster` play nicer with other tools that configure dask clusters based on a class name and class kwargs, since the classmethod will no longer be needed. Also deprecates `from_dict`/`from_yaml`.

Fixes #254.


jcrist:
Tests are hanging, and for some reason the github action logs for this are unavailable (cc @jacobtomlinson) so I'm not sure where the hang occurs. I don't currently have a setup available for testing this, will attempt to clean things up next week if no one else has a quick fix. Was hoping to debug via CI alone.

jacobtomlinson:
Things seem good here no that we're using the new CI and pytest machinery.

I'm going to go ahead and merge this. Sorry this took so long!

PR #260, opened 2020-08-06
Issue 170 ssl error when cleaning up pods

csala:
This change handles the issue mentioned in #170, which apparently was not completely solved even though the issue was closed.

The change consists on passing down `self.core_api` to `_cleanup_resources` instead of creating a new instance inside it.


jacobtomlinson:
The reason we do not pass this down is because it is an instance of the async version of the Kubernetes library. When the `_cleanup_resources` code is called it is likely the event loop will already be stopped. Hence us using the sync version.

I notice your changes `yield` the coroutines rather than `await` them. I've not seen this before. Is this an intentional workaround for the event loop being stopped?

csala:
> I notice your changes `yield` the coroutines rather than `await` them. I've not seen this before. Is this an intentional workaround for the event loop being stopped?

I thought it did, but since you asked I dug a bit further to see exactly why and I realized that it does not seem actually work. Instead, I think that the flow is silently getting stuck at that `yield` and not doing anything after it.

However, the cleanup seems to be happening either way, even if I completely remove the `_cleanup_resources` call, which makes wonder whether it is actually needed.

I'm closing this PR and opening an issue to discuss the right way to approach this.

PR #263, opened 2020-08-26
Make the helm command an optional dependency

### LINKED ISSUE #262: Helm command is now a mandatory dependency
Unless I'm wrong, this implicitly makes `helm` a new mandatory dependency of `dask-kubernetes`, if that's the case, I think the doc should be updated to mention that.

_Originally posted by @Timost in https://github.com/dask/dask-kubernetes/pull/255#issuecomment-680084182_

jacobtomlinson:
The check for the `helm` command was accidentally at the module level, so it would raise on import of `dask_kubernetes`.

Updated the check so that it happens when `HelmCluster` is instantiated.

Closes #262

### DIFF
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -154,8 +154,8 @@ class Scheduler(Pod):
     """
 
     def __init__(self, idle_timeout: str, service_wait_timeout_s: int = None, **kwargs):
-        self.cluster._log("Creating scheduler pod on cluster. This may take some time.")
         super().__init__(**kwargs)
+        self.cluster._log("Creating scheduler pod on cluster. This may take some time.")
         self.service = None
         self._idle_timeout = idle_timeout
         self._service_wait_timeout_s = service_wait_timeout_s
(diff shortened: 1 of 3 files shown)


dask_kubernetes/core.py
```diff
@@ -154,6 +155,7 @@ class Scheduler(Pod):
 
     def __init__(self, idle_timeout: str, service_wait_timeout_s: int = None, **kwargs):
         super().__init__(**kwargs)
+        self.cluster._log("Creating scheduler pod on cluster. This may take some time.")
```
Timost, line 158: I'm hitting a 
```
AttributeError: 'KubeCluster' object has no attribute '_log'
```
When instantiating a `KubeCluster`

jacobtomlinson: Thanks @Timost. This is dependent on an upstream change which will be released today. Hence why we haven't released here yet. If you are testing `dask-kubernetes` from master please also install `distributed` from master.

Timost: Ha my bad ! Sorry for the noise.

Issue #144, opened 2019-05-01
Add docs for GPU deployment


mrocklin:
Should we include a small section informing users how to request containers with GPUs?  My gut reaction to this is "yes", but I'm also in a biased situation.  Ideally I'd like for some non-NVIDIA affiliated person to sign off on this.  

Looking at you @yuvipanda :)

yuvipanda:
yes! I haven't used it personally, but let's do it!

beberg:
I should finish the docs that go along with the cuda-worker and Helm chart changes... I'll try and do that in the next few days.

PR #272, opened 2020-10-01
Doc Updates

quasiben:
Resolves https://github.com/dask/dask-kubernetes/issues/144

This is a first pass at getting a working example and updating docs on using dask-kubernetes with GPUs (this has been coming up a more for me and having some docs I can point to would be very valuable)

Also adds docs on remote deploys

### DIFF
--- a/doc/source/kubecluster.rst
+++ b/doc/source/kubecluster.rst
@@ -78,7 +78,37 @@ Best Practices
     This ensures that these pods will clean themselves up if your Python
     process disappears unexpectedly.
 
+GPUs
+----
 
+Because ``dask-kubernetes`` uses standard kubernetes pod specifications, we can
+use `kubernetes device plugins
+<https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/#using-device-plugins>`_
+and add resource limits defining the number of GPUs per pod/worker.
+Additionally, we can also use tools like `dask-cuda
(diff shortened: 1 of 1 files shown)


doc/source/kubecluster.rst
```diff
@@ -192,4 +225,45 @@ The pods are created with two default `tolerations <https://kubernetes.io/docs/c
...
+
+When deploying remotely, the following k8s resources are created:
+
+- A pod with a scheduler running
+- (optional) A pod with a LoadBalancer and complimentary service (svc) to
+  expose scheduler and dashobard ports
+
+By default, the ``scheduler-service-type`` is set to ``ClusterIp``. To
+optionally use the LoadBalancer, change ``scheduler-service-type`` to
+``LoadBalancer``
```
holdenk, line 268: Where should folks set this? Looking at the code it seems like it's pulled from `kubernetes.yaml`, which does not seem to be configurable.

holdenk: Is it with `dask.config`?

holdenk: Answered my own question, was with `dask.config.set({"kubernetes.scheduler-service-type": "LoadBalancer"})`, maybe we can add that?

quasiben: Thanks the for the suggestion @holdenk.  Added your thoughts in the last commit

jacobtomlinson (approved):
Thanks @quasiben

PR #275, opened 2020-10-09
Make deployment names configurable in HelmCluster

bnaul:
Not sure if others would find this useful but in our case we deploy our own helm chart that's based on, but not identical to, the standard Dask chart. In particular the deployment names are slightly different, so the hard-coded versions in `HelmCluster` don't work for us.

For now I'm just monkey patching the functions with hard-coded deployment names but it doesn't seem like it'd hurt to make it configurable 🙂.


quasiben:
👍

jacobtomlinson (approved):
This is great!

PR #278, opened 2020-11-07
Transform username to lowercase when escaping

Yuhuishishishi:
Add lower() after escaping the username.
This is a fix to #213.

Test method:
```
make test EXTRA_TEST_ARGS='-k escape'
```

### DIFF
--- a/dask_kubernetes/tests/test_async.py
+++ b/dask_kubernetes/tests/test_async.py
@@ -619,8 +619,9 @@ async def test_repr(cluster):
 
 
 @pytest.mark.asyncio
-async def test_escape_username(pod_spec, ns, auth, monkeypatch):
-    monkeypatch.setenv("LOGNAME", "foo!._")
+@pytest.mark.parametrize("name", ["foo!._", "Foo!._"])
+async def test_escape_username(pod_spec, ns, auth, monkeypatch, name):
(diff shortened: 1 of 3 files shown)


Yuhuishishishi:
I am pretty new. Thanks for your review in advance.

dask_kubernetes/tests/test_async.py
```diff
@@ -619,8 +619,9 @@ async def test_repr(cluster):
 
 
 @pytest.mark.asyncio
-async def test_escape_username(pod_spec, ns, auth, monkeypatch):
-    monkeypatch.setenv("LOGNAME", "foo!._")
+@pytest.mark.parametrize("name", ["foo!._", "Foo!._"])
```
jacobtomlinson, line 622: I think parametrizing the tests here is unnecessary. Changing it to uppercase would be acceptable.

jacobtomlinson (commented):
Firstly, welcome!

Thanks for raising this. Just one small comment.

dask_kubernetes/tests/test_async.py
Yuhuishishishi: Thanks. Yes, I think it makes sense to keep just one test case. Reverted the change.

jacobtomlinson:
Many thanks!

Issue #279, opened 2020-11-07
`make k8s-deploy` pointing to wrong path of `test-runner-setup.yaml`


Yuhuishishishi:
**What happened**:

`make k8s-deploy` from local dev environment will fail because it's pointing to `kubernetes/test-runner-setup.yaml` instead of `ci/test-runner-setup.yaml`

https://github.com/dask/dask-kubernetes/blob/ccb9864d463a3e14988af53a947a658cf32e7e10/Makefile#L59-L60

Instead, I think it should point to

```
k8s-deploy:
	kubectl --context=${K8S_TEST_CONTEXT} apply -f ci/test-runner-setup.yaml
                                                       ^^^^
```

**What you expected to happen**:

**Minimal Complete Verifiable Example**:

```bash
make k8s-deploy
```
gives 

```
kubectl --context=kind-kind apply -f kubernetes/test-runner-setup.yaml
error: context "kind-kind" does not exist
Makefile:60: recipe for target 'k8s-deploy' failed
make: *** [k8s-deploy] Error 1
```

**Anything else we need to know?**:

**Environment**:

- Dask version: 2.30.0+43.g80239dab
- Python version: Python 3.7.0
- Operating System: 18.04.1-Ubuntu 5.4.0-47-generic
- Install method (conda, pip, source): conda

Yuhuishishishi:
closing due to merged PR. thank @jacobtomlinson

PR #280, opened 2020-11-08
Change test-runner-setup.yaml path under make k8s-deploy

Yuhuishishishi:
Fix to #279 .

test-runner-setup.yaml is under ci/ instead of kubernetes/


jacobtomlinson:
Thanks!

PR #289, opened 2020-12-18
Fix docker file to be able to use the docker-make make command

pvanderlinden:


pvanderlinden:
This fixes the make file. As it assumes the docker image has make installed.

jacobtomlinson:
Thanks!

PR #291, opened 2021-01-13
add nodehost and nodeport kwargs to HelmCluster init

adammrozik:
This is a small PullRequest to allow user side parametrization of dask scheduler service when NodePort is chosen as its type.

As this is a very minor change, do you require any additional testing and/or documentation? Manually I have been able to connect without an issue with this change.

### DIFF
--- a/dask_kubernetes/helm.py
+++ b/dask_kubernetes/helm.py
@@ -250,7 +250,11 @@ class HelmCluster(Cluster):
         await self.apps_api.patch_namespaced_deployment(
             name=f"{self.release_name}-{self.worker_name}",
             namespace=self.namespace,
-            body={"spec": {"replicas": n_workers,}},
+            body={
+                "spec": {
+                    "replicas": n_workers,
+                }
+            },
         )
 
     def adapt(self, *args, **kwargs):
--- a/dask_kubernetes/helm.py
+++ b/dask_kubernetes/helm.py
@@ -41,8 +41,11 @@ class HelmCluster(Cluster):
         Name of the Dask worker deployment in the current release.
         Defaults to "worker".
     **kwargs: dict
-        Additional keyword arguments to pass to Cluster
-
+        nodeport_host string (optional)
+            A Node address. Can be provided in case scheduler service type is NodePort
+            and you want to avoid listing Nodes at a cluster scope.
+        node_port: int (optional)
+            A NodePort for your scheduler service.
     Examples
     --------
     >>> from dask_kubernetes import HelmCluster
@@ -77,6 +80,7 @@ class HelmCluster(Cluster):
         asynchronous=False,
         scheduler_name="scheduler",
         worker_name="worker",
+        **kwargs
     ):
         self.release_name = release_name
         self.namespace = namespace or _namespace_default()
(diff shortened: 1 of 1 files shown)


dask_kubernetes/helm.py
```diff
@@ -41,8 +41,11 @@ class HelmCluster(Cluster):
         Name of the Dask worker deployment in the current release.
         Defaults to "worker".
     **kwargs: dict
-        Additional keyword arguments to pass to Cluster
-
+        nodeport_host string (optional)
+            A Node address. Can be provided in case scheduler service type is NodePort
+            and you want to avoid listing Nodes at a cluster scope.
+        node_port: int (optional)
+            A NodePort for your scheduler service.
```
jacobtomlinson, line 48: I think the ordering and indentation is off here.

dask_kubernetes/helm.py
```diff
@@ -99,6 +103,10 @@ def __init__(
         self.scheduler_name = scheduler_name
         self.worker_name = worker_name
 
+        # Optional argument for explicit node host connection parameters.
+        self.nodeport_host = kwargs.get('nodeport_host')
+        self.node_port = kwargs.get('node_port')
```
jacobtomlinson, line 108: Let's make these full kwargs

dask_kubernetes/helm.py
```diff
@@ -131,8 +139,12 @@ async def _get_scheduler_address(self):
             host = lb.hostname or lb.ip
             return f"tcp://{host}:{port}"
         elif service.spec.type == "NodePort":
-            nodes = await self.core_api.list_node()
-            host = nodes.items[0].status.addresses[0].address
+            if self.nodeport_host:
+                host = self.nodeport_host
+                port = self.node_port
```
jacobtomlinson, line 144: You check if host is set here, but what if port is set without host?

jacobtomlinson (commented):
Thanks for this.

It would be best if we could not use the `kwargs` dict and make them full kwarg options.

dask_kubernetes/helm.py
adammrozik: Adjusted

dask_kubernetes/helm.py
adammrozik: You are right! Adjusted

dask_kubernetes/helm.py
adammrozik: Adjusted docstrings a little bit to be more compact. Not sure what should be an appropriate way of documenting `kwargs`. Please advise me if it looks ok now.

jacobtomlinson (approved):
Thanks @adammrozik. 

Your changes were almost what I meant. I've pushed a couple of changes in myself.

Happy to merge on passing CI.

adammrozik:
> Thanks @adammrozik.
> 
> Your changes were almost what I meant. I've pushed a couple of changes in myself.
> 
> Happy to merge on passing CI.

Thanks @jacobtomlinson !

Pipeline is failing though... 

but it does not seem to be related to the changes introduced in this PR. Do you know what could be the reason?

adammrozik:
@jacobtomlinson , do you know what could be the reason of these fails?

jacobtomlinson:
Looks like some issues in the pip resolver. I'm happy to just get this merged.

Thanks @adammrozik

PR #297, opened 2021-02-08
Change default branch from master to main

jsignell:


.github/workflows/ci.yaml
```diff
@@ -5,7 +5,7 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@master
+      - uses: actions/checkout@main
```
jacobtomlinson, line 8: These are branches on external repos, so are not under our control to change.

versioneer.py
```diff
@@ -620,7 +620,7 @@ def git_versions_from_keywords(keywords, tag_prefix, verbose):
         # refs/heads/ and refs/tags/ prefixes that would let us distinguish
         # between branches and tags. By ignoring refnames without digits, we
         # filter out many common branch names like "release" and
-        # "stabilization", as well as "HEAD" and "master".
+        # "stabilization", as well as "HEAD" and "main".
```
jacobtomlinson, line 623: This is a comment made within the versioneer code. No harm in changing but I think it would invalidate the comment.

jacobtomlinson (commented):
Thanks for the work here @jsignell. Couple of comments but happy to move forwards. CI failures are known issue and unrelated.

.github/workflows/ci.yaml
jsignell:
```suggestion
      - uses: actions/checkout@master
```

.github/workflows/ci.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@master
+      - uses: actions/checkout@main
```
jsignell, line 29:
```suggestion
      - uses: actions/checkout@master
```

versioneer.py
jsignell: I think the comment is still valid.

jacobtomlinson: Oh yeah I guess so!

jsignell (commented):
I think this one is good to merge whenever you are ready @jacobtomlinson

jrbourbeau (approved):
Thanks @jsignell! (test failures are unrelated)

jacobtomlinson:
Thanks folks!

PR #301, opened 2021-02-17
Checks for non "dask-in-name" helm releases

MridulS:
According to https://github.com/dask/helm-chart/blob/master/dask/templates/_helpers.tpl#L14,
`helm install testdask dask/dask` will lead to pods with names `testdask-scheduler`, `testdask-worker`, `testdask-jupyter` and something like `helm install thisisnotdsak dask/dask` will lead to `thisisnotdsak-dask-scheduler`, ... pods.

This logic isn't present in `HelmCluster`, so if a release name without `dask` in it is given, it errors out.

``` python
>>> HelmCluster(release_name="thisisnotdsak")
.....
.....
ApiException: (404)
Reason: Not Found
HTTP response headers: <CIMultiDictProxy('Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'X-
Kubernetes-Pf-Flowschema-Uid': 'd181afca-99cd-42f8-9e80-243323d159e7', 'X-Kubernetes-Pf-Prioritylevel-Uid': 
'94319e3e-d012-40a5-bb2b-97063c3a1b9c', 'Date': 'Wed, 17 Feb 2021 17:15:07 GMT', 'Content-Length': '222')>
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"services \"thisisnotdsak-
scheduler\" not found","reason":"NotFound","details":{"name":"thisisnotdsak-scheduler","kind":"services"},"code":404}

```

I am not sure if there is a cleaner way of getting the chart name from a helm release. (Or should I just hard-code `dask` in it?)

PS: The failing tests should pass after https://github.com/dask/dask-kubernetes/pull/295 is merged in.


dask_kubernetes/helm.py
```diff
@@ -121,7 +122,30 @@ async def _start(self):
         await super()._start()
 
     async def _get_scheduler_address(self):
-        service_name = f"{self.release_name}-{self.scheduler_name}"
+        # find the chart name, is there a better way?
+        chart = subprocess.run(
```
jacobtomlinson, line 126: Perhaps `subprocess.check_output` would be neater here?

jacobtomlinson (commented):
This looks great. Just a small comment.

dask_kubernetes/helm.py
MridulS: yup, that's better :)

jacobtomlinson:
CI should pass now if you pull in latest changes.

MridulS:
Did some preliminary search for the test failures
- The name set in cluster init would fail as the metadata isn't updated until `_start` is called, reverted https://github.com/dask/dask-kubernetes/commit/3c441966b61de23b8ad58fd31affccf6405f88e2 in https://github.com/dask/dask-kubernetes/pull/301/commits/58d73ad787c0168d9654a5b40e9f4f952b2e4f1a
- There seems to be a regression due to recent changes to https://github.com/dask/distributed/commit/383ea0326ae103b5d5e0b62ed9c3cb18510c5b9e (locally the CI passed with an older commit, waiting to confirm with Github Actions)

requirements-dev.txt
```diff
@@ -3,4 +3,4 @@ black>=18.9b0
 pytest>=5.3
 pytest-asyncio>=0.10.0
 git+https://github.com/dask/dask@master
-git+https://github.com/dask/distributed@master
+git+https://github.com/dask/distributed@02a85f7891b5872aa8e7a0fc62c2891f503a7963
```
jacobtomlinson, line 6: Can you unpin this please. If there are failures against distributed master we need to fix those first in a separate PR.

MridulS: unpinned it now, just did it for a sanity check.

PR #302, opened 2021-02-17
name CLI argument has been removed from helm3, use helm install releasename instead

MridulS:


jacobtomlinson (approved):
Good spot thanks!

PR #304, opened 2021-03-05
Add support for port forwarding and switch tests to use pytest-kind

jacobtomlinson:
This PR started out as an effort to simplify testing locally and in CI. As part of this effort I've ended up doing a bunch of refactoring and have also implemented automatic port forwarding.

The new requirements for running tests should just be Docker. Pytest will automatically start a kubernetes cluster using `pytest-kind` and everything involved in the setup is now handled entirely in fixtures (no more Makefile).

To use `dask-kubernetes` there is now a hard dependency on the `kubectl` binary for `KubeCluster` and also the `helm` binary for `HelmCluster`. Tests also require the `docker` command for image building.

To enable this simplification I have done two things:
- Switched the default scheduler mode to remote, a scheduler pod will now be started by default. I am hopefully intending to remove local mode altogether in the future.
- In the case that the scheduler service is using a ClusterIP and is not exposed externally we now check if we can connect directly (because we are inside the cluster ourselves) and if that fails we use `kubectl` to port forward the service locally, similar to an SSH reverse tunnel.

This PR also includes some code simplification as many things like shutdown hooks are now handled upstream in `SpecCluster` and have been for some time.


jacobtomlinson:
I intend to merge this in 24 hours if there are no further comments 😃.

pvanderlinden:
I have a few questions:
- kubectl is now required for the python code running dask-kubernetes, but not for the images running the actual scheduler?
- What is the impact of the external scheduler? So far, when the python process reunning dask-kubernetes with an embedded scheduler, if the python process dies the workers will be cleaned up. Will this still happen with the external scheduler on a dirty shutdown (python process dies with segfault or something else severe)?
- I only quickly scanned through the code, but wouldn't it make more sense to make kubectl only a requirement when port forwarding is required (the only functionality you are using?), which would be during development? Instead of requiring even in production usage when kubectl is probably not needed, nor installed?

pvanderlinden:
I haven't completely ran through the code, but my impression is that this will probably make it work out of the box in local testing environments, which is great!

jacobtomlinson:
> kubectl is now required for the python code running dask-kubernetes, but not for the images running the actual scheduler

Correct. 

It's actually only required if the scheduler service type is `ClusterIP` and you are using `KubeCluster` from outside the Kubernetes cluster. It is used to port forward the scheduler service. Ideally this would be possible with the kubernetes Python library (which should have all the same capabilities as kubectl) but it seems it does not work properly. 

Perhaps in the future this will become possible and we could remove this dependency again in a future Pull Request.

> What is the impact of the external scheduler? So far, when the python process reunning dask-kubernetes with an embedded scheduler, if the python process dies the workers will be cleaned up. Will this still happen with the external scheduler on a dirty shutdown (python process dies with segfault or something else severe)?

Having the scheduler externally has been supported for some time. This change switches it to become the default mode.
To answer your question directly, yes the scheduler will time out along with the workers.

To unpack this a little more there are a couple of things going on here:

- The `SpecCluster` class used as a base for `KubeCluster` will attempt to clean up and remove all resources at garbage collection time.
- In the case of a messy segfault which disrupts garbage collection nothing is cleaned up. However the worker processes will eventually time out allowing the pods to enter a `Completed` phase and no longer take up cluster resources. Depending on your Kubernetes config these `Completed` pods may or may not be cleaned up at some time in the future. We tend to try to clean them up for you on future uses of `KubeCluster`.

With a remote scheduler there is also a Service resource which may not be cleaned up correctly as there is no equivalent to timing out.

One thing I want to address in a future PR is handling the case where we explicitly do not want this cleanup to happen. We want to use `KubeCluster` to create a cluster and then have our Python session end. Then later on (maybe at some future step in some MLOps workflow for example) we may want to recreate a `KubeCluster` object to represent our existing cluster and perform lifecycle actions on it. But all of this will be an explicit user decision. See [dask-ctl](https://github.com/dask-contrib/dask-ctl) for more info.

> I only quickly scanned through the code, but wouldn't it make more sense to make kubectl only a requirement when port forwarding is required (the only functionality you are using?), which would be during development? Instead of requiring even in production usage when kubectl is probably not needed, nor installed?

Sure! I think that's reasonable. I'll move the dependency check.

> I haven't completely ran through the code, but my impression is that this will probably make it work out of the box in local testing environments, which is great!

That's the plan! 🎉

PR #305, opened 2021-03-07
fix: make use of a dummy api client to fix serialization/deserialization

marwan116:
I noted a number of `# FIXME` comments concerning serializing and deserializing kubernetes api objects. I suggest a simple fix of creating a `DummyClient` without the unintended side-effects of launching a thread pool at initialization and with an adapter to the `deserialize` method allowing for the passing of python dictionaries directly instead of having to build a `_FakeResponse` instance.


marwan116:
It looks to me that it is not the change that I introduced that broke the CI but instead the CI was failing already.

I see [this line is potentially problematic](https://github.com/dask/dask-kubernetes/blob/ab863925099b158426575af0800773b9214b385c/dask_kubernetes/core.py#L451) when initializing from a pod_template created from a "minimal" pod spec given `pod.metadata` is still `None` at this point.

jacobtomlinson:
Thanks for picking these up!

The CI is being overhauled in #304 so don't worry about the failures. Once that PR is in it would be good to rebase here.

Issue #286, opened 2020-12-02
KubeCluster loses the ability to run commands against a deployed cluster after periods of inactivity.


drobison00:
Cluster is created as:

```python
cluster = KubeCluster(pod_template=worker_pod, scheduler_pod=sched_pod, deploy_mode='remote')
```

**What happened**:
Initially, all API commands run against the cluster, such as 'scale', 'close', etc.. function as expected. After some period of time, the cluster object fails to authenticate with the cluster and all API commands return a failed status due to being 'unauthorized'.

```
site-packages/dask_kubernetes/core.py", line 81, in close
    await self.core_api.delete_namespaced_pod(name, namespace)
...
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Unauthorized","reason":"Unauthorized","code":401}
```

**What you expected to happen**:
I would expect the cluster object to re-authenticate, or provide some mechanism to re-attach to the created cluster.

```python
# Put your MCVE code here
```

**Anything else we need to know?**:

**Environment**:
Environment is a GKE cluster. Prior to losing the ability to authenticate, everything else functions as expected.

- Dask version: 2.30.0
- Python version: 3.8.6
- Operating System: GKE + CoS
- Install method (conda, pip, source):

jacobtomlinson:
Thanks for raising this @drobison00! This definitely sounds like a bug. 

Are you able to provide the full traceback? It would be helpful to know what the exception was so we can catch it and retry.

drobison00:
Full trace for cluster.scale(n)

```python
ERROR:asyncio:Task exception was never retrieved
future: <Task finished name='Task-13108' coro=<Pod.close() done, defined at /home/drobison/anaconda3/envs/rapids-0.16/lib/python3.8/site-packages/dask_kubernetes/core.py:77> exception=ApiException()>
Traceback (most recent call last):
  File "/home/drobison/anaconda3/envs/rapids-0.16/lib/python3.8/site-packages/dask_kubernetes/core.py", line 81, in close
    await self.core_api.delete_namespaced_pod(name, namespace)
  File "/home/drobison/anaconda3/envs/rapids-0.16/lib/python3.8/site-packages/kubernetes_asyncio/client/api_client.py", line 189, in __call_api
    raise e
  File "/home/drobison/anaconda3/envs/rapids-0.16/lib/python3.8/site-packages/kubernetes_asyncio/client/api_client.py", line 182, in __call_api
[...]

PR #317, opened 2021-03-17
Fix for authentication expiry - Automatically attempt to refresh auth tokens before they expire

drobison00:
Resolves https://github.com/dask/dask-kubernetes/issues/286

@jacobtomlinson Putting this up to get feedback on the approach. If it seems reasonable, I can extend it to other authentication methods.


jacobtomlinson (commented):
This looks great!

From what I can see this will handle updating your local k8s config with tokens from GKE periodically. I'm not sure I see where the auth gets refreshed though. I expect the client objects will pick up the config once and then also need refreshing.

What do you think?

drobison00:
@jacobtomlinson

This should actually be handled by this chunk:

```python
class AutoRefreshConfiguration(Configuration):
    """
    Extends kubernetes_async Configuration to support automatic token refresh.
    Lets us keep track of the loader object
    """
    def __init__(self, loader, *args, **kwargs):
        super(AutoRefreshConfiguration, self).__init__(*args, **kwargs)
        self.refresh_api_key_hook = AutoRefreshConfiguration.refresh_api_key
        self.loader = loader
```

Every time you have a cluster interaction via __call_api, it retrieves the latest credentials for Configuration in client.api_client.py with a call to 'update_params_for_auth', which eventually calls 'get_api_key_with_prefix' on the Configuration object. At this point, if you've set 'refresh_api_key_hook', it will be called before the auth token is retrieved from the Configuration.

I wanted to trigger the auth update here, but Configuration isn't written to be async, so I don't think we can await here. Instead, I have the periodic credential refresh on the loader, and then in 'refresh_api_key_hook' update to the latest token retrieved by the loader.

jacobtomlinson:
Ah ok great!

jacobtomlinson (approved):
This looks great thanks @drobison00

orcahmlee:
Hi, I currently hit the same issue about #286 and seems this PR would fix that.
Do yo have any plan to update new version including this feature on the PyPI?

PR #318, opened 2021-03-18
Add kubectl cluster discovery

jacobtomlinson:
Adds a `discovery` method for `KubeCluster` objects and also the `from_name` method required to reconstruct a `KubeCluster` object.

This PR works by serializing the `KubeCluster` object into a Kubernetes `Secret` resource, allowing it to be recreated at a later time.

This may not be an ideal approach, but opening this anyway as a reference.

```python
from dask_kubernetes import KubeCluster

cluster = KubeCluster(shutdown_on_close=True)  # Minimal pod spec stored in config
cluster.scale(5)

del cluster
```

```console
$ daskctl cluster list
NAME                        ADDRESS                TYPE         WORKERS  THREADS  MEMORY   CREATED   STATUS
dask-jtomlinson-bffca7fc-7  tcp://localhost:65125  KubeCluster        5        5  5.20 GB  Just now  Running
```

```python
from dask_ctl import get_cluster

cluster = get_cluster("dask-jtomlinson-bffca7fc-7")
```

_Note the address is `localhost` as I used port forwarding in my config. This port will randomly change every time we run `get_cluster` (which is used within `daskctl`). For clusters using a LoadBalancer or NodePort this address will be consistent._

### DIFF
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -1,5 +1,6 @@
 import asyncio
 import copy
+from contextlib import suppress
 import getpass
 import logging
 import os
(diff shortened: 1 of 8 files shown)


jacobtomlinson:
Hopefully this can be replaced with something based on dask/distributed#5033 which would enable use of the scheduler as the key/value store instead of a secret.

jacobtomlinson:
Closing in favour of #468

Issue #311, opened 2021-03-10
Fix port forward collisions


jacobtomlinson:
When using `ClusterIP` outside of the k8s cluster we use `kubectl` to port forward the scheduler service to a random high local port.

https://github.com/dask/dask-kubernetes/blob/858d4b21659702e8ce9d3aa128cfd41a759c4c73/dask_kubernetes/utils.py#L73

However we do not check for port collisions. This comes up occasionally in CI and therefore will also come up in the wild.

We should catch this and try a different port.

Issue #321, opened 2021-03-22
HelmCluster does not work when deployed in cluster that is not defined in current-context


sushi30:
## Description

When deploying the dask helm chart in a namespace other than `default`, HelmCluster fails to initiate with `ConnectionError: kubectl port forward failed`.

## Minimal scenario:
In console

```console
helm repo add dask https://helm.dask.org
helm repo update
helm upgrade -i dask dask/dask --create-namespace -n dask
```
Then in jupyter:

```python
from dask_kubernetes import HelmCluster
cluster = HelmCluster("dask", namespace="dask")
```

```python
---------------------------------------------------------------------------
ConnectionError                           Traceback (most recent call last)
<ipython-input-31-06ad5cb9da3a> in <module>
----> 1 cluster = HelmCluster("dask", namespace="dask")

~/dev/notebooks/.venv/lib/python3.8/site-packages/dask_kubernetes/helm.py in __init__(self, release_name, auth, namespace, port_forward_cluster_ip, loop, asynchronous, scheduler_name, worker_name, node_host, node_port, **kwargs)
    118         if not self.asynchronous:
    119             self._loop_runner.start()
--> 120             self.sync(self._start)
    121 
    122     async def _start(self):

~/dev/notebooks/.venv/lib/python3.8/site-packages/distributed/deploy/cluster.py in sync(self, func, asynchronous, callback_timeout, *args, **kwargs)
    187             return future
    188         else:
--> 189             return sync(self.loop, func, *args, **kwargs)
    190 
    191     def _log(self, log):

~/dev/notebooks/.venv/lib/python3.8/site-packages/distributed/utils.py in sync(loop, func, callback_timeout, *args, **kwargs)
    349     if error[0]:
    350         typ, exc, tb = error[0]
--> 351         raise exc.with_traceback(tb)
    352     else:
    353         return result[0]

~/dev/notebooks/.venv/lib/python3.8/site-packages/distributed/utils.py in f()
    332             if callback_timeout is not None:
    333                 future = asyncio.wait_for(future, callback_timeout)
--> 334             result[0] = yield future
    335         except Exception as exc:
    336             error[0] = sys.exc_info()

~/dev/notebooks/.venv/lib/python3.8/site-packages/tornado/gen.py in run(self)
    760 
    761                     try:
--> 762                         value = future.result()
    763                     except Exception:
    764                         exc_info = sys.exc_info()

[...]

PR #325, opened 2021-03-29
Prevent collisions in port-forward and use namespaces while port-forwarding

MridulS:
fixes https://github.com/dask/dask-kubernetes/issues/311 and https://github.com/dask/dask-kubernetes/issues/321.


jacobtomlinson:
Awesome many thanks!

PR #328, opened 2021-04-08
Updated kubecluster to remove depreciated docs

v2thegreat:
Noticed that the docs showing `Deprecated, please use the KubeCluster constructor directly` [here](https://kubernetes.dask.org/en/latest/kubecluster.html#dask_kubernetes.KubeCluster.from_yaml). There might be other places where issues like this exist, I probably didn't catch them all


jacobtomlinson:
Awesom thanks!

PR #330, opened 2021-04-20
Add type argument to make_pod_spec

sfcdota:
Added `type` argument to `make_pod_spec` function. 

I had an issue using `make_pod_spec` element as `scheduler_pod_template` in `KubeCluster` constructor. My scheduler was unable to start with `ValueError: missing port number in address '$(DASK_SCHEDULER_ADDRESS)'` because `args` list contains elements, that relate to "dask-worker" explicitly.


jacobtomlinson (approved):
Many thanks this is much appreciated.

PR #331, opened 2021-04-21
Better fix for make_pod_spec ValueError case

sfcdota:
Sorry, I was in a hurry on my previous pull request [#330](https://github.com/dask/dask-kubernetes/pull/330) looking for a solution. There was a laconic fix with deletion of a tabulation. Returned the previous state of `make_pod_spec`.

### DIFF
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -525,7 +525,7 @@ class KubeCluster(SpecCluster):
 
         if not self.scheduler_pod_template:
             self.scheduler_pod_template = base_pod_template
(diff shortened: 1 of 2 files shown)


dask_kubernetes/core.py
```diff
@@ -525,7 +525,7 @@ async def _start(self):
 
         if not self.scheduler_pod_template:
             self.scheduler_pod_template = base_pod_template
-            self.scheduler_pod_template.spec.containers[0].args = ["dask-scheduler"]
+        self.scheduler_pod_template.spec.containers[0].args = ["dask-scheduler"]
```
bhperry, line 528: This breaks functionality for people who want to set custom args for the scheduler.

PR #335, opened 2021-05-19
[ci] remove duplicate unit test

jameslamb:
I've been getting familiar with `dask-kubernetes` recently and wanted to make some small contributions.

I found what looks like a duplicated unit test in `tests/test_objects.py`. There are two tests there named `test_extra_container_config_merge()`, and both seem to be testing the exact same thing.

This PR proposes removing one of the duplicate tests.

## Notes for Reviewers

I found this by running `mypy --ignore-missing-imports .`, which yielded a few warnings including the following:

> dask_kubernetes/tests/test_objects.py:92: error: Name 'test_extra_container_config_merge' already defined on line 65

I think it would be useful for maintainers here to set up `mypy` in continuous integration in the future, to catch such issues. In this case it doesn't look like there was any loss of testing coverage because these were testing the same thing. But I've seen other cases where two *different* tests have the same name. In that situation, `pytest` will only run the most-recently-defined one. This can lead to some tests silently be skipped :grimacing: 

Thanks for your time and consideration.

### DIFF
--- a/dask_kubernetes/tests/test_objects.py
+++ b/dask_kubernetes/tests/test_objects.py
@@ -69,33 +69,6 @@ def test_extra_container_config_merge(docker_image, loop):
     with KubeCluster(
         make_pod_spec(
             docker_image,
-            extra_container_config={
-                "env": [{"name": "BOO", "value": "FOO"}],
-                "args": ["last-item"],
-            },
-        ),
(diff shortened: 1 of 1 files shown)


jacobtomlinson (approved):
Thanks!

PR #352, opened 2021-07-28
Explicitly set a host when using NodePort Service

### LINKED ISSUE #348: Ability to set NodePort IP manually
I love the new setup for KubeCluster with the in-cluster Scheduler. We have the same usecase of wanting to run Dask jobs on remote K8s cluster from laptops and I pretty much have it running in the static version. 

Our blocker for transitioning to KubeCluster is that we need to use Nodeports for accessing the scheduler, but our service accounts aren't allowed to have the "list node" privilege. Is there a way to pass in the IP to use for the scheduler, bypassing the NodePort lookup?

andersbogsnes:
Previously, if the service type was a NodePort, KubeCluster would attempt to discover the IP address by listing nodes. This is a set of permissions that not all users will have, so this PR adds the ability to explicitly set a known host.

Closes #348

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -138,3 +138,6 @@ credentials.csv
 *.pub
 *.rdp
 *_rsa
+
+# IDE settings
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -138,6 +138,3 @@ credentials.csv
 *.pub
 *.rdp
 *_rsa
-
-# IDE settings
-.idea
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -152,7 +152,7 @@ class Worker(Pod):
 
 
 class Scheduler(Pod):
-    """A Remote Dask Scheduler controled by Kubernetes
+    """A Remote Dask Scheduler controlled by Kubernetes
     Parameters
     ----------
     idle_timeout: str, optional
@@ -198,7 +198,7 @@ class Scheduler(Pod):
             port=SCHEDULER_PORT,
         )
         self.external_address = await get_external_address_for_scheduler_service(
-            self.core_api, self.service
+            self.core_api, self.service, host=self.kwargs["host"]
         )
 
         self.pdb = await self._create_pdb()
@@ -578,6 +578,7 @@ class KubeCluster(SpecCluster):
                     "idle_timeout": self._idle_timeout,
                     "service_wait_timeout_s": self._scheduler_service_wait_timeout,
                     "pod_template": self.scheduler_pod_template,
+                    "host": self.host,
                     **common_options,
                 },
             }
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -351,8 +351,7 @@ class KubeCluster(SpecCluster):
     >>> pod_spec = make_pod_spec(image='daskdev/dask:latest',
     ...                          memory_limit='4G', memory_request='4G',
     ...                          cpu_limit=1, cpu_request=1,
-    ...                          env={'EXTRA_PIP_PACKAGES': 'fastparquet
-    git+https://github.com/dask/distributed'})
+    ...                          env={'EXTRA_PIP_PACKAGES': 'fastparquet git+https://github.com/dask/distributed'})
     >>> cluster = KubeCluster(pod_spec)
     >>> cluster.scale(10)
 
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -425,7 +425,7 @@ class KubeCluster(SpecCluster):
         security=None,
         scheduler_service_wait_timeout=None,
         scheduler_pod_template=None,
-        scheduler_service_type="ClusterIP",
+        scheduler_service_type=None,
         nodeport_host=None,
         **kwargs
     ):
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -321,8 +321,8 @@ class KubeCluster(SpecCluster):
         Listen address for local scheduler. Only used when deploy_mode == "local". Defaults to
         0.0.0.0
     nodeport_host: str
-        When scheduler_service_type == ``"NodePort"``, override KubeCluster's default behaviour
-        of listing nodes by specifying a nodeport_host.
+        When scheduler_service_type == ``"NodePort"``, select a specific node IP to connect to.
+        When this is not set the node IP will be autodetected.
     port: int
         Port of local scheduler
     auth: List[ClusterAuth] (optional)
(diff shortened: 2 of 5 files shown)


andersbogsnes:
I'm unsure about how to test this implementation, as I can't see a means of getting the node IP address when using pytest-kind. 

Even `docker inspect` gives me a blank IP Adress

.gitignore
```diff
@@ -138,3 +138,6 @@ credentials.csv
 *.pub
 *.rdp
 *_rsa
+
+# IDE settings
+.idea
```
jacobtomlinson, line 143: I usually recommend folks put this kind of thing in a [global gitignore file](https://stackoverflow.com/questions/7335420/global-git-ignore).

dask_kubernetes/core.py
```diff
@@ -578,6 +578,7 @@ async def _start(self):
                     "idle_timeout": self._idle_timeout,
                     "service_wait_timeout_s": self._scheduler_service_wait_timeout,
                     "pod_template": self.scheduler_pod_template,
+                    "host": self.host,
```
jacobtomlinson, line 581: The `host` kwarg and attribute is for setting the host IP when running `deploy_mode="local"`. I think we need to add a new kwarg called something like `nodeport_host`. Perhaps we also need to update the docstring to be clear that `host` is for the local deploy mode only.

jacobtomlinson (commented):
Thanks so much for raising this @andersbogsnes.

I think for testing it will be tricky as the Kubernetes cluster is inside a docker container and I expect it will not be possible to access the nodeport directly. I would perhaps add some check that the `external_address` attribute gets set correctly when the nodeport IP is provided. But I do not expect to actually be able to connect to it.

I've added some comments below because you seem to be reusing the `host` kwarg which is used for a different purpose.

andersbogsnes:
Thanks for the review! I'll focus on testing some verifications that the correct values have been set.

In terms of the API - since we don't want to reuse the host arg, and we can currently only set scheduler type in the config, I see two options.

- Set scheduler-type and NodePort-host in the config
- Have both available as kwargs in the __init__. 

Personal preference is #2, but I fully appreciate the danger of a million-param __init__

jacobtomlinson:
The `host` arg is used for a different purpose. I see not harm in adding an additional kwarg for this. We should also support this in the config.

andersbogsnes:
I've implemented the two keywords nodeport_host and scheduler_service_type. I've modified the test to simulate calling KubeCluster._start() as closely as possible, as actually calling it will cause the test to fail, since it can't connect to the Kind cluster with a NodePort

dask_kubernetes/core.py
```diff
@@ -315,7 +318,11 @@ class KubeCluster(SpecCluster):
     env: Dict[str, str]
         Dictionary of environment variables to pass to worker pod
     host: str
-        Listen address for local scheduler.  Defaults to 0.0.0.0
+        Listen address for local scheduler. Only used when deploy_mode == "local". Defaults to
+        0.0.0.0
+    nodeport_host: str
+        When scheduler_service_type == ``"NodePort"``, override KubeCluster's default behaviour
+        of listing nodes by specifying a nodeport_host.
```
jacobtomlinson, line 325: May I suggest a more user focused docstring here. 

```suggestion
    nodeport_host: str
        When scheduler_service_type == ``"NodePort"``, select a specific node IP to connect to.
        When this is not set the node IP will be autodetected.
```

dask_kubernetes/core.py
```diff
@@ -415,6 +426,8 @@ def __init__(
         security=None,
         scheduler_service_wait_timeout=None,
         scheduler_pod_template=None,
+        scheduler_service_type="ClusterIP",
```
jacobtomlinson, line 429: This should be set to `None` as the default is already set in the configuration YAML.

dask_kubernetes/utils.py
```diff
@@ -53,8 +53,10 @@ async def get_external_address_for_scheduler_service(
         lb = service.status.load_balancer.ingress[0]
         host = lb.hostname or lb.ip
     elif service.spec.type == "NodePort":
-        nodes = await core_api.list_node()
-        host = nodes.items[0].status.addresses[0].address
+        host = nodeport_host
+        if host is None:
+            nodes = await core_api.list_node()
+            host = nodes.items[0].status.addresses[0].address
```
jacobtomlinson, line 59: Small nit to improve readibility.

```suggestion
        if nodeport_host is not None:
            host = nodeport_host
        else:
            nodes = await core_api.list_node()
            host = nodes.items[0].status.addresses[0].address
```

dask_kubernetes/tests/test_async.py
```diff
@@ -390,6 +392,61 @@ async def test_constructor_parameters(k8s_cluster, pod_spec):
...
+        scheduler.kwargs["scheduler_service_type"]
+    )
+
+    external_address = await get_external_address_for_scheduler_service(
+        scheduler.core_api,
+        scheduler.service,
+        nodeport_host=scheduler.kwargs["nodeport_host"],
+    )
+    assert external_address == "tcp://10.10.10.10:8786"
+    assert scheduler.service.spec.type == "NodePort"
```
jacobtomlinson, line 447: I appreciate how in depth you've gone here with this test but given that this had gone so deep into the guts of the library I worry that it will be brittle and break if we make internal changes.

Perhaps it would be best to focus on testing the `get_external_address_for_scheduler_service` in isolation instead.

```suggestion
    from dask_kubernetes.objects import make_service_from_dict
    
    service = make_service_from_dict({"spec": {"ports": [{"port": 8786, "name": "comm"}], "type": "NodePort"}})
    
    external_address = await get_external_address_for_scheduler_service(
        None, service, nodeport_host="10.10.10.10",
    )
    assert external_address == "tcp://10.10.10.10:8786"
```

andersbogsnes: Thanks! Couldn't figure out how to create an instance of the service without going through the cluster - `make_service_from_dict` was the missing piece :smile:

jacobtomlinson:
Hey @andersbogsnes. This PR is still marked as a draft. Is it ready for a full review yet?

andersbogsnes:
@jacobtomlinson Thanks for the ping - had left this one a bit fallow! I believe so, incorporated your previous suggestions, though could never get through the CI tests - they would hang and time out, despite working fine locally

jacobtomlinson:
Looks like a valid failure in CI now related to this PR. @andersbogsnes would you mind taking a look?

jacobtomlinson:
Sorry we never got to merging this. The project has changed a lot since this was raised so going to close it out.

PR #356, opened 2021-08-24
Helm cluster close should raise not warn

jacobtomlinson:


jacobtomlinson:
Raising here breaks the context manager. So I'm not sure this actually makes sense.

PR #368, opened 2021-10-14
Fix the CI

holdenk:
Running pytest locall works for me now (and it wasn't before), but lets make sure this fixes the github actions issue.


jacobtomlinson:
Perhaps we (Dask) should maintain a fork of it.

PR #375, opened 2021-11-15
Update Quickstarts docs to from_yaml()

carderne:
Using the instructions as they are gives:
```
TypeError: Expected a kubernetes.client.V1Pod object, got worker-spec.ymlIf trying to pass a yaml filename then use KubeCluster.from_yaml
```


carderne:
Never mind, I spoke to soon. Accidentally ran on py27 on Google Cloud Shell...

Issue #256, opened 2020-06-13
Kubernetes Operator


mrocklin:
Should we maintain a Kubernetes Operator?

There are a few Kubernetes projects out there (dask-kubernetes, dask-helm, dask-gateway, various internal company projects).  They exist because they all have slightly different needs and serve different use cases.  This makes sense.  In the end it's not terribly difficult to make a new Kubernetes + Dask thing.  However, I'm curious if there are lower level structures that we could build that could be shared among a few of these.

(disclaimer: I suspect that I'll be in this situation again , and so am looking around)

One reason why we keep doing our own logic, rather than using internal k8s controllers is that our scale-up/scale-down logic is more complex than most.  As a result we tend to manage pods directly.  I wonder if putting this logic into an operator might allow more code reuse among projects.  

I don't know this space well though, and it may be that Operators bring in their own baggage that we want to avoid.

As maybe a pragmatic question.  If we had an operator what would dask-kubernetes and dask-helm look like on top of that system?  My guess is that ...

-  For dask-kubernetes we would move a lot of the scaling logic in dask-kubernetes to dask-kubernetes-operator (or some better name) and mostly change this library to just modify values in a custom resource.  We would have to launch one additional pod (one operator per cluster?)
-  For dask-helm we would gain the ability to scale down smoothly I think

jacobtomlinson:
I've been considering the same thing recently. There are positives and negatives to using operators, but I think the scaling logic may make a good case for doing this.

I'd be keen to hear what @yuvipanda thinks.

jcrist:
dask-gateway contains an operator, but its highly tailored to dask-gateway's internals. In my experience, operators as you probably mean them are for the following use case:

- You have a custom resource definition (CRD) for describing your objects. In this case you'd probably have a `DaskCluster` object that would store specs for a dask cluster.
- You want to interact with that object using the k8s api (e.g. use `kubectl` to scale that object).

Dask-Gateway is a weird use case, where we're using k8s objects instead of a database to store state, but we don't want users to interact with dask clusters via the k8s api.

I think an operator for dask could be nice, but they also add complications - you need a new service running on k8s, with SA permissions, it's one-more-thing to watch. I personally don't see a use case in the ecosystem for one at this point.

jacobtomlinson:
Those are good points @jcrist.

I see value in having a CRD with operator handle the scaling in a bespoke way for Dask. It would almost be like a custom HPA which asks the scheduler for information on what pod to remove.

[...]

PR #380, opened 2022-01-11
Experiment with Dask operator

jacobtomlinson:
This was an initial attempt I made last year to build an operator using [kopf](https://kopf.readthedocs.io/en/stable/).

I got distracted with other things and haven't looked at this for a while. Given that we are going to be restarting this work soon I wanted to capture this as something to refer to. But this PR will not be continued.


jacobtomlinson:
xref #256

PR #381, opened 2022-01-12
Poke RTD

jacobtomlinson:
RTD seems to have failed with the latest `main` build a while back but recent PRs have built ok.

Could possibly by a dependency issue that was resolved. Pushing an empty commit to get things moving again.


jacobtomlinson:
Failure is unrelated to RTD, may be a change in the Actions runner as I am unable to reproduce locally.

PR #386, opened 2022-01-20
Removed comment in requirements file

Matt711:


jacobtomlinson:
This was necessary because when installing on windows with `pip install -r requirements-test.txt` it was unhappy with the comment.

```
  fatal: unable to access 'https://codeberg.org/hjacobs/pytest-kind.git Ã‚/': URL using bad/illegal format or missing URL
```

jacobtomlinson (approved):

Issue #385, opened 2022-01-20
Allow scaling of additional groups with HelmCluster


jacobtomlinson:
In dask/helm-chart#222 we added support for additional worker groups.

It would be great if the `HelmCluster.scale()` method took an optional group name to scale.

jacobtomlinson:
Closed in #387

PR #387, opened 2022-01-21
Allow scaling of additional worker groups with HelmCluster (#385)

Matt711:
A solution to this [issue](https://github.com/dask/dask-kubernetes/issues/385). I added an optional group name to `HelmCluster.scale` method. So now users can scale the additional worker groups that they create based on the worker group name.

### DIFF
--- a/dask_kubernetes/helm.py
+++ b/dask_kubernetes/helm.py
@@ -236,10 +236,11 @@ class HelmCluster(Cluster):
 
         return _().__await__()
 
-    def scale(self, n_workers, worker_name=None):
+    def scale(self, n_workers, worker_group=None):
         """Scale cluster to n workers.
 
         This sets the Dask worker deployment size to the requested number.
+        It also allows you to set the worker deployment size of another worker group.
         Workers will not be terminated gracefull so be sure to only scale down
         when all futures have been retrieved by the client and the cluster is idle.
 
@@ -251,22 +252,28 @@ class HelmCluster(Cluster):
         >>> cluster.scale(4)
         >>> cluster
         HelmCluster('tcp://localhost:8786', workers=4, threads=24, memory=24.96 GB)
-
+        >>> cluster.scale(5, worker_group = "high-mem-worker")
+        >>> cluster
+        # TODO: Finish scaling example
         """
-        return self.sync(self._scale, n_workers, worker_name=worker_name)
+        return self.sync(self._scale, n_workers, worker_group=worker_group)
 
-    async def _scale(self, n_workers, worker_name=None):
-        if worker_name:
-            self.worker_name = worker_name
-        await self.apps_api.patch_namespaced_deployment(
-            name=f"{self.release_name}-{self.chart_name}{self.worker_name}",
-            namespace=self.namespace,
-            body={
-                "spec": {
-                    "replicas": n_workers,
-                }
-            },
-        )
+    async def _scale(self, n_workers, worker_group=None):
+        deployment = f"{self.release_name}-{self.chart_name}{self.worker_name}"
+        if worker_group:
+            deployment += f"-{worker_group}"
+        try:
+            await self.apps_api.patch_namespaced_deployment(
+                name=deployment,
+                namespace=self.namespace,
+                body={
+                    "spec": {
+                        "replicas": n_workers,
+                    }
+                },
+            )
+        except kubernetes.client.exceptions.ApiValueError:
+            raise ValueError(f"No such worker group {worker_group}")
 
     def adapt(self, *args, **kwargs):
         """Turn on adaptivity (Not recommended)."""
--- a/dask_kubernetes/helm.py
+++ b/dask_kubernetes/helm.py
@@ -272,8 +272,11 @@ class HelmCluster(Cluster):
                     }
                 },
             )
-        except kubernetes.client.exceptions.ApiValueError:
-            raise ValueError(f"No such worker group {worker_group}")
+        except kubernetes.client.exceptions.ApiException:
+            if worker_group:
+                raise ValueError(f"No such worker group {worker_group}")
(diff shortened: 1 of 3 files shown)


dask_kubernetes/helm.py
```diff
@@ -236,7 +236,7 @@ async def _():
 
         return _().__await__()
 
-    def scale(self, n_workers):
+    def scale(self, n_workers, worker_name=None):
```
Matt711, line 239: @jacobtomlinson Would you prefer me to call this `group_name` instead of `worker_name`?

jacobtomlinson: Yeah I think `group_name` makes sense here.

dask_kubernetes/helm.py
```diff
@@ -253,9 +253,11 @@ def scale(self, n_workers):
...
         """
-        return self.sync(self._scale, n_workers)
+        return self.sync(self._scale, n_workers, worker_name=worker_name)
 
-    async def _scale(self, n_workers):
+    async def _scale(self, n_workers, worker_name=None):
+        if worker_name:
+            self.worker_name = worker_name
         await self.apps_api.patch_namespaced_deployment(
             name=f"{self.release_name}-{self.chart_name}{self.worker_name}",
```
jacobtomlinson, line 262: Given that `worker_name` is a transient variable let's not worry about assigning to self. It might make sense to build the deployment name in two steps. Also we shouldn't forget the `-` between the chart name and worker group name.

```suggestion
        deployment = f"{self.release_name}-{self.chart_name}"
        if worker_name:
            deployment += f"-{worker_name}"
        await self.apps_api.patch_namespaced_deployment(
            name=deployment,
```

jacobtomlinson (changes requested):
Thanks for starting on this!

I agree we should rename `worker_name`. 

We should also think about some error checking. Currently we assume that the worker deployment exists, this is reasonably safe because we were able to connect to it in the first place. But the user could provide any old name here which may or may not exist. Maybe just wrap the `patch_namespaced_deployment` call in a `try/except` and then raise a helpful error from there.

We should also update the examples in the `HelmCluster` docstring to include one that uses this feature.

We also need tests. We could update the test [`values.yaml`](https://github.com/dask/dask-kubernetes/blob/main/dask_kubernetes/tests/helm/values.yaml) file with an additional group but set the replicas to `0` to avoid interfering with other tests. Then extend the [`test_scale_cluster`](https://github.com/dask/dask-kubernetes/blob/af6526e8497715878d6b3c603910d9269e5d5206/dask_kubernetes/tests/test_helm.py#L136) test to also scale this group up and down and also check that scaling a non-existant group raises an error.

dask_kubernetes/helm.py
jacobtomlinson: Yeah I would catch `ApiValueError` and then raise something that is useful to the user that called `scale` like `No such group foo`. Maybe a `ValueError` would be an appropriate thing to raise.

dask_kubernetes/helm.py
```diff
@@ -251,20 +252,28 @@ def scale(self, n_workers):
...
+                name=deployment,
+                namespace=self.namespace,
+                body={
+                    "spec": {
+                        "replicas": n_workers,
+                    }
+                },
+            )
+        except kubernetes.client.exceptions.ApiValueError:
+            raise ValueError(f"No such worker group {worker_group}")
```
jacobtomlinson, line 276: What happens here if this exception is raised and `worker_group` is not set?

We should probably check if `worker_group` is `None` and if so reraise the exception.

dask_kubernetes/helm.py
```diff
@@ -251,20 +252,31 @@ def scale(self, n_workers):
...
+            await self.apps_api.patch_namespaced_deployment(
+                name=deployment,
+                namespace=self.namespace,
+                body={
+                    "spec": {
+                        "replicas": n_workers,
+                    }
+                },
+            )
+        except kubernetes.client.exceptions.ApiException:
```
Matt711, line 275: When testing I got an `ApiException` exception instead of `ApiValueError` exception, so I caught that instead.

dask_kubernetes/helm.py
```diff
@@ -251,20 +252,31 @@ def scale(self, n_workers):
         >>> cluster.scale(4)
         >>> cluster
         HelmCluster('tcp://localhost:8786', workers=4, threads=24, memory=24.96 GB)
-
+        >>> cluster.scale(5, worker_group = "high-mem-worker")
+        >>> cluster
+        # TODO: Finish scaling example
```
Matt711, line 257: Do you mind if I update all the examples, so it will look consistent?

jacobtomlinson: Sure

dask_kubernetes/helm.py
```diff
@@ -251,20 +252,31 @@ def scale(self, n_workers):
...
+                    "spec": {
+                        "replicas": n_workers,
+                    }
+                },
+            )
+        except kubernetes.client.exceptions.ApiException:
+            if worker_group:
+                raise ValueError(f"No such worker group {worker_group}")
+            else:
+                raise kubernetes.client.exceptions.ApiException
```
jacobtomlinson, line 279: Let's not break the traceback chain here. The else is also redundant because raising stops the execution.

```suggestion
        except kubernetes.client.exceptions.ApiException as e:
            if worker_group:
                raise ValueError(f"No such worker group {worker_group}") from e
            raise e
```

jacobtomlinson (commented):
CI seems to be timing out during the `test_scale_cluster` test. Are you able to run it locally?

jacobtomlinson (approved):
I made a couple of tweaks and this should be good to merge now.

Thanks for the effort here @Matt711

PR #389, opened 2022-01-24
[WIP] add execplugin support for AutoRefreshKubeConfigLoader

### LINKED ISSUE #388: Enable AutoRefreshKubeConfigLoader ExecPlugin support
Related to #317 

### Current behavior

The AutoRefreshKubeConfigLoader overwrites KubeConfigLoader following methods:
 - _load_oid_token
 - load_gcp_token

Get the expiration timestamp of the provided token and create a task than run before the token expiration.
The user authentification method with an executable:
```
"exec" in self._user
```
is not supported and the warning in the refresh_after method is not logged because the kubernetes_asyncio KubeConfigLoader._load_from_exec_plugin is not overwritten.

### Desired Behavior

We should be able to get the expiration timestamp from the response of the executable and use it to make the auto refresh supporting executable plugin authentification method.

gregoiregermain:
Resolves #388

@jacobtomlinson This should adapt AutoRefreshKubeConfigLoader to support execPlugin authentification method.

### DIFF
--- a/dask_kubernetes/auth.py
+++ b/dask_kubernetes/auth.py
@@ -12,11 +12,12 @@ import os
 
 import kubernetes
 import kubernetes_asyncio
-
 from kubernetes_asyncio.client import Configuration
-from kubernetes_asyncio.config.kube_config import KubeConfigLoader, KubeConfigMerger
-from kubernetes_asyncio.config.google_auth import google_auth_credentials
 from kubernetes_asyncio.config.dateutil import parse_rfc3339
+from kubernetes_asyncio.config.exec_provider import ExecProvider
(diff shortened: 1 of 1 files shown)


jacobtomlinson (commented):
Thanks for raising this. It generally looks good to me, although I am unfamiliar with the exec method for authentication so am unable to test locally.

The auth section of `dask-kubernetes` is not well tested today. Tests use `kind` and auth via a token so much of this code isn'y covered. @gregoiregermain do you have any thoughts on how we could more robustly test things here?

jacobtomlinson:
Sorry we never got to merging this. The project has changed a lot since this was raised so going to close it out.

PR #392, opened 2022-01-31
Add Dask Operator

### LINKED ISSUE #256: Kubernetes Operator

jacobtomlinson:
Closes #256. This draft PR tracks merging the `dask-operator` feature branch where @Matt711 and I are iterating on a Dask operator. Once we have an MVP we will merge this PR and continue iterating on `main`, but for now this allows us to make smaller mini-PRs into here before doing the big merge.

High level goals of this PR:

- [x] Create some custom resource definitions
  - [ ] ~DaskJob~ Tracked in #483 
  - [x] DaskCluster
  - [x] DaskWorkerGroup
- [x] Write the operator
  - [x] Needs to create/update/delete pods and things on Kubernetes whenever our custom resources are created
  - [x] Package as a container
- [x] Write helm chart to install the CRDs and operator
- [x] Write a Python Cluster manager that creates and deletes instances of the CRDs
  - [ ] ~Implement dask-ctl discovery~ Tracked in #481
- [x] Automate deployment
  - [x] Push docker images to container registry
  - [x] Publish helm chart
- [x] Ensure fully documented
  - [x] Installing the CRDs and operator
  - [x] Creating a cluster via Kubernetes API
  - [x] Creating cluster from Python
  - [x] Connecting to existing cluster
  - [x] Scaling cluster
  - [x] Deleting cluster

### DIFF
--- a/dask_kubernetes/operator/daskcluster.py
+++ b/dask_kubernetes/operator/daskcluster.py
@@ -1,5 +1,4 @@
 import asyncio
-import subprocess
 import threading
 
 from distributed.core import rpc
@@ -9,9 +8,9 @@ import kubernetes_asyncio as kubernetes
 
 from uuid import uuid4
 
+from dask_kubernetes.auth import ClusterAuth
 from dask_kubernetes.utils import (
     get_scheduler_address,
-    check_dependency,
 )
 
 
@@ -181,12 +180,13 @@ def build_cluster_spec(name, image, replicas, resources, env):
     }
 
 
+@kopf.on.startup()
+async def startup(**kwargs):
+    await ClusterAuth.load_first()
+
+
 @kopf.on.create("daskcluster")
 async def daskcluster_create(spec, name, namespace, logger, **kwargs):
-    try:
-        await kubernetes.config.load_kube_config()
-    except kubernetes.config.config_exception.ConfigException:
-        kubernetes.config.load_incluster_config()
     logger.info(
         f"A DaskCluster has been created called {name} in {namespace} with the following config: {spec}"
     )
@@ -196,7 +196,7 @@ async def daskcluster_create(spec, name, namespace, logger, **kwargs):
         # TODO Check for existing scheduler pod
         data = build_scheduler_pod_spec(name, spec.get("image"))
         kopf.adopt(data)
-        scheduler_pod = await api.create_namespaced_pod(
+        await api.create_namespaced_pod(
             namespace=namespace,
             body=data,
         )
@@ -209,23 +209,16 @@ async def daskcluster_create(spec, name, namespace, logger, **kwargs):
         # TODO Check for existing scheduler service
         data = build_scheduler_service_spec(name)
         kopf.adopt(data)
-        scheduler_service = await api.create_namespaced_service(
+        await api.create_namespaced_service(
             namespace=namespace,
             body=data,
         )
-        check_dependency("kubectl")
-        services = subprocess.check_output(
-            [
-                "kubectl",
-                "get",
-                "service",
-                "-n",
-                namespace,
-            ],
-            encoding="utf-8",
-        )
-        while data["metadata"]["name"] not in services:
-            asyncio.sleep(0.1)
+        while True:
+            try:
+                await api.read_namespaced_service(data["metadata"]["name"], namespace)
+                break
+            except Exception:
+                asyncio.sleep(0.1)
         logger.info(
             f"A scheduler service has been created called {data['metadata']['name']} in {namespace} \
             with the following config: {data['spec']}"
@@ -240,7 +233,7 @@ async def daskcluster_create(spec, name, namespace, logger, **kwargs):
         # TODO: Next line is not needed if we can get worker groups adopted by the cluster
         kopf.adopt(data)
         api = kubernetes.client.CustomObjectsApi(api_client)
-        worker_pods = await api.create_namespaced_custom_object(
+        await api.create_namespaced_custom_object(
             group="kubernetes.dask.org",
             version="v1",
             plural="daskworkergroups",
--- a/dask_kubernetes/operator/deployment/Dockerfile
+++ b/dask_kubernetes/operator/deployment/Dockerfile
@@ -10,13 +10,5 @@ WORKDIR /src/dask_kubernetes
 # Install dependencies
 RUN pip install .
 
-# We don't use TARGETARCH so as to support non-buildkit builds
-RUN MAGICARCH=$(dpkg --print-architecture) && \
-    curl -LO "https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/${MAGICARCH}/kubectl" && \
-    mkdir -p /usr/local/bin && \
-    mv ./kubectl /usr/local/bin/kubectl && \
-    chmod +x /usr/local/bin/kubectl && \
-    kubectl version --client
-
 # Start operator
 CMD kopf run -m dask_kubernetes.operator --verbose --all-namespaces


BitTheByte:
Hi @jacobtomlinson 

Do we have an ETA for this?

jacobtomlinson:
We are working towards a hard deadline of the end of May, but hope to merge this and get a first release out much sooner and then iterate in follow up PRs.

dask_kubernetes/operator/daskcluster.py
```diff
@@ -0,0 +1,274 @@
...
+        for wid in worker_ids:
+            worker_pod = api.delete_namespaced_pod(
+                name=wid,
+                namespace=namespace,
+            )
+        logger.info(f"Scaled worker group {name} down to {spec['replicas']} workers.")
+
+
+@kopf.on.delete("daskcluster")
+async def daskcluster_delete(spec, name, namespace, logger, **kwargs):
```
philipp-sontag-by, line 263: Instead of handling the delete manually you could rely on [Kubernetes Owner References](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/) for the cleanup. 
Basically you could set the `daskcluster` resource as an owner for everything else you create here (the pods, services and daskworkergroups) and when it gets deleted kubernetes will take care of the rest.

I think you are already doing this for the native kubernetes objects you create in here (via the `kopf.adopt` calls).

philipp-sontag-by (commented):
I just had a quick look at what you have here and left a couple of notes. 
Looks pretty good already even though it is still early 👍

dask_kubernetes/operator/daskcluster.py
```diff
@@ -0,0 +1,274 @@
...
+        label_selector=f"dask.org/cluster-name={cluster_name},dask.org/component=scheduler",
+        timeout_seconds=60,
+    ):
+        if event["object"].status.phase == "Running":
+            watch.stop()
+        await asyncio.sleep(0.1)
+
+
+@kopf.on.create("daskcluster")
+async def daskcluster_create(spec, name, namespace, logger, **kwargs):
```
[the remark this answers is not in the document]
jacobtomlinson: Thanks, the goal is to switch things to the asyncio version.

dask_kubernetes/operator/daskcluster.py
jacobtomlinson: Yeah we are aware of this. There was a discussion about it in [#406](https://github.com/dask/dask-kubernetes/pull/406#pullrequestreview-885663075). We were having a little trouble getting the adoption to work when scaling up and it was blocking the PR. But we decided to just merge in this for now and figure out the adoption issues later.

jacobtomlinson:
Thanks so much for taking the time to review this @philipp-sontag-by. Really helpful! If it's ok we will ping you when this is closer to being ready for a more thorough review.

philipp-sontag-by:
Sure feel free to ping me if you have questions.

dask_kubernetes/operator/deployment/Dockerfile
```diff
@@ -0,0 +1,14 @@
+# This images needs to be built from the top level of the project
+# $ docker build -t ghcr.io/dask/dask-kubernetes-operator:latest -f dask_kubernetes/operator/deployment/Dockerfile .
+
+FROM python:3.8
+
+# Copy source
+COPY . /src/dask_kubernetes
+WORKDIR /src/dask_kubernetes
+
```
jacobtomlinson, line 9: Need to install `kubectl`.

dask_kubernetes/operator/customresources/daskcluster.yaml
```diff
@@ -0,0 +1,32 @@
...
+    singular: daskcluster
+    shortNames:
+      - daskcluster
+      - dsk
+  versions:
+    - name: v1
+      served: true
+      storage: true
+      schema:
+        openAPIV3Schema:
```
jacobtomlinson, line 20: We need to lock this down to the required feeds.

dask_kubernetes/operator/daskcluster.py
```diff
@@ -0,0 +1,401 @@
...
+    port_forward_cluster_ip = None
+    address = await get_external_address_for_scheduler_service(
+        api, service, port_forward_cluster_ip=port_forward_cluster_ip
+    )
+    return address
+
+
+@kopf.on.create("daskcluster")
+async def daskcluster_create(spec, name, namespace, logger, **kwargs):
+    global SCHEDULER_NAME
```
jacobtomlinson, line 197: We need to switch this out for some other way of handling these references. Perhaps a singleton class?

dask_kubernetes/operator/daskcluster.py
```diff
@@ -0,0 +1,401 @@
...
+        # TODO Check for existing scheduler service
+        data = build_scheduler_service_spec(name)
+        kopf.adopt(data)
+        scheduler_service = await api.create_namespaced_service(
+            namespace=namespace,
+            body=data,
+        )
+        services = subprocess.check_output(
+            [
+                "kubectl",
```
jacobtomlinson, line 228: We should check that `kubectl` is installed.

dask_kubernetes/operator/daskcluster.py
```diff
@@ -0,0 +1,401 @@
...
+            "--patch",
+            str(json_patch),
+            "--type=merge",
+        ],
+        encoding="utf-8",
+    )
+
+
+@kopf.timer("daskworkergroup", interval=5.0)
+async def adapt(spec, name, namespace, logger, **kwargs):
```
jacobtomlinson, line 361: Need to handle multiple clusters.

dask_kubernetes/operator/tests/test_operator.py
```diff
@@ -0,0 +1,161 @@
...
+        cluster_path = os.path.join(DIR, "resources", "simplecluster.yaml")
+        cluster_name = "foo"
+        try:
+            yield cluster_name
+        finally:
+            k8s_cluster.kubectl("delete", "-f", cluster_path, "--wait=true")
+            while "foo-cluster" in k8s_cluster.kubectl("get", "daskclusters"):
+                await asyncio.sleep(0.1)
+
+    yield cm
```
jacobtomlinson, line 139: Needs deleting

dask_kubernetes/operator/tests/test_operator.py
```diff
@@ -0,0 +1,161 @@
...
+@pytest.fixture
+def client(cluster):
+    with Client(cluster) as client:
+        yield client
+
+
+def test_fixtures_kubecluster2(client, cluster):
+    client.scheduler_info()
+    cluster.scale(1)
+    assert client.submit(lambda x: x + 1, 10).result() == 11
```
jacobtomlinson, line 161: We add another test that creates a couple of instances of `KubeCluster2` to ensure they don't tread on each other.

jacobtomlinson (commented):
This is getting really close to MVP levels of ready. I've left a few comments that we absolutely need to resolve. But after that I'm keen to convert most feedback to issues and merge this. We can get an early release out so folks can start playing with it and we can continue working in regular PRs to `main`.

dask_kubernetes/operator/daskcluster.py
Matt711: @philipp-sontag-by "We are now using a single thread that periodically iterates over all dask resources and makes the scaling decision for all of them."

Thank you for your comments! Can you tell me a little more about how you are doing this? I've run into a couple of the issues you listed.

philipp-sontag-by: Basically we have a singleton custom resource that we have defined a `kopf.timer` for. Since this resource only exists once on the cluster we don't run into the worker problems. https://github.com/nolar/kopf/issues/642 has not been an issue for us since kopf introduced the retries. Alternatively you could just start a separate thread yourself in a `@kopf.configure` handler and do it in there.

In that `kopf.timer` handler we then iterate over all Dask resources on the cluster. We make use of [kopf indexing](https://kopf.readthedocs.io/en/stable/indexing/) here so that we don't have to query the API server in each iteration.
We then compute the worker allocation for each resource based on the total amount of resources available and the Dominant Resource Fairness algorithm. The allocation is the applied via a patch API call to our custom Dask resource.

PR #394, opened 2022-02-01
Add Dask Worker Group CRD

Matt711:
As one of the first steps in #392, we need to create some Kubernetes custom resource definitions (CRDs). This CRD is the DaskWorkerGroup that the Operator will use to develop Dask workers.


dask_kubernetes/operator/tests/test_customresources.py
```diff
@@ -1,19 +1,20 @@
...
-    k8s_cluster.kubectl("delete", "-f", *crd_path)
-
-
-def test_customresources(k8s_cluster):
-    assert "daskclusters.kubernetes.dask.org" in k8s_cluster.kubectl("get", "crd")
+import pytest
+
+from glob import glob
+import pathlib
+import os
```
jacobtomlinson, line 5: This diff looks a bit strange given these lines haven't changed.

I suspect this is Windows CRLF vs *nix LF shenanigans. You probably need to tweak your git settings to fix this. [This blog post looks useful.](https://www.aleksandrhovhannisyan.com/blog/crlf-vs-lf-normalizing-line-endings-in-git/)

jacobtomlinson:
Test failures are related to dask-contrib/dask-ctl#25.

dask_kubernetes/operator/customresources/daskworkergroup.yaml
```diff
@@ -0,0 +1,27 @@
...
+      schema:
+        openAPIV3Schema:
+          type: object
+          properties:
+            spec:
+              type: object
+              x-kubernetes-preserve-unknown-fields: true
+            status:
+              type: object
+              x-kubernetes-preserve-unknown-fields: true
```
Matt711, line 27: So now I need to fill out this spec section more. @jacobtomlinson 
```suggestion
            spec:
              type: object
              properties:
                replicas:
                  type: integer
                image:
                  type: string  
              x-kubernetes-preserve-unknown-fields: true
            status:
              type: object
              x-kubernetes-preserve-unknown-fields: true
```

jacobtomlinson: Sure. I'm tempted to suggest we leave things with the `x-kubernetes-preserve-unknown-fields: true` for now and come back and fully nail down the spec later.

Once things are finished we shouldn't have `x-kubernetes-preserve-unknown-fields: true` anywhere (and should probably add a test for that at some point).

PR #397, opened 2022-02-02
Create a scheduler pod when DaskCluster resource is created

Matt711:
As a part of #392, the Dask Operator needs to create/update/delete pods and things on Kubernetes whenever we create our custom resources. This PR is concerned with creating and deleting the scheduler pod when we create or delete the DaskCluster resource, respectively.


Matt711:
Start the operator with `kopf run dask_kubernetes/operator/daskcluster.py  --verbose` and create the DaskCluster resource with `kubectl apply -f dask_kubernetes/operator/tests/resources/simplecluster.yaml`
```(base) mmurray@dgx15:~/dask-kubernetes$ kopf run dask_kubernetes/operator/daskcluster.py
/home/nfs/mmurray/anaconda3/lib/python3.9/site-packages/kopf/_core/reactor/running.py:176: FutureWarning: Absence of either namespaces or cluster-wide flag will become an error soon. For now, switching to the cluster-wide mode for backward compatibility.
  warnings.warn("Absence of either namespaces or cluster-wide flag will become an error soon."
[2022-02-02 10:34:05,852] kopf._core.engines.a [INFO    ] Initial authentication has been initiated.
[2022-02-02 10:34:05,868] kopf.activities.auth [INFO    ] Activity 'login_via_pykube' succeeded.
[2022-02-02 10:34:05,883] kopf.activities.auth [INFO    ] Activity 'login_via_client' succeeded.
[2022-02-02 10:34:05,883] kopf._core.engines.a [INFO    ] Initial authentication has finished.
[2022-02-02 10:34:17,641] kopf.objects         [INFO    ] [default/simple-cluster] A DaskCluster has been created called simple-cluster in default with the following config: {'image': 'daskdev/dask:latest', 'imagePullPolicy': 'IfNotPresent', 'imagePullSecrets': None, 'protocol': 'tcp', 'scheduler': {'env': {}, 'resources': {}, 'serviceType': 'ClusterIP'}}
[2022-02-02 10:34:17,661] kopf.objects         [INFO    ] [default/simple-cluster] Scheduler pod  in default is created
[2022-02-02 10:34:17,680] kopf.objects         [INFO    ] [default/simple-cluster] Scheduler service in default is created
[2022-02-02 10:34:17,680] kopf.objects         [INFO    ] [default/simple-cluster] Handler 'daskcluster_create' succeeded.
[2022-02-02 10:34:17,680] kopf.objects         [INFO    ] [default/simple-cluster] Creation is processed: 1 succeeded; 0 failed.
^C[2022-02-02 10:37:39,942] kopf._core.reactor.r [INFO    ] Signal SIGINT is received. Operator is stopping.
```
Delete the DaskCluster resource with `kubectl delete daskcluster my-cluster` which will also delete the scheduler pod and service

jacobtomlinson:
This is great! I ran your instructions locally and saw that when creating the `DaskCluster` resource it created a scheduler `Pod`. 

Now we need to translate those instructions into a test. If we think about those instructions from the perspective of starting and finishing at a clean and default Kubernetes cluster they would be:

- Create the CRDs
- Start the operator
- Create the example `DaskCluster` resource
- Look at the operator logs to make sure the `Scheduler pod  in default is created` happened
  - We might also want to do `kubectl get pods`  and `kubectl get svc` to make sure our scheduler exists
  - Also we could do `kubectl logs <scheduler pod ID>` to make sure the scheduler is happy
- Delete the cluster resource
  - Check that the scheduler pod and service was also deleted
- Stop the operator
- Delete the CRDs

If you look at the first and last steps in that list they are a mirror. Create the CRDs first, delete the CRDs last. We can do this with a Pytest fixture. The test will run when the fixture yields, so it's a great way to set up and tear down.

```python
@pytest.fixture
def customresources():
    # Create CRDs
    yield
    # Delete CRDs
```

In the test prep PR that I did, I've already created fixtures for [the CRDs](https://github.com/dask/dask-kubernetes/blob/b13f0031f5ea54466dfd8a2bd83656b045430b4f/dask_kubernetes/operator/tests/conftest.py#L10-L15), [the operator](https://github.com/dask/dask-kubernetes/blob/b13f0031f5ea54466dfd8a2bd83656b045430b4f/dask_kubernetes/operator/tests/test_operator.py#L12-L19) and [the example cluster resource](https://github.com/dask/dask-kubernetes/blob/b13f0031f5ea54466dfd8a2bd83656b045430b4f/dask_kubernetes/operator/tests/test_operator.py#L22-L40).

I also added a test that uses those fixtures, so when we run it all the test does is `assert simplecluster`. 

https://github.com/dask/dask-kubernetes/blob/b13f0031f5ea54466dfd8a2bd83656b045430b4f/dask_kubernetes/operator/tests/test_operator.py#L48-L52

A good first step would be to extend that test to check that the scheduler pod and service exists.

dask_kubernetes/operator/tests/test_operator.py
```diff
@@ -31,6 +33,13 @@ async def simplecluster(k8s_cluster, operator):
 
     yield cluster_name
 
+    # Check that the schedeuler pod and service exist
+    scheduler_pod_name = "simple-cluster-scheduler"
+    scheduler_service_name = "simple-cluster"
+
+    assert scheduler_pod_name in k8s_cluster.kubectl("get", "pods")
+    assert scheduler_service_name in k8s_cluster.kubectl("get", "services")
+
```
jacobtomlinson, line 42: This should be in `test_simplecluster`, not the fixture. 

Also it is failing because there isn't enough time for the pod to be created before we do `get pods` here. You could turn each of these into a `while`.

```python
while scheduler_pod_name not in k8s_cluster.kubectl("get", "pods"):
    await asyncio.sleep(0.1)
```

If the pod never gets created the test will timeout at 60 seconds and fail.

jacobtomlinson:
Sorry I accidentally pushed a commit here instead of my own branch, just force pushed to remove it again.

jacobtomlinson (approved):
This is looking good. We've got a nice structure forming here that we can expand on in future PRs.

PR #399, opened 2022-02-03
Fix label assignment for pdb_template

dirkcjelli:
The labels weren't correctly assigned for the pdb which made selecting pdbs belonging to a `KubeCluster` impossible. Functionality is the same as with `_create_service` above.


jacobtomlinson (approved):
Good catch many thanks!

PR #402, opened 2022-02-07
Retry service name resolution socket.getaddrinfo several times

### LINKED ISSUE #401: KubeCluster fails occasionally because it falls back to port forwarding
**What happened**:

I've been creating `KubeCluster` with:

```python
worker_spec = make_pod_spec(
    ...
)

scheduler_spec = make_pod_spec(
    ...
)

KubeCluster(
    pod_template=worker_spec,
    scheduler_pod_template=scheduler_spec,
    namespace=...,
    idle_timeout=600,
)
```

The code above is executed in a pod living in a kubernetes cluster. The pod uses a service account with the required permissions.

Most of the time, this works without any isues. But occasionally it fails with the following stack (sensitive parts ommited):

```
  [...]
  [...]
  File [...] line 98, in _setup_dask_cluster
    self.cluster = KubeCluster(
  File "/usr/local/lib/python3.8/site-packages/dask_kubernetes/core.py", line 466, in __init__
  File "/usr/local/lib/python3.8/site-packages/distributed/deploy/spec.py", line 275, in __init__
    self.sync(self._start)
  File "/usr/local/lib/python3.8/site-packages/distributed/deploy/cluster.py", line 220, in sync
    return sync(self.loop, func, *args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/distributed/utils.py", line 327, in sync
    raise exc.with_traceback(tb)
  File "/usr/local/lib/python3.8/site-packages/distributed/utils.py", line 310, in f
    result[0] = yield future
  File "/usr/local/lib/python3.8/site-packages/tornado/gen.py", line 762, in run
    value = future.result()
  File "/usr/local/lib/python3.8/site-packages/dask_kubernetes/core.py", line 595, in _start
    await super()._start()
  File "/usr/local/lib/python3.8/site-packages/distributed/deploy/spec.py", line 304, in _start
    self.scheduler = await self.scheduler
  File "/usr/local/lib/python3.8/site-packages/distributed/deploy/spec.py", line 59, in _
    await self.start()
  File "/usr/local/lib/python3.8/site-packages/dask_kubernetes/core.py", line 200, in start
    self.external_address = await get_external_address_for_scheduler_service(
  File "/usr/local/lib/python3.8/site-packages/dask_kubernetes/utils.py", line 66, in get_external_address_for_scheduler_service
    port = await port_forward_service(
  File "/usr/local/lib/python3.8/site-packages/dask_kubernetes/utils.py", line 106, in port_forward_service
    raise ConnectionError("kubectl port forward failed")
ConnectionError: kubectl port forward failed
```

Here's what I was able to conclude so far:

- tipically (when no failure occurs, which is most of the time) it [resolves the service name](https://github.com/dask/dask-kubernetes/blob/dd0a112d6f37a84a7b4f0e1f761b53e628e3da01/dask_kubernetes/utils.py#L60-L62), and continues without any problem.
- I suspect occasionally the service resolution fails and it fallsback to trying to connect to the scheduler via [port forwarding](https://github.com/dask/dask-kubernetes/blob/dd0a112d6f37a84a7b4f0e1f761b53e628e3da01/dask_kubernetes/utils.py#L64-L68) using kubectl which is not available (and is expected to not be available).
- I'm not 100% sure of what is the cause for the occasional failure and I may be biased. But it seems to me it often happens when there is a higher resources demand in the cluster and it either: takes longer to start the scheduler pod; or raises insufficient node warning and scales to new nodes (which in turn takes longer to start the scheduler pod).

**My guess**

I'm guessing that if there was some retrial logic [here](https://github.com/dask/dask-kubernetes/blob/dd0a112d6f37a84a7b4f0e1f761b53e628e3da01/dask_kubernetes/utils.py#L60-L62) with a timeout or similar. This would be solved.

Does my reasoning make sense? Happy to make a PR but would need some guidance.

**Environment**:

- Dask version: dask-kubernetes==2021.10.0, dask[complete]==2021.10.0
- Python version: 3.8.10
- Operating System: ubuntu1604
- Install method (conda, pip, source): pip

andrethrill:
Fixes #401

### DIFF
--- a/dask_kubernetes/utils.py
+++ b/dask_kubernetes/utils.py
@@ -41,7 +41,7 @@ def namespace_default():
 
 
 async def get_external_address_for_scheduler_service(
-    core_api, service, port_forward_cluster_ip=None
+    core_api, service, port_forward_cluster_ip=None, service_name_resolution_retries = 20
 ):
     """Take a service object and return the scheduler address."""
     [port] = [
@@ -56,17 +56,29 @@ async def get_external_address_for_scheduler_service(
         nodes = await core_api.list_node()
         host = nodes.items[0].status.addresses[0].address
     elif service.spec.type == "ClusterIP":
-        try:
-            # Try to resolve the service name. If we are inside the cluster this should succeeed.
-            host = f"{service.metadata.name}.{service.metadata.namespace}"
-            socket.getaddrinfo(host, port)
-        except socket.gaierror:
-            # If we are outside it will fail and we need to port forward the service.
+        # Try to resolve the service name. If we are inside the cluster this should succeed.
+        host = f"{service.metadata.name}.{service.metadata.namespace}"
+        success = _is_service_available(host=host, port=port, retries=service_name_resolution_retries)
+
+        # If we are outside it will fail and we need to port forward the service.
+        if not success:
             host = "localhost"
             port = await port_forward_service(
                 service.metadata.name, service.metadata.namespace, port
             )
     return f"tcp://{host}:{port}"
+
+def _is_service_available(host, port, retries=20):
+    success = False
+    while retries > 0:
+        try:
+            socket.getaddrinfo(host, port)
+            success = True
+        except socket.gaierror:
+            time.sleep(0.5)
+            retries -= 1
+
+    return success
 
 
 def _random_free_port(low, high, retries=20):
--- a/dask_kubernetes/utils.py
+++ b/dask_kubernetes/utils.py
@@ -56,14 +56,14 @@ async def get_external_address_for_scheduler_service(
         nodes = await core_api.list_node()
         host = nodes.items[0].status.addresses[0].address
     elif service.spec.type == "ClusterIP":
-        # Try to resolve the service name. If we are inside the cluster this should succeed.
-        host = f"{service.metadata.name}.{service.metadata.namespace}"
-        success = _is_service_available(
-            host=host, port=port, retries=service_name_resolution_retries
-        )
-
-        # If we are outside it will fail and we need to port forward the service.
-        if not success:
+        try:
+            # Try to resolve the service name. If we are inside the cluster this should succeed.
+            host = f"{service.metadata.name}.{service.metadata.namespace}"
+            _is_service_available(
+                host=host, port=port, retries=service_name_resolution_retries
+            )
+        except socket.gaierror:
+            # If we are outside it will fail and we need to port forward the service.
             host = "localhost"
             port = await port_forward_service(
                 service.metadata.name, service.metadata.namespace, port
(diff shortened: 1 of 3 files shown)


dask_kubernetes/utils.py
```diff
@@ -41,7 +41,7 @@ def namespace_default():
 
 
 async def get_external_address_for_scheduler_service(
-    core_api, service, port_forward_cluster_ip=None
+    core_api, service, port_forward_cluster_ip=None, service_name_resolution_retries = 20
```
andrethrill, line 44: I think it makes sense and would like to expose `service_name_resolution_retries` as a config option when instantiating `KubeCluster`. 

If you agree, any suggestion on the best way to do this is appreciated @jacobtomlinson

dask_kubernetes/utils.py
```diff
@@ -56,19 +56,34 @@ async def get_external_address_for_scheduler_service(
...
+    success = False
+    while retries > 0:
+        try:
+            socket.getaddrinfo(host, port)
+            success = True
+        except socket.gaierror:
+            time.sleep(0.5)
+            retries -= 1
+
+    return success
```
jacobtomlinson, line 84: This could be done a little differently to avoid the counting boilerplate. I would also reraise the exception again so that we don't have to change any other logic.

```suggestion
def _is_service_available(host, port, retries=20):
    for i in range(retries):
        try:
            return socket.getaddrinfo(host, port)
        except socket.gaierror as e:
            if i >= retries - 1:
                raise e
            time.sleep(0.5)
```

dask_kubernetes/utils.py
```diff
@@ -56,19 +56,34 @@ async def get_external_address_for_scheduler_service(
...
-        except socket.gaierror:
-            # If we are outside it will fail and we need to port forward the service.
+        # Try to resolve the service name. If we are inside the cluster this should succeed.
+        host = f"{service.metadata.name}.{service.metadata.namespace}"
+        success = _is_service_available(
+            host=host, port=port, retries=service_name_resolution_retries
+        )
+
+        # If we are outside it will fail and we need to port forward the service.
+        if not success:
```
jacobtomlinson, line 66: If `_is_service_available` raises the `socket.gaierror` after retrying then this can stay as it is and we can avoid the temporary `success` variable.

jacobtomlinson (commented):
Thanks for raising this! I've made a few comments.

Dask has a config system that we can use. 

- Default values [go in here](https://github.com/dask/dask-kubernetes/blob/main/dask_kubernetes/kubernetes.yaml). 
- We should also add a [kwarg to the class](https://github.com/dask/dask-kubernetes/blob/dd0a112d6f37a84a7b4f0e1f761b53e628e3da01/dask_kubernetes/core.py#L400). 
- Then set an attribute [like this](https://github.com/dask/dask-kubernetes/blob/main/dask_kubernetes/core.py#L442-L444). 
- The attribute can then be passed on to our the `get_external_address_for_scheduler_service` call.

dask_kubernetes/utils.py
andrethrill: This is great approach. Thanks!

jacobtomlinson (approved):
This looks great thanks! Happy to merge on passing tests.

PR #406, opened 2022-02-15
Add Scaling to the Dask Operator

Matt711:
As a part of https://github.com/dask/dask-kubernetes/pull/392, the Dask Operator needs to scale workers on Kubernetes whenever we create our DaskWorkerGroup custom resources are modified. This PR is concerned with scaling worker pods (manually) when we change the replicas key in a DaskWorkerGroup resource.

### DIFF
--- a/dask_kubernetes/operator/daskcluster.py
+++ b/dask_kubernetes/operator/daskcluster.py
@@ -235,7 +235,7 @@ async def daskworkergroup_update(spec, name, namespace, logger, **kwargs):
         label_selector=f"dask.org/workergroup-name={name}",
     )
     current_workers = len(workers.items)
-    desired_workers = spec["workers"]["replicas"]
+    desired_workers = spec["replicas"]
     workers_needed = desired_workers - current_workers
     if workers_needed > 0:
         for i in range(current_workers + 1, desired_workers + 1):
--- a/dask_kubernetes/operator/daskcluster.py
+++ b/dask_kubernetes/operator/daskcluster.py
@@ -173,7 +173,7 @@ async def daskcluster_create(spec, name, namespace, logger, **kwargs):
         spec.get("resources"),
         spec.get("env"),
     )
-    # kopf.adopt(data) # Unecessary if we can get worker groups adopted by the cluster
+    kopf.adopt(data)  # Unecessary if we can get worker groups adopted by the cluster
     api = kubernetes.client.CustomObjectsApi()
     worker_pods = api.create_namespaced_custom_object(
         group="kubernetes.dask.org",
@@ -256,15 +256,15 @@ async def daskworkergroup_update(spec, name, namespace, logger, **kwargs):
         logger.info(f"Scaled worker group {name} down to {spec['replicas']} workers.")
 
 
-# @kopf.on.delete("daskcluster")
-# async def daskcluster_delete(spec, name, namespace, logger, **kwargs):
-#     api = kubernetes.client.CustomObjectsApi()
-#     workergroups = api.list_cluster_custom_object(
-#         group="kubernetes.dask.org", version="v1", plural="daskworkergroups"
-#     )
-#     workergroups = api.delete_collection_namespaced_custom_object(
-#         group="kubernetes.dask.org",
-#         version="v1",
-#         plural="daskworkergroups",
-#         namespace=namespace,
-#     )
+@kopf.on.delete("daskcluster")
+async def daskcluster_delete(spec, name, namespace, logger, **kwargs):
+    api = kubernetes.client.CustomObjectsApi()
+    workergroups = api.list_cluster_custom_object(
+        group="kubernetes.dask.org", version="v1", plural="daskworkergroups"
+    )
+    workergroups = api.delete_collection_namespaced_custom_object(
+        group="kubernetes.dask.org",
+        version="v1",
+        plural="daskworkergroups",
+        namespace=namespace,
+    )
--- a/dask_kubernetes/operator/daskcluster.py
+++ b/dask_kubernetes/operator/daskcluster.py
@@ -82,14 +82,15 @@ def build_scheduler_service_spec(name):
     }
 
 
-def build_worker_pod_spec(name, namespace, image, n):
+def build_worker_pod_spec(name, namespace, image, n, scheduler_name):
     return {
         "apiVersion": "v1",
         "kind": "Pod",
         "metadata": {
             "name": f"{name}-worker-{n}",
             "labels": {
-                "dask.org/cluster-name": name,
+                "dask.org/cluster-name": scheduler_name,
+                "dask.org/workergroup-name": name,
                 "dask.org/component": "worker",
             },
         },
@@ -98,7 +99,7 @@ def build_worker_pod_spec(name, namespace, image, n):
                 {
                     "name": "scheduler",
                     "image": image,
-                    "args": ["dask-worker", f"tcp://{name}.{namespace}:8786"],
+                    "args": ["dask-worker", f"tcp://{scheduler_name}.{namespace}:8786"],
                 }
             ]
         },
@@ -183,6 +184,7 @@ async def daskcluster_create(spec, name, namespace, logger, **kwargs):
     )
 
 
+@kopf.timer("daskworkergroup", interval=5.0)
 @kopf.on.create("daskworkergroup")
 async def daskworkergroup_create(spec, name, namespace, logger, **kwargs):
     logger.info(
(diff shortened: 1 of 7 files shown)


dask_kubernetes/operator/daskcluster.py
```diff
@@ -80,6 +82,60 @@ def build_scheduler_service_spec(name):
...
+
+def build_worker_group_spec(name, image, workers):
+    return {
+        "apiVersion": "kubernetes.dask.org/v1",
+        "kind": "DaskWorkerGroup",
+        "metadata": {"name": f"{name}-worker-group"},
+        "spec": {
+            "image": image,
+            "workers": {
+                "replicas": workers["replicas"],
```
jacobtomlinson, line 117: I was surprised to see `replicas` nested under `workers. Can we make this a top level thing under `spec`?

dask_kubernetes/operator/daskcluster.py
```diff
@@ -111,3 +167,78 @@ async def daskcluster_create(spec, name, namespace, logger, **kwargs):
...
+        namespace=namespace,
+        body=data,
+    )
+    logger.info(
+        f"A worker group has been created called {data['metadata']['name']} in {namespace} \
+        with the following config: {data['spec']}"
+    )
+
+
+@kopf.timer("daskworkergroup", interval=5.0)
```
jacobtomlinson, line 187: I am curious why this is on a timer?

dask_kubernetes/operator/tests/test_operator.py
```diff
@@ -40,7 +40,7 @@ async def cm():
     yield cm
 
 
-def test_customresources(k8s_cluster):
+def test_customresources(k8s_cluster, gen_cluster):
```
jacobtomlinson, line 43: Why do we need a cluster here?

dask_kubernetes/operator/daskcluster.py
```diff
@@ -111,3 +167,78 @@ async def daskcluster_create(spec, name, namespace, logger, **kwargs):
...
+                namespace=namespace,
+                body=data,
+            )
+        logger.info(f"{workers_needed} Worker pods in created in {namespace}")
+    if workers_needed < 0:
+        pass
+
+
+@kopf.timer("daskworkergroup", interval=5.0)
+async def scale_workers(spec, name, namespace, logger, **kwargs):
```
jacobtomlinson, line 224: I'm not really sure what this is doing.

dask_kubernetes/operator/tests/test_operator.py
Matt711: We don't. I'll get rid of that

dask_kubernetes/operator/daskcluster.py
Matt711: This patching/changing a `DaskWorkerGroup` resource with new data/spec. When the `DaskWorkerGroup` is patched, the `daskworkergroup_create` function will update the number of workers to reflect the new spec.

dask_kubernetes/operator/daskcluster.py
Matt711: I think I added that because the function was not working without it. I'll look at this again and update you.

dask_kubernetes/operator/daskcluster.py
Matt711: Yes we can

dask_kubernetes/operator/daskcluster.py
```diff
@@ -191,14 +196,85 @@ async def daskworkergroup_create(spec, name, namespace, logger, **kwargs):
...
+    )
+
+
+@kopf.on.delete("daskworkergroup")
+async def daskworkergroup_delete(spec, name, namespace, logger, **kwargs):
+    api = kubernetes.client.CoreV1Api()
+    workers = api.delete_collection_namespaced_pod(
+        namespace=namespace,
+        label_selector=f"dask.org/workergroup-name={name}",
+    )
```
jacobtomlinson, line 280: I don't think we need this. When we use `kopf.adopt` that ties the resources together and when we create the top level `DaskCluster` Kubernetes cleans everything up for us.

jacobtomlinson (commented):
This is looking nice! I'll give it a test this morning.

We should add some tests here. A small test would probably be to create the cluster, wait for the default number of workers, patch the resource to scale up, wait for the number of workers, patch the resource to scale down, wait for the number of workers.

dask_kubernetes/operator/daskcluster.py
```diff
@@ -187,18 +192,79 @@ async def daskcluster_create(spec, name, namespace, logger, **kwargs):
...
+    #     name=scheduler_name, image=scheduler_spec.get("image")
+    # )
+    # data = build_worker_group_spec(
+    #     name.split('-')[0],
+    #     spec.get("image"),
+    #     spec.get("replicas"),
+    #     spec.get("resources"),
+    #     spec.get("env"),
+    # )
+    # kopf.adopt(data, owner = cluster["items"][0])
```
Matt711, line 210: Having trouble getting the `DaskWorkerGroup` to become adopted by the `DaskCluster`. The `owner` executing without any errors and when I check log the `DaskWorkerGroup` I get this:
```
{
   "apiVersion":"kubernetes.dask.org/v1",
   "kind":"DaskWorkerGroup",
   "metadata":{
      "name":"default-worker-group",
      "ownerReferences":[
         {
            "controller":true,
            "blockOwnerDeletion":true,
            "apiVersion":"kubernetes.dask.org/v1",
            "kind":"DaskCluster",
            "name":"simple-cluster",
            "uid":"a65120a5-ce6f-485f-acbc-513645582d16"
         }
      ],
      "namespace":"default",
      "labels":{
         
      }
   },
   "spec":{
      "image":"daskdev/dask:latest",
      "replicas":3,
      "resources":{
         
      },
      "env":{
         
      }
   }
}
```
Which is what I expect. The `DaskWorkerGroup` resource is now a child of the `DaskCluster`. But when I delete the `DaskCluster` resource, the `DaskWorkerGroup`'s are not deleted. Therefore, the worker pods are not deleted either. It seems like `kopf.adopt` needs to be called inside a handler that pertains to the parent resource eg. `@kopf.oncreate("daskcluster")`. Maybe the `DaskCluster` resource should also hold a list of names of the `DaskWorkerGroup`s as they are added to the cluster. So we would need an update handler. And as `DaskWorkerGroup`s are added, the update handler would adopt them.

Matt711: Maybe the `DaskCluster` resource should also hold a list of names of the `DaskWorkerGroups`. The names would be added to the `DaskCluster` in `daskworkergroup_create` handler. This would update the `DaskCluster`, so we would also need a `daskcluster_update` handler. And as `DaskWorkerGroups` are added, the update handler would adopt them.

Matt711: @jacobtomlinson

dask_kubernetes/operator/tests/test_operator.py
```diff
@@ -77,6 +77,12 @@ async def test_simplecluster(k8s_cluster, kopf_runner, gen_cluster):
                 async with Client(
                     f"tcp://localhost:{port}", asynchronous=True
                 ) as client:
+                    k8s_cluster.port_forward(
+                        "scale",
+                        "--replicas=5",
+                        "daskworkergroup",
+                        "default-worker-group",
+                    )
```
jacobtomlinson, line 85: I think we should make a separate test for this instead of extending the existing test.

dask_kubernetes/operator/daskcluster.py
jacobtomlinson: Hrm that's frustrating. I think we should definitely keep persuing the adoption method rather than having deletion handlers.

Perhaps this is a good time to built a small separate experiment with a couple of CRDs and an operator and try and understand the adoption mechanism better.

dask_kubernetes/operator/daskcluster.py
```diff
@@ -82,14 +82,15 @@ def build_scheduler_service_spec(name):
     }
 
 
-def build_worker_pod_spec(name, namespace, image, n):
+def build_worker_pod_spec(name, namespace, image, n, scheduler_name):
     return {
         "apiVersion": "v1",
         "kind": "Pod",
         "metadata": {
             "name": f"{name}-worker-{n}",
```
jacobtomlinson, line 90: I think this name should also include the `scheduler_name` otherwise we will have collisions with multiple clusters.

dask_kubernetes/operator/daskcluster.py
```diff
@@ -187,18 +193,67 @@ async def daskcluster_create(spec, name, namespace, logger, **kwargs):
...
+    api = kubernetes.client.CustomObjectsApi()
+    workergroups = api.list_cluster_custom_object(
+        group="kubernetes.dask.org", version="v1", plural="daskworkergroups"
+    )
+    workergroups = api.delete_collection_namespaced_custom_object(
+        group="kubernetes.dask.org",
+        version="v1",
+        plural="daskworkergroups",
+        namespace=namespace,
+    )
```
jacobtomlinson, line 259: As we already discussed it would be nice to not need this. Let's get this in and then worry about removing this again.

dask_kubernetes/operator/tests/resources/simplecluster.yaml
```diff
@@ -16,7 +16,8 @@ spec:
...
-  workers:
-    replicas: 3
-    resources: {}
-    env: {}
\ No newline at end of file
+  replicas: 3
+  resources: {}
+  env: {}
+status:
+  replicas: 3
```
jacobtomlinson, line 23: I don't think we should be setting any `status` config here, this should be generated by Kubernetes.

dask_kubernetes/operator/tests/resources/simpleworkergroup.yaml
```diff
@@ -6,12 +6,13 @@ spec:
...
+  replicas: 2
+  resources: {}
+  env: {}
+  # nodeSelector: null
+  # securityContext: null
+  # affinity: null
+  # tolerations: null
+  # serviceAccountName: null
+status:
+  replicas: 2
```
jacobtomlinson, line 18: Same here

dask_kubernetes/operator/tests/test_operator.py
```diff
@@ -35,7 +35,7 @@ async def cm():
             yield cluster_name
         finally:
             # Delete cluster resource
-            k8s_cluster.kubectl("delete", "-f", cluster_path)
+            k8s_cluster.kubectl("delete", "dsk", "--all")
```
jacobtomlinson, line 38: I think we should revert this. The test should only delete the cluster that it created.

jacobtomlinson (commented):
This is awesome! I just had a play around on the command line creating clusters, scaling them, deleting them, connecting to them, etc. It works so nicely and feels so Kubernetes native!

I've left a couple of last comments that could do with being addressed, but this is 99% of the way there and we can probably get this merged today.

PR #411, opened 2022-02-24
Scale Dask clusters using Scheduler information

Matt711:
As a part of https://github.com/dask/dask-kubernetes/pull/392, the Dask Operator needs to scale workers on Kubernetes whenever we create our DaskWorkerGroup custom resources are modified. This PR is concerned with scaling workers using information from the scheduler (like what workers are idle and can be deleted when scaling down).


Matt711:
1. Start the operator with `kopf run dask_kubernetes/operator/daskcluster.py` 
2. Create the DaskCluster resource with `kubectl apply -f dask_kubernetes/operator/tests/resources/simplecluster.yaml`
3. Create an additional worker group with `kubectl apply -f dask_kubernetes/operator/tests/resources/simpleworkergroup.yaml`
Looking at the cluster:
```
(base) mmurray@dgx15:~/dask-kubernetes$ kubectl get all
NAME                                   READY   STATUS    RESTARTS   AGE
pod/additional-worker-group-worker-1   1/1     Running   0          19s
pod/additional-worker-group-worker-2   1/1     Running   0          19s
pod/default-worker-group-worker-1      1/1     Running   0          63s
pod/default-worker-group-worker-2      1/1     Running   0          63s
pod/default-worker-group-worker-3      1/1     Running   0          63s
pod/simple-cluster-scheduler           1/1     Running   0          64s
```
4. Scale the worker groups with `kubectl scale --replicas=5 daskworkergroup additional-worker-group` and `kubectl scale --replicas=7 daskworkergroup default-worker-group`
Looking at the cluster:
```
(base) mmurray@dgx15:~/dask-kubernetes$ kubectl get all
NAME                                   READY   STATUS    RESTARTS   AGE
pod/additional-worker-group-worker-1   1/1     Running   0          8m
pod/additional-worker-group-worker-2   1/1     Running   0          8m
pod/additional-worker-group-worker-3   1/1     Running   0          97s
pod/additional-worker-group-worker-4   1/1     Running   0          97s
pod/additional-worker-group-worker-5   1/1     Running   0          97s
pod/default-worker-group-worker-1      1/1     Running   0          8m44s
pod/default-worker-group-worker-2      1/1     Running   0          8m44s
pod/default-worker-group-worker-3      1/1     Running   0          8m44s
pod/default-worker-group-worker-4      1/1     Running   0          41s
pod/default-worker-group-worker-5      1/1     Running   0          41s
pod/default-worker-group-worker-6      1/1     Running   0          41s
pod/default-worker-group-worker-7      1/1     Running   0          41s
pod/simple-cluster-scheduler           1/1     Running   0          8m45s

NAME                     TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)             AGE
service/kubernetes       ClusterIP   10.96.0.1      <none>        443/TCP             15d
service/simple-cluster   ClusterIP   10.96.116.80   <none>        8786/TCP,8787/TCP   8m45s
```

dask_kubernetes/operator/daskcluster.py
```diff
@@ -237,9 +238,12 @@ async def daskworkergroup_update(spec, name, namespace, logger, **kwargs):
             )
         logger.info(f"Scaled worker group {name} up to {spec['replicas']} workers.")
     if workers_needed < 0:
+        with rpc(f"tcp://{cluster_name}.{namespace}:8786") as scheduler:
+            scheduler_info = scheduler
+        logger.info(f"Scheduler info: {scheduler_info}")
```
Matt711, line 243: Is this the right way to connect to the scheduler with dask rpc? In the logs I get this:
```
[2022-02-24 07:26:24,129] kopf.objects         [INFO    ] [default/default-worker-group] Scheduler info: <rpc to 'tcp://10.244.0.46:8786', 0 comms>
```
We shouldn't expect there to be zero comms, right?

dask_kubernetes/operator/daskcluster.py
```diff
@@ -237,6 +239,9 @@ async def daskworkergroup_update(spec, name, namespace, logger, **kwargs):
             )
         logger.info(f"Scaled worker group {name} up to {spec['replicas']} workers.")
     if workers_needed < 0:
+        # with rpc(f"tcp://{scheduler_name}.{namespace}:8786") as scheduler:
+        #     scheduler_info = scheduler
```
jacobtomlinson, line 243: So this should be fine. The `scheduler` here would be an RPC representation of the scheduler where you can asynchronously invoke any of the [exposed handlers](https://github.com/dask/distributed/blob/a364d426b0565f5cce17bf814a576a8fd84206af/distributed/scheduler.py#L3893-L3973).

```python
from distributed.core import rpc
with rpc("localhost:8786") as scheduler:
    worker_ids = await scheduler.workers_to_close(n=5)
```

The above code tells the scheduler that you want to remove `5` workers. The scheduler would select 5 workers to remove, move any futures they are holding onto different workers and then return a list of the IDs of the workers that are ready for deletion. We can then use that list to select which pods to delete.

jacobtomlinson: In `build_worker_pod_spec` we probaby want to ensure that we set the `--name` flag on each worker with a name that matches the pod name, because this is the ID that the `workers_to_close` method will give us. Should make it easier to delete the right pods that way.

dask_kubernetes/operator/daskcluster.py
jacobtomlinson: That's just because you haven't used the RPC yet so no comms have been opened.

dask_kubernetes/operator/daskcluster.py
```diff
@@ -99,7 +104,11 @@ def build_worker_pod_spec(name, namespace, image, n, scheduler_name):
                 {
                     "name": "scheduler",
                     "image": image,
-                    "args": ["dask-worker", f"tcp://{scheduler_name}.{namespace}:8786"],
+                    "args": [
+                        "dask-worker",
+                        f"tcp://{scheduler_name}.{namespace}:8786",
+                        f"--name={scheduler_name}-{name}-worker-{n}",
```
jacobtomlinson, line 110: The name is already being formatted on line 95. It would be best to assign it to a variable and reuse it in both places to avoid one getting updated and not the other in the future.

dask_kubernetes/operator/daskcluster.py
```diff
@@ -237,9 +245,15 @@ async def daskworkergroup_update(spec, name, namespace, logger, **kwargs):
             )
         logger.info(f"Scaled worker group {name} up to {spec['replicas']} workers.")
     if workers_needed < 0:
-        for i in range(current_workers, desired_workers, -1):
+        nest_asyncio.apply()
```
jacobtomlinson, line 248: I am curious why this needs to be done?

dask_kubernetes/operator/daskcluster.py
```diff
@@ -237,9 +245,15 @@ async def daskworkergroup_update(spec, name, namespace, logger, **kwargs):
...
+        with rpc("localhost:8786") as scheduler:
```
jacobtomlinson, line 249: This should be the scheduler service name, not `localhost`.

jacobtomlinson (commented):
Some general comments.

dask_kubernetes/operator/daskcluster.py
Matt711: The library [`nest-asyncio`](https://github.com/erdewit/nest_asyncio) allows for nested event loops in `asyncio`. I wondered if you another of doing this without it?

dask_kubernetes/operator/daskcluster.py
Matt711: The argument the workers use is `f"tcp://{scheduler_name}.{namespace}:8786"` but that  failed.

Matt711:
The test for building the documentation is failing ```AttributeError: module 'functools' has no attribute 'cached_property'```. Since this wasn't failing before, it might have to do with the `nest-asyncio` library I added as a dependency. But anyway, it looks like the `cached_property` was added in python 3.8, and docs are built with 3,7?

dask_kubernetes/operator/daskcluster.py
jacobtomlinson: I guess I'm asking why we need to nest the event loop. Everything should just be able to use the default one. Kopf suppoers asyncio and you are already inside an async function.

dask_kubernetes/operator/daskcluster.py
jacobtomlinson: That is what I expected to be used. I think it might fail in development and when running tests because kopf is running outside the cluster. But in practice kopf will be inside the cluster and able to connect. I'm not immediately sure what the right thing is to do here.

jacobtomlinson:
I think the docs issues may be because we dropped Python `3.7` in Dask as a whole recently. We should bump the docs to build on a more recent version. I'll raise a PR.

dask_kubernetes/operator/daskcluster.py
Matt711: I'll add a TODO for it.

jacobtomlinson (commented):
This generally seems great. We may need to refine this a little but let's do that in future PRs.

PR #413, opened 2022-03-07
Add a cluster manager that supports that Dask Operator

Matt711:
As a part of https://github.com/dask/dask-kubernetes/pull/392, the Dask Operator needs a python cluster manager create and delete kubernetes CRDs on Kubernetes. This PR is concerned withcreating the python cluster manager class. The new cluster manager that we're calling KubeCluster2 (for now) will subclass Cluster (from distributed). And will overwrite a lot of its methods, supporting the work we've done with the [Operator](https://github.com/dask/dask-kubernetes/pull/392).


dask_kubernetes/operator/core.py
```diff
@@ -0,0 +1,52 @@
...
+    def __init__(self, name, loop=None, asynchronous=False, **kwargs):
+        self.name = name
+        self.core_api = None
+        self.scheduler_comm = None
+
+        status = subprocess.run(
+            ["kopf", "run", ".daskcluster.py"],
+            capture_output=True,
+            encoding="utf-8",
+        )
```
jacobtomlinson, line 28: We shouldn't be running the operator here. That should be running already on our cluster independently of what Dask clusters we create.

jacobtomlinson (commented):
Some quick thoughts.

This looks inspired by the `HelmCluster` which is a great starting point but a slightly different paradigm. In `HelmCluster` we are always connecting to an existing cluster, because they can only be created by `helm`.

In this cluster manager we want two different ways of creating things, one where we create the cluster, and another where we connect to an existing cluster.

I would expect if someone did `cluster = KubeCluster2(name="foo", ...)` it will create a `DaskCluster` resource called `foo` via the Kubernetes API, wait for the service to become available and then connect an RPC to it. This should fail if the cluster already exists.

I would also expect folks to be able to run `cluster = KubeCluster2.from_name("foo")` which skips creating the resources and just connects to the existing cluster. This should fail if this doesn't already exist.

dask_kubernetes/operator/core.py
```diff
@@ -0,0 +1,88 @@
...
+        self.custom_api = None
+        self.image = image
+        self.replicas = replicas
+        self.resources = resources
+        self.env = env
+
+        super().__init__(asynchronous=asynchronous, **kwargs)
+        if not self.asynchronous:
+            self._loop_runner.start()
+            self.sync(self._start)
```
Matt711, line 48: This is failing with `NewConnectionError`. Has this happened to you before @jacobtomlinson?

jacobtomlinson: I don't think so. Which module does`NewConnectionError` come from?

jacobtomlinson: Hmm no idea. Perhaps this is happening somewhere inside `_start` and by running it async like this is causing the traceback to report the wrong line. This could well be related to the `kubernetes` API calls.

dask_kubernetes/operator/core.py
```diff
@@ -0,0 +1,88 @@
...
+        data = build_cluster_spec(
+            self.name, self.image, self.replicas, self.resources, self.env
+        )
+        cluster = self.custom_api.create_namespaced_custom_object(
+            group="kubernetes.dask.org",
+            version="v1",
+            plural="daskclusters",
+            namespace=self.namespace,
+            body=data,
+        )
```
Matt711, line 62: The `NewConnectionError` is coming from here. I'm not sure why I can't create custom resources here, but I can [here](https://github.com/dask/dask-kubernetes/pull/392/files#diff-d63febde65b9397396016375ad5bee42b517b65a423e69918aebac8c6c0b07a9R187-R193).

Matt711: I wonder if it can't connect to the Kubernetes API. Something to do with kubeconfig, maybe?

jacobtomlinson: This might be important. I don't see it here.

https://github.com/dask/dask-kubernetes/blob/2aa7212cc7fbffd25caa8cf474d950dd876821d0/dask_kubernetes/core.py#L561

dask_kubernetes/operator/core.py
```diff
@@ -0,0 +1,170 @@
...
+                "kubectl",
+                "apply",
+                "-f",
+                "data.json",
+                "-n",
+                self.namespace,
+            ],
+            encoding="utf-8",
+        )
+        time.sleep(5)
```
Matt711, line 65: Waits for the pods to be in Running state. Not ideal, but I could not get it to work with `while "Running" not in kubectl ...`

dask_kubernetes/operator/core.py
```diff
@@ -0,0 +1,170 @@
...
+                "8786:8786",
+                "-n",
+                f"{self.namespace}",
+            ],
+            stdin=subprocess.DEVNULL,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
+            start_new_session=True,
+        )
+        time.sleep(2)
```
Matt711, line 80: Similar thing here. Waits for port-forward service to start. Not sure how to do this a better way.

Matt711:
```
>>> cluster
KubeCluster2(foo, 'tcp://localhost:8786', workers=2, threads=160, memory=1.97 TiB)
>>> cluster.scale('default', 7)
>>> cluster
KubeCluster2(foo, 'tcp://localhost:8786', workers=7, threads=560, memory=6.89 TiB)
>>> cluster.scale('default', 3)
>>> cluster
KubeCluster2(foo, 'tcp://localhost:8786', workers=3, threads=240, memory=2.95 TiB) 
```

Matt711:
```
>>> cluster.add_worker_group('additional')
```
image

dask_kubernetes/operator/core.py
```diff
@@ -0,0 +1,167 @@
...
+                "apply",
+                "-f",
+                "data.json",
+                "-n",
+                self.namespace,
+            ],
+            encoding="utf-8",
+        )
+
+    def scale(self, worker_group, n):
```
Matt711, line 145: Switch argument order and make make `worker_group` a keyword arg

dask_kubernetes/operator/core.py
```diff
@@ -0,0 +1,167 @@
...
+                        f"{name}",
+                        "-n",
+                        self.namespace,
+                    ]
+                )
+            )
+            logs[name] = log
+        return logs
+
+    def add_worker_group(self, name, n=3):
```
Matt711, line 129: Switch argument order and make make `worker_group` a keyword arg.

dask_kubernetes/operator/core.py
```diff
@@ -0,0 +1,167 @@
...
+)
+
+
+class KubeCluster2(Cluster):
+    """Launch a Dask Cluster on Kubernetes using the Operator"""
+
+    def __init__(
+        self,
+        name,
+        namespace="default",
```
jacobtomlinson, line 24: I wonder if we can set this to `None` and use whatever namespace they have set in their context if they don't set it.

dask_kubernetes/operator/core.py
```diff
@@ -0,0 +1,167 @@
...
+        image="daskdev/dask:latest",
+        replicas=3,
```
jacobtomlinson, line 26: I would be tempted to call this `n_workers` to be consistent with other cluster managers.

dask_kubernetes/operator/core.py
```diff
@@ -0,0 +1,167 @@
...
+        super().__init__(asynchronous=asynchronous, **kwargs)
+        if not self.asynchronous:
+            self._loop_runner.start()
+            self.sync(self._start)
+
+    async def _start(self):
+        data = build_cluster_spec(
+            self.name, self.image, self.replicas, self.resources, self.env
+        )
+        with open("data.json", "w") as jfile:
```
jacobtomlinson, line 53: We should dump this to a [named temporary file](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile).

dask_kubernetes/operator/core.py
jacobtomlinson: I don't feel confident merging with a sleep in here like this. I know we are going to fix some things up later but this is probably going to be very brittle. 

Could we use `kubectl wait` here instead? 

https://stackoverflow.com/questions/53536907/kubectl-wait-for-condition-complete-timeout-30s

dask_kubernetes/operator/core.py
```diff
@@ -0,0 +1,167 @@
...
+                "apply",
+                "-f",
+                "data.json",
+                "-n",
+                self.namespace,
+            ],
+            encoding="utf-8",
+        )
+        time.sleep(5)
+        pf = subprocess.Popen(
```
jacobtomlinson, line 67: Can we reuse [get_external_address_for_scheduler](https://github.com/dask/dask-kubernetes/blob/2aa7212cc7fbffd25caa8cf474d950dd876821d0/dask_kubernetes/utils.py#L43) here instead?

dask_kubernetes/operator/core.py
```diff
@@ -0,0 +1,167 @@
...
+                        self.namespace,
+                    ]
+                )
+            )
+            logs[name] = log
+        return logs
+
+    def add_worker_group(self, name, n=3):
+        data = build_worker_group_spec(name, self.image, n, self.resources, self.env)
+        with open("data.json", "w") as jfile:
```
jacobtomlinson, line 131: Same with a tempfile here.

dask_kubernetes/operator/core.py
jacobtomlinson: This should be compatible with other `scale` commands. If there is only one worker group then `scale(5)` should still work. I think this should be an optional kwarg.

```suggestion
    def scale(self, n, workergroup=None):
```

dask_kubernetes/operator/daskcluster.py
```diff
@@ -157,6 +172,7 @@ async def daskcluster_create(spec, name, namespace, logger, **kwargs):
         namespace=namespace,
         body=data,
     )
+    # await wait_for_scheduler(name, namespace)
```
jacobtomlinson, line 175: Why is this commented?

dask_kubernetes/operator/core.py
Matt711: And if they don't have is set in their context, should we go with 'default'? @jacobtomlinson

Matt711:
Example instantiating a `distributed.Client`
```
>>> from dask_kubernetes.operator.core import KubeCluster2
>>> cluster  = KubeCluster2(name='foo')
/home/nfs/mmurray/anaconda3/lib/python3.9/site-packages/distributed/client.py:1265: VersionMismatchWarning: Mismatched versions found

+-------------+-----------+-----------+---------+
| Package     | client    | scheduler | workers |
+-------------+-----------+-----------+---------+
| blosc       | None      | 1.10.2    | None    |
| dask        | 2022.01.1 | 2022.02.1 | None    |
| distributed | 2022.01.1 | 2022.2.1  | None    |
| lz4         | None      | 3.1.10    | None    |
| msgpack     | 1.0.2     | 1.0.3     | None    |
| toolz       | 0.11.1    | 0.11.2    | None    |
+-------------+-----------+-----------+---------+
Notes: 
-  msgpack: Variation is ok, as long as everything is above 0.6
  warnings.warn(version_module.VersionMismatchWarning(msg[0]["warning"]))
>>> cluster
KubeCluster2(foo, 'tcp://localhost:61957', workers=3, threads=240, memory=2.95 TiB)
>>> from dask.distributed import Client
>>> client = Client(cluster)
/home/nfs/mmurray/anaconda3/lib/python3.9/site-packages/distributed/client.py:1265: VersionMismatchWarning: Mismatched versions found

+-------------+---------------+----------------+----------------+
| Package     | client        | scheduler      | workers        |
+-------------+---------------+----------------+----------------+
| blosc       | None          | 1.10.2         | 1.10.2         |
| dask        | 2022.01.1     | 2022.02.1      | 2022.02.1      |
| distributed | 2022.01.1     | 2022.2.1       | 2022.2.1       |
| lz4         | None          | 3.1.10         | 3.1.10         |
| msgpack     | 1.0.2         | 1.0.3          | 1.0.3          |
| numpy       | 1.20.3        | 1.21.1         | 1.21.1         |
| pandas      | 1.3.4         | 1.3.0          | 1.3.0          |
| python      | 3.9.7.final.0 | 3.8.12.final.0 | 3.8.12.final.0 |
| toolz       | 0.11.1        | 0.11.2         | 0.11.2         |
+-------------+---------------+----------------+----------------+
Notes: 
-  msgpack: Variation is ok, as long as everything is above 0.6
  warnings.warn(version_module.VersionMismatchWarning(msg[0]["warning"]))
>>> from pprint import pprint as pp
>>> pp(client.scheduler_info())
{'address': 'tcp://10.244.0.22:8786',
 'id': 'Scheduler-2d8396f8-60a8-49ab-aba6-d962049cc182',
 'services': {'dashboard': 8787},
 'started': 1647652462.7280612,
 'type': 'Scheduler',
 'workers': {'tcp://10.244.0.23:36795': {'host': '10.244.0.23',
                                         'id': 'foo-cluster-default-worker-group-worker-a7dc2ef1f2be430991edb0bcb9877143',
                                         'last_seen': 1647652673.7929082,
                                         'local_directory': '/dask-worker-space/worker-a7_m9f1h',
                                         'memory_limit': 1082106646528,
                                         'metrics': {'bandwidth': {'total': 100000000,
                                                                   'types': {},
                                                                   'workers': {}},
                                                     'cpu': 0.0,
                                                     'executing': 0,
                                                     'in_flight': 0,
                                                     'in_memory': 0,
                                                     'memory': 120635392,
                                                     'num_fds': 25,
                                                     'read_bytes': 549.3930131857151,
                                                     'read_bytes_disk': 0.0,
                                                     'ready': 0,
                                                     'spilled_nbytes': {'disk': 0,
                                                                        'memory': 0},
                                                     'time': 1647652673.791038,
                                                     'write_bytes': 1256.6116556138718,
                                                     'write_bytes_disk': 515526.43006017234},
                                         'name': 'foo-cluster-default-worker-group-worker-a7dc2ef1f2be430991edb0bcb9877143',
                                         'nanny': 'tcp://10.244.0.23:36525',
                                         'nthreads': 80,
                                         'resources': {},
                                         'services': {'dashboard': 34903},
                                         'type': 'Worker'},
             'tcp://10.244.0.24:46283': {'host': '10.244.0.24',
                                         'id': 'foo-cluster-default-worker-group-worker-11a4cef58d454b35aa35664380679798',
                                         'last_seen': 1647652673.9110675,
                                         'local_directory': '/dask-worker-space/worker-7ncwlbo2',
                                         'memory_limit': 1082106646528,
                                         'metrics': {'bandwidth': {'total': 100000000,
                                                                   'types': {},
                                                                   'workers': {}},
                                                     'cpu': 0.0,
                                                     'executing': 0,
                                                     'in_flight': 0,
                                                     'in_memory': 0,
                                                     'memory': 119832576,
                                                     'num_fds': 25,
                                                     'read_bytes': 286.1724824990648,
                                                     'read_bytes_disk': 0.0,
                                                     'ready': 0,
                                                     'spilled_nbytes': {'disk': 0,
                                                                        'memory': 0},
                                                     'time': 1647652673.9092755,
                                                     'write_bytes': 994.5994671470995,
                                                     'write_bytes_disk': 786906.2858626033},
                                         'name': 'foo-cluster-default-worker-group-worker-11a4cef58d454b35aa35664380679798',
                                         'nanny': 'tcp://10.244.0.24:41277',
                                         'nthreads': 80,
                                         'resources': {},
                                         'services': {'dashboard': 43005},
                                         'type': 'Worker'},
             'tcp://10.244.0.25:40571': {'host': '10.244.0.25',
                                         'id': 'foo-cluster-default-worker-group-worker-f264282455a64f55a50b02c251487c87',
                                         'last_seen': 1647652673.778796,
                                         'local_directory': '/dask-worker-space/worker-1mvdgpz3',
                                         'memory_limit': 1082106646528,
                                         'metrics': {'bandwidth': {'total': 100000000,
                                                                   'types': {},
                                                                   'workers': {}},
                                                     'cpu': 2.0,
                                                     'executing': 0,
                                                     'in_flight': 0,
                                                     'in_memory': 0,
                                                     'memory': 120475648,
                                                     'num_fds': 25,
                                                     'read_bytes': 285.99359049856497,
                                                     'read_bytes_disk': 0.0,
                                                     'ready': 0,
                                                     'spilled_nbytes': {'disk': 0,
                                                                        'memory': 0},
                                                     'time': 1647652673.777244,
                                                     'write_bytes': 993.9777236208868,
                                                     'write_bytes_disk': 516084.4338529629},
                                         'name': 'foo-cluster-default-worker-group-worker-f264282455a64f55a50b02c251487c87',
                                         'nanny': 'tcp://10.244.0.25:41547',
                                         'nthreads': 80,
                                         'resources': {},
                                         'services': {'dashboard': 40213},
                                         'type': 'Worker'}}}
```

dask_kubernetes/operator/core.py
jacobtomlinson: We should just omit the `-n` flag from the command if this is set to `None`.

dask_kubernetes/operator/core.py
```diff
@@ -12,33 +12,98 @@
...
+        If the chart uses ClusterIP type services, forward the
+        ports locally. If you are running it locally it should
+        be the port you are forwarding to ``<port>``.
+    **kwargs: dict
+        Additional keyword arguments to pass to LocalCluster
+
+    Examples
+    --------
+    >>> from dask_kubernetes import KubeCluster2
+    >>> cluster = KubeCluster(name="foo")
```
jacobtomlinson, line 62: `KubeCluster2`?

dask_kubernetes/operator/core.py
```diff
@@ -47,102 +112,108 @@ def __init__(
...
+        kubernetes.config.load_kube_config()
+        self.core_api = kubernetes.client.CoreV1Api()
         data = build_cluster_spec(
-            self.name, self.image, self.replicas, self.resources, self.env
+            self.name, self.image, self.n_workers, self.resources, self.env
         )
-        with open("data.json", "w") as jfile:
-            json.dump(data, jfile)
+        temp_file = tempfile.NamedTemporaryFile()
+        config_path = f"{temp_file.name}.json"
```
jacobtomlinson, line 122: Why not just use `temp_file`? The `.json` extension doesn't really matter.

dask_kubernetes/tests/test_async.py
```diff
@@ -405,7 +405,7 @@ async def test_reject_evicted_workers(cluster):
     await cluster.core_api.create_namespaced_pod_eviction(
         (await worker.describe_pod()).metadata.name,
         (await worker.describe_pod()).metadata.namespace,
-        kubernetes.client.V1beta1Eviction(
+        kubernetes.client.V1Eviction(
```
Matt711, line 408: `V1beta1Eviction` was replaced in `kubernetes-asyncio` recently

jacobtomlinson (approved):
So happy to see green checks here!

I've had a play around locally and haven't run into any huge blockers. There are definitely things in here that need replacing and unpacking, especially the use of subprocess and nuking the finalizers. But let's do that in a follow up.

PR #415, opened 2022-03-09
Add docker image and manifest for deployment

jacobtomlinson:
Adds some initial packaging for the operator.

First the operator gets wrapped in a container image:

```
$ docker build -t ghcr.io/dask/dask-kubernetes-operator:latest -f dask_kubernetes/operator/deployment/Dockerfile .
```

_For local dev you'll need to push this to your kubernetes cluster, for me on `kind` it meant running `kind load docker-image --name pytest-kind ghcr.io/dask/dask-kubernetes-operator:latest`_

Then to install the operator on Kubernetes you need to apply the example manifest which creates a service account, cluster role, binding and deployment.

```
$ kubectl apply -f dask_kubernetes/operator/deployment/manifest.yaml
```

Now the operator is running and you can create `daskcluster` resources (assuming you have the CRDs installed).

### DIFF
--- a/dask_kubernetes/operator/deployment/Dockerfile
+++ b/dask_kubernetes/operator/deployment/Dockerfile
@@ -11,4 +11,4 @@ WORKDIR /src/dask_kubernetes
 RUN pip install .
 
 # Start operator
-CMD kopf run -m dask_kubernetes.operator.daskcluster --verbose --all-namespaces
+CMD kopf run -m dask_kubernetes.operator --verbose --all-namespaces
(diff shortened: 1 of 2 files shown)


dask_kubernetes/operator/deployment/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+# Copy source
+COPY . /src/dask_kubernetes
+WORKDIR /src/dask_kubernetes
+
+# Install dependencies
+RUN pip install .
+
+# Start operator
+CMD kopf run -m dask_kubernetes.operator.daskcluster --verbose --all-namespaces
```
jacobtomlinson, line 14: We should possibly tweak the way the imports work so that this could become 
```suggestion
CMD kopf run -m dask_kubernetes.operator--verbose --all-namespaces
```

PR #418, opened 2022-03-15
Support HPA style autoscaling

Matt711:
As a part of https://github.com/dask/dask-kubernetes/pull/392, the Dask Operator needs to support HPA-style autoscaling. This PR is concerned with adding this autoscaling to the Operator. Autoscaling is supported through a [timer](https://kopf.readthedocs.io/en/latest/timers/).


Matt711:
How should autoscaling work with worker groups?

dask_kubernetes/operator/tests/resources/simplecluster.yaml
```diff
@@ -18,4 +18,5 @@ spec:
     # serviceAccountName: null
   replicas: 3
   resources: {}
-  env: {}
\ No newline at end of file
+  env: {}
+  adaptive: "false"
```
Matt711, line 22: Kopf timer handler keeps checking `adaptive`. Once `adaptive` is set true, the operator ensures that the number of workers is between `minimum` and `maximum` and scales up and down depending on the workload. Then we can set `adaptive` to false when manual scaling is used and/or we provide a method for the user to stop adaptive mode.

jacobtomlinson: This is one approach. However it does means we would have `replicas`, `minimum`, `maximum` and `adaptive` to control scaling and we effectively would have to modes, adaptive and not adaptive.

I wonder if we could introduce `minimum` and `maximum` as a replacement and effectively just have an always-on adaptive mode. And somehow configure things so that if you scale to a fixed number then `minimum` and `maximum` are just set to the same value. This also has the benefit of failed workers being restarted automatically, something that doesn't happen with static scaling today.

jacobtomlinson:
I think something has gone wrong with this PR. I merged in the changed from `dask-operator` and it looks like everything here is gone. I suspect you may have force pushed the contents of #413 here by accident. Hopefully you still have the correct branch locally that you can push here.

PR #398, opened 2022-02-02
Pass worker resources to pod args

jhamman:
This adds a new keyword argument to `make_pod_spec`, facilitating passing worker resources through to the `dask-worker --resources` option.

This avoids the pattern of manipulating the pod spec after creating.

Before:

```python
pod_spec = make_pod_spec(...)
pod_spec.spec.containers[0].args.extend(["--resources", "FOO=1"])
```

Now:
```python
pod_spec = make_pod_spec(..., resources="FOO=1")
```

Issue #419, opened 2022-03-18
kubernetes_asyncio major version release breaks: service account not used anymore ("system/anonymous" not allowed to create "pods")


mkarbo:
First up: **Apologies**, I don't have a good example for reproducing the issue, as I use prefect and not dask directly.

*TL;DR*: New release of kubernetes-asyncio (22.6) broke something in dask-kubernetes (or in prefect, I don't know), pinning version fixed my workflows. Hope this helps someone.

---

Yesterday, my dask pipelines all worked perfectly in AWS EKS. Today, after re-building images, I consistently got the following:

```python
...
  File "/pipelines/flow_test/.venv/lib/python3.9/site-packages/dask_kubernetes/core.py", line 474, in __init__
    super().__init__(**self.kwargs)
  File "/pipelines/flow_test/.venv/lib/python3.9/site-packages/distributed/deploy/spec.py", line 260, in __init__
    self.sync(self._start)
  File "/pipelines/flow_test/.venv/lib/python3.9/site-packages/distributed/utils.py", line 311, in sync
    return sync(
  File "/pipelines/flow_test/.venv/lib/python3.9/site-packages/distributed/utils.py", line 378, in sync
    raise exc.with_traceback(tb)
  File "/pipelines/flow_test/.venv/lib/python3.9/site-packages/distributed/utils.py", line 351, in f
    result = yield future
  File "/pipelines/flow_test/.venv/lib/python3.9/site-packages/tornado/gen.py", line 762, in run
    value = future.result()
  File "/pipelines/flow_test/.venv/lib/python3.9/site-packages/dask_kubernetes/core.py", line 603, in _start
    await super()._start()
  File "/pipelines/flow_test/.venv/lib/python3.9/site-packages/distributed/deploy/spec.py", line 293, in _start
    self.scheduler = await self.scheduler
  File "/pipelines/flow_test/.venv/lib/python3.9/site-packages/distributed/deploy/spec.py", line 59, in _
    await self.start()
  File "/pipelines/flow_test/.venv/lib/python3.9/site-packages/dask_kubernetes/core.py", line 182, in start
    await super().start(**kwargs)
  File "/pipelines/flow_test/.venv/lib/python3.9/site-packages/dask_kubernetes/core.py", line 87, in start
    raise e
  File "/pipelines/flow_test/.venv/lib/python3.9/site-packages/dask_kubernetes/core.py", line 77, in start
    self._pod = await self.core_api.create_namespaced_pod(
[...]

PR #421, opened 2022-03-23
Add gpu_limit to make_pod_spec

ddelange:
- Add support for scheduling GPUs ([k8s docs](https://v1-22.docs.kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/)) ref https://github.com/dask/dask-kubernetes/pull/398#issuecomment-1076297582
- Correct `memory_limit` docstring: fractions (float) and 'GB' (str, internally converted to 'GiB') are [supported](https://github.com/dask/distributed/blob/2d3fddc14d1e06fd06b9f0f4c3256f254f3c670e/distributed/worker_memory.py#L344) by `--memory-limit` ([CLI docs](https://docs.dask.org/en/latest/deploying-cli.html)) but not by k8s' Pod resources spec ([k8s docs](https://v1-22.docs.kubernetes.io/docs/concepts/configuration/manage-resources-containers/)). It took me some time to find this out also ref https://github.com/dask/dask/issues/8224
- Sync docstring arg order with the function signature arg order
- People looking to schedule on AMD gpu's (vast minority) can still achieve this by passing `amd.com/gpu` along with `memory` and `cpu` requests/limits to `extra_pod_config` instead of using the separate request/limit args.

### DIFF
--- a/dask_kubernetes/objects.py
+++ b/dask_kubernetes/objects.py
@@ -122,7 +122,6 @@ def make_pod_spec(
     cpu_limit=None,
     cpu_request=None,
     gpu_limit=None,
-    gpu_request=None,
     annotations={},
 ):
     """
--- a/dask_kubernetes/objects.py
+++ b/dask_kubernetes/objects.py
@@ -163,8 +163,6 @@ def make_pod_spec(
         CPU resource requests (applied to ``spec.containers[].resources.requests.cpu``).
     gpu_limit : int
         GPU resource limits (applied to ``spec.containers[].resources.limits."nvidia.com/gpu"``).
-    gpu_request : int
-        GPU resource requests (applied to ``spec.containers[].resources.requests."nvidia.com/gpu"``).
     annotations : dict
         Dict of annotations passed to ``V1ObjectMeta``
 
@@ -207,8 +205,6 @@ def make_pod_spec(
 
     if cpu_request:
         resources.requests["cpu"] = cpu_request
-    if gpu_request:
-        resources.requests["nvidia.com/gpu"] = gpu_request
     if memory_request:
         resources.requests["memory"] = memory_request
 
--- a/dask_kubernetes/objects.py
+++ b/dask_kubernetes/objects.py
@@ -116,11 +116,13 @@ def make_pod_spec(
     env={},
     extra_container_config={},
     extra_pod_config={},
-    memory_limit=None,
     resources=None,
+    memory_limit=None,
     memory_request=None,
     cpu_limit=None,
     cpu_request=None,
+    gpu_limit=None,
+    gpu_request=None,
     annotations={},
 ):
     """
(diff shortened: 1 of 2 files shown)


jacobtomlinson (commented):
This looks great! With GPUs on Kubernetes the limit and request must always be set to the same value, I wonder if we should not expose two different options.

As you say AMD GPUs and also be configured via extra pod config, along with NVIDIA MIG.

ddelange:
Ah interesting, didn't know. Agreed :) Is it similar behaviour to memory, i.e. make a pod with memory limits and k8s will automatically set memory requests to the same value? In that case could leave only `gpu_limit` kwarg?

dask_kubernetes/objects.py
```diff
@@ -140,21 +142,29 @@ def make_pod_spec(
...
-        CPU resource limits (applied to ``spec.containers[].resources.limits.cpu``)
-    cpu_requests : float or str
-        CPU resource requests (applied to ``spec.containers[].resources.requests.cpu``)
+        CPU resource limits (applied to ``spec.containers[].resources.limits.cpu``).
+    cpu_request : float or str
+        CPU resource requests (applied to ``spec.containers[].resources.requests.cpu``).
+    gpu_limit : int
+        GPU resource limits (applied to ``spec.containers[].resources.limits."nvidia.com/gpu"``).
+    gpu_request : int
+        GPU resource requests (applied to ``spec.containers[].resources.requests."nvidia.com/gpu"``).
```
ddelange, line 167:
```suggestion
```
PR Suggestion

dask_kubernetes/objects.py
```diff
@@ -197,11 +207,15 @@ def make_pod_spec(
 
     if cpu_request:
         resources.requests["cpu"] = cpu_request
+    if gpu_request:
+        resources.requests["nvidia.com/gpu"] = gpu_request
```
ddelange, line 211:
```suggestion
```
PR Suggestion

ddelange:
> Some CI errors maybe due to new major release of kubernetes_asyncio ref #419 ?
> 
> ```
> AttributeError: module 'kubernetes_asyncio.client' has no attribute 'V1beta1Eviction'
> ```

Pushed a fix, V1beta1Eviction was [removed](https://github.com/tomplus/kubernetes_asyncio/blame/34c67ac27576f8b57b56b4c7432e0e56bd925b1f/kubernetes_asyncio/client/__init__.py#L220) in v22

dask_kubernetes/objects.py
```diff
@@ -116,11 +116,13 @@ def make_pod_spec(
...
     extra_container_config={},
     extra_pod_config={},
-    memory_limit=None,
     resources=None,
+    memory_limit=None,
     memory_request=None,
     cpu_limit=None,
     cpu_request=None,
+    gpu_limit=None,
+    gpu_request=None,
```
ddelange, line 125:
```suggestion
```
Leftover

jacobtomlinson (approved):
Woo this looks great! Many thanks.

PR #422, opened 2022-03-24
Fix missing container name in `Pod.logs()`

bstadlbauer:
`Pod.logs()` will fail when there are multiple containers within one pod. 

It will fail with e.g: 
```
kubernetes_asyncio.client.exceptions.ApiException: (400)
Reason: Bad Request
HTTP response headers: <CIMultiDictProxy('Audit-Id': '6665d5a7-0ace-40f6-8986-96f9713d1275', 'Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'Date': 'Thu, 24 Mar 2022 11:30:43 GMT', 'Content-Length': '249')>
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"a container name must be specified for pod dask-bstadlbauer-5abb6a0e-2s5xvr, choose one of: [vault-agent-init dask-worker vault-agent]","reason":"BadRequest","code":400}
```

This is an issue, as some secret injection frameworks use a sidecar container within the same pod to inject secrets.
Whilst fixing this, I've also tried to extract all strings referencing "dask-worker" container name into `constants.DASK_CONTAINER_NAME`. Let me know if you want to keep that as it was before and I'll revert it. 

Side-note: it seems like both, the scheduler, as well as the workers are named "dask-worker" in k8s. As I had troubles starting tests on my local machine, I have not added a test to replicate the issue.

### DIFF
--- a/dask_kubernetes/constants.py
+++ b/dask_kubernetes/constants.py
@@ -1,0 +1,1 @@
+DASK_CONTAINER_NAME = "dask-worker"
--- a/dask_kubernetes/constants.py
+++ b/dask_kubernetes/constants.py
@@ -1,1 +1,1 @@
-DASK_CONTAINER_NAME = "dask-worker"
+KUBECLUSTER_WORKER_CONTAINER_NAME = "dask-worker"
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -16,7 +16,7 @@ from distributed.utils import Log, Logs
 import kubernetes_asyncio as kubernetes
 from kubernetes_asyncio.client.rest import ApiException
 
-from .constants import DASK_CONTAINER_NAME
+from .constants import KUBECLUSTER_WORKER_CONTAINER_NAME
 from .objects import (
     make_pod_from_dict,
     make_service_from_dict,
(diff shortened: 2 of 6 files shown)


dask_kubernetes/constants.py
```diff
@@ -0,0 +1 @@
+DASK_CONTAINER_NAME = "dask-worker"
```
jacobtomlinson, line 1: Given that this is specific only to `KubeCluster` and not `HelmCluster` or the new operator could we name this as something more specific.

Something like this

```suggestion
KUBECLUSTER_WORKER_CONTAINER_NAME = "dask-worker"
```

jacobtomlinson (commented):
Thank you so much for working on this. This seems like a great improvement, and removes a magic string!

I've left one small comment.

dask_kubernetes/constants.py
bstadlbauer: Done

bstadlbauer:
@jacobtomlinson Renamed `DASK_CONTAINER_NAME` to`KUBECLUSTER_WORKER_CONTAINER_NAME` - thanks for getting back this quickly!

jacobtomlinson (approved):
Awesome!

PR #438, opened 2022-04-26
Remove kubectl dependency from operator

jacobtomlinson:
- Moved the auth stuff into a startup event
- Removed use of `kubectl` in the operator code
- Removed `kubectl` from the docker image

### DIFF
--- a/dask_kubernetes/operator/deployment/Dockerfile
+++ b/dask_kubernetes/operator/deployment/Dockerfile
@@ -10,13 +10,5 @@ WORKDIR /src/dask_kubernetes
 # Install dependencies
 RUN pip install .
 
-# We don't use TARGETARCH so as to support non-buildkit builds
-RUN MAGICARCH=$(dpkg --print-architecture) && \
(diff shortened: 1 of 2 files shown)


dask_kubernetes/operator/deployment/Dockerfile
```diff
@@ -10,13 +10,5 @@ WORKDIR /src/dask_kubernetes
...
 RUN pip install .
 
-# We don't use TARGETARCH so as to support non-buildkit builds
-RUN MAGICARCH=$(dpkg --print-architecture) && \
-    curl -LO "https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/${MAGICARCH}/kubectl" && \
-    mkdir -p /usr/local/bin && \
-    mv ./kubectl /usr/local/bin/kubectl && \
-    chmod +x /usr/local/bin/kubectl && \
-    kubectl version --client
-
```
Matt711, line 20: The cluster manager still uses `kubectl` in a few of the methods.

jacobtomlinson: The cluster manager will not be used within the operator container though right? Just on the user's machine.

Matt711: Yes, that's right.

PR #441, opened 2022-04-26
Operator docs

Matt711:
Raise not implemented error for adapt. And add missing methods from `KubeCluster` to the methods section


doc/source/operator.rst
```diff
@@ -246,7 +246,10 @@ API
 .. autosummary::
    KubeCluster
    KubeCluster.scale
+   KubeCluster.adapt
```
jacobtomlinson, line 249: If this is not implemented shouldn't we exclude it from here?

PR #445, opened 2022-04-28
Remove kubectl from dask_kubernetes.experimental.KubeCluster

jacobtomlinson:
Start removing all uses of `kubectl` from the new `dask_kubernetes.experimental.KubeCluster` class.

Only one left is scale because I can't figure out what to put in the body of [`patch_namespaced_custom_object_scale`](https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CustomObjectsApi.md#patch_namespaced_custom_object_scale).

### DIFF
--- a/dask_kubernetes/experimental/kubecluster.py
+++ b/dask_kubernetes/experimental/kubecluster.py
@@ -1,4 +1,3 @@
-import subprocess
 import kubernetes_asyncio as kubernetes
 
 from distributed.core import rpc
@@ -282,18 +281,19 @@ class KubeCluster(Cluster):
         return self.sync(self._scale, n, worker_group)
 
     async def _scale(self, n, worker_group="default"):
-        subprocess.check_output(
-            [
-                "kubectl",
-                "scale",
-                f"--replicas={n}",
-                "daskworkergroup",
-                f"{self.name}-cluster-{worker_group}-worker-group",
-                "-n",
-                self.namespace,
-            ],
-            encoding="utf-8",
-        )
+        async with kubernetes.client.api_client.ApiClient() as api_client:
+            custom_objects_api = kubernetes.client.CustomObjectsApi(api_client)
+            custom_objects_api.api_client.set_default_header(
+                "content-type", "application/merge-patch+json"
+            )
+            await custom_objects_api.patch_namespaced_custom_object_scale(
+                group="kubernetes.dask.org",
+                version="v1",
+                plural="daskworkergroups",
+                namespace=self.namespace,
+                name=f"{self.name}-cluster-{worker_group}-worker-group",
+                body={"replicas": n},
+            )
 
     def adapt(self, *args, **kwargs):
         """Turn on adaptivity"""
--- a/dask_kubernetes/experimental/kubecluster.py
+++ b/dask_kubernetes/experimental/kubecluster.py
@@ -13,6 +13,7 @@ from dask_kubernetes.auth import ClusterAuth
 from dask_kubernetes.operator.daskcluster import (
     build_cluster_spec,
     build_worker_group_spec,
+    wait_for_service,
 )
 
 from dask_kubernetes.utils import (
@@ -95,8 +96,6 @@ class KubeCluster(Cluster):
         self.name = name
         # TODO: Set namespace to None and get default namespace from user's context
         self.namespace = namespace
-        self.core_api = None
-        self.custom_api = None
         self.image = image
         self.n_workers = n_workers
         self.resources = resources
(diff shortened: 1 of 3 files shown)


Matt711:
Is the [patch_namespaced_custom_object_scale](https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CustomObjectsApi.md#patch_namespaced_custom_object_scale) covered in the  `kubernetes-asyncio` API?

jacobtomlinson:
It should be (everything should be though 😂). I can call it successfully but I'm getting errors about unexpected content type.

This is what I tried:

```diff
diff --git a/dask_kubernetes/experimental/kubecluster.py b/dask_kubernetes/experimental/kubecluster.py
index 53dde27..bd6ec8c 100644
--- a/dask_kubernetes/experimental/kubecluster.py
+++ b/dask_kubernetes/experimental/kubecluster.py
@@ -1,4 +1,3 @@
-import subprocess
 import kubernetes_asyncio as kubernetes
 
 from distributed.core import rpc
@@ -282,18 +281,16 @@ class KubeCluster(Cluster):
         return self.sync(self._scale, n, worker_group)
 
     async def _scale(self, n, worker_group="default"):
-        subprocess.check_output(
-            [
-                "kubectl",
-                "scale",
-                f"--replicas={n}",
-                "daskworkergroup",
-                f"{self.name}-cluster-{worker_group}-worker-group",
-                "-n",
-                self.namespace,
-            ],
-            encoding="utf-8",
-        )
+        async with kubernetes.client.api_client.ApiClient() as api_client:
+            custom_objects_api = kubernetes.client.CustomObjectsApi(api_client)
+            await custom_objects_api.patch_namespaced_custom_object_scale(
+                group="kubernetes.dask.org",
+                version="v1",
+                plural="daskworkergroups",
+                namespace=self.namespace,
+                name=f"{self.name}-cluster-{worker_group}-worker-group",
+                body={"replicas": n},
+            )
 
     def adapt(self, *args, **kwargs):
         """Turn on adaptivity"""
```

But it results in this error.

```python-traceback
E               kubernetes_asyncio.client.exceptions.ApiException: (415)
E               Reason: Unsupported Media Type
E               HTTP response headers: <CIMultiDictProxy('Audit-Id': 'a09fea8d-9f0a-454a-aba3-b99cfe1154c6', 'Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'Date': 'Thu, 28 Apr 2022 13:19:44 GMT', 'Content-Length': '293')>
E               HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"the body of the request was in an unknown format - accepted media types include: application/json-patch+json, application/merge-patch+json, application/apply-patch+yaml","reason":"UnsupportedMediaType","code":415}
```

Matt711:
The `body` argument is not in a suitable format? Convert it to JSON?

jacobtomlinson:
Yeah I also tried `body=json.dumps({"replicas": n}),` but got the same error.

jacobtomlinson:
Same error with `{"spec": {"replicas": n}}` and `json.dumps({"spec": {"replicas": n}})`.

Matt711:
The relevant issue in `kubernetes-asyncio`. I think the problem is with the headers like you said. Can you try to change the default headers like so?
`custom_objects_api.api_client.set_default_header('content-type', 'application/merge-patch+json')`
Or 
`custom_objects_api.api_client.set_default_header('Content-Type', 'application/merge-patch+json')`

jacobtomlinson:
Looks like the tests are passing but the CRDs fixture is failing on teardown with a timeout. I guess that's because some resource if failing to remove and so the CRDs can't be cleaned up. I'm seeing the same thing locally.

jacobtomlinson:
So I think the reason things were failing were due to this sequence of events:

```python
with kopf_runner:
    with KubeCluster(name="foo") as cluster:
        with Client(cluster) as client:
            assert client.submit(lambda x: x + 1, 10).result() == 11
```

The operator starts in the outer context manager, then a cluster is created, then a client is connected and a future is submitted. Once the future completes the client disconnects, the cluster closes and the operator stops.

The issue was that when the cluster closed it called `delete_namespaced_custom_object` which marks the object for deletion in the Kubernetes API and then returns immediately. The operator would then also close as there was nothing else happening inside the context manager, however the operator was required to actually delete the object, so this never happens and the tests eventually time out.

I've updated the cluster close method to also wait until the resource has actually been deleted before returning. This gives the operator breathing room to actually do what it needs before everything exits.

This seem ok for me locally, hopefully this will fix things in CI too.

PR #446, opened 2022-04-30
feat: pass CRD environment variables through to workers and scheduler for Kubernetes operator

samdyzon:
Hi there!

My team and I have been eagerly watching the progress of your work on the Kubernetes operator since you started, and we are really excited to start getting into it and testing it. Thank you so much for all your hard work on this!

During my testing this morning, I realised that the `env` property for the `DaskCluster` CRD is not being passed through to the workers and schedulers. I also realised that the type was set to an `object` which doesn't match the expected values you would use when attaching environment variables to a pod.

This PR aims to:

* Update the CRD templates to use an array type with all of the properties used for environment variables in pods: `value`, `valueFrom` and `secretKeyRef`/`configMapRef`.
* Update the operator code to pass through the environment variables into the `build_worker_pod_spec` and `build_scheduler_pod_spec` methods (with a default of an empty list `[]` if not provided)
* Update the `dask_kubernetes.experimental.KubeCluster` class to set the default env to an empty list **[1]**
* Update test fixtures, and add a check to see if the environment variables have passed through correctly.
* Update the documentation to reflect the env value type change.

**[1]** There is one point in this PR that may be an issue: The type of value passed into `dask_kubernetes.experimental.KubeCluster` now more accurately matches the expected values in Kubernetes, but it is a departure from the type required by `core.DaskCluster` and even the helper method `objects.make_pod_spec` which expect a `Dict[str, str]` type. This configuration in the core library prevents the ability to use kubernetes secrets in environment variables, which this PR aim's to enable for the CRD. This difference between the current core is a point of contention for my team, because we need to pass in kubernetes secrets into our worker environment variables but cannot do so with the current core module (we work around this by injecting the required environment variable directly into the pod spec after calling `make_pod_spec` but this PR and the operator will remove the need for this work around). I've updated the docstring of the `KubeCluster` class, but I'm not sure how to handle type hints of such a complicated type (see the CRD changes) in python, so I hope this will suffice for now.

Please let me know if there are any issues, I'm happy to help if I can.

Related to #256


samdyzon:
I'm having a hard time getting tests to pass - mostly looks like timeouts/networking issues that I don't think would be caused by my changes. Running the test suite locally has similar problems, but I have tested by successfully deploying everything to the test cluster so I'm not sure what more I can do at the moment.

Matt711:
> **[1]** There is one point in this PR that may be an issue: The type of value passed into `dask_kubernetes.experimental.KubeCluster` now more accurately matches the expected values in Kubernetes, but it is a departure from the type required by `core.DaskCluster` and even the helper method `objects.make_pod_spec` which expect a `Dict[str, str]` type. This configuration in the core library prevents the ability to use kubernetes secrets in environment variables, which this PR aim's to enable for the CRD. This difference between the current core is a point of contention for my team, because we need to pass in kubernetes secrets into our worker environment variables but cannot do so with the current core module (we work around this by injecting the required environment variable directly into the pod spec after calling `make_pod_spec` but this PR and the operator will remove the need for this work around). I've updated the docstring of the `KubeCluster` class, but I'm not sure how to handle type hints of such a complicated type (see the CRD changes) in python, so I hope this will suffice for now.

Sorry, `core.DaskCluster`? Do you mean `core.KubeCluster`? The goal of `dask_kubernetes.experimental.KubeCluster` is to eventually replace `core.KubeCluster` which is why the new `KubeCluster` (based on the operator) is in an `experimental` module right now. We plan on keeping the old `KubeCluster` for a while until it becomes easily replaceable. So, I don't think it's a problem that there is a difference between them. Even more, it's good that now `env` more accurately matches what Kubernetes expects.

jacobtomlinson (commented):
This is awesome, thanks for contributing here and glad the operator will be valuable to you.

I'm generally happy with these changes. I'm keen to note that the new `dask_kubernetes.experimental.KubeCluster` doesn't have to match the API for the existing `dask_kubernetes.KubeCluster`. We are free to deviate however we feel makes sense. It is still undecided if it will replace the existing at some point or just sit along side it, but if it does replace we will just make one breaking change at some point. So don't worry about making large changes here.

It's interesting that you note the existing cluster manager stops you from using secrets in your env vars. Sorry about that! These are things I'm definitely keen to fix in the new implementation.

I want to try and balance user experience with functionality. For folks that just want plain old env vars it sucks to have to do things like `[{"name": "foo", "value": "bar"}]` when dictionaries literally mean this and you could just pass `{"foo": "bar"}`. This is one of my particular grumbles about using `boto3`. It's common to see users write helper functions to convert dicts to/from the list of dict form and they shouldn't have to do that.

However as you rightly point out k8s supports more complex options here, like using secrets. In which case you would need to pass something like `[{"name": "foo",  "valueFrom": {"secretKeyRef": {"name": "mysecret", "key": "foo"}}}]`.

I would be really keen to support both dicts and lists of dicts, and we can just type check and convert dicts to lists of dicts automatically. What do you think?

samdyzon:
Good morning gents, thanks for the feedback on this. I'm glad to hear that you guys are happy with this PR. To reply to specific comments:

> I'm going to take a look at your changes locally. The tests are passing for me.

Tests are working in the CI runner, I find that the tests are somewhat flaky when running locally and end up timing out fairly often, but this may be due to my machine. I was able to get them to run fully in CI and found a few places where my test logic was actually failing, so they should be sweet now.

> It's interesting that you note the existing cluster manager stops you from using secrets in your env vars. Sorry about that! These are things I'm definitely keen to fix in the new implementation.

Thats all good, our work around is pretty straight forward - we just update the output dictionary from `make_pod_spec` by appending the appropriate dictionaries to the `env` list. Its not pretty but it works. 

> I would be really keen to support both dicts and lists of dicts, and we can just type check and convert dicts to lists of dicts automatically. What do you think?

Totally makes sense to me, and I reckon it would be easy enough to implement. I left that part out mostly because I didnt want to make any assumptions about changes to the API - the only change that had to be made as part of this PR is to change the default input type to be a list so that line 119 has the correct type and wont throw a kubernetes API error when trying to create the dask cluster resource type.

After having written #447 and thinking about it more, I think that this PR is probably not worth merging since it would most likely be completely rewritten. I'd recommend holding off on this PR for now and I'll respond to comments in the issue.

jacobtomlinson:
> I left that part out mostly because I didnt want to make any assumptions about changes to the API

Changes to the API are totally fair game for the time being. The main reason we merged #392 and tagged it in a release was to reduce the friction for early adopters to try it out, but that in no way means it is stable or closed to modification yet.

> I think that this PR is probably not worth merging since it would most likely be completely rewritten

That's fair, however I'm very much in a "yes and" mindset with this. This PR improves things. We gave some feedback on how it could be even better, but I am also very happy to defer those changes to a future PR. If #447 removes the need for that follow-up PR because everything here gets rewritten then that's fine too. But this PR is in a good state to merge so let's do it.

PR #450, opened 2022-05-04
Fix worker to scheduler pairing

### LINKED ISSUE #449: Workers join wrong scheduler when creating multiple clusters
**What happened**:

When creating multiple clusters all the workers join the scheduler of the first cluster.

**What you expected to happen**:

Each worker group should join their own scheduler.

```python

# add to dask_kubernetes/operator/tests/test_operator.py

def test_multiple_clusters_simultaneously(kopf_runner):
    with kopf_runner:
        with KubeCluster(name="bar") as cluster1, KubeCluster(name="baz") as cluster2:
            with Client(cluster1) as client1, Client(cluster2) as client2:
                assert client1.submit(lambda x: x + 1, 10).result() == 11
                assert client2.submit(lambda x: x + 1, 10).result() == 11
```

The second assertion hangs and times out because no workers ever join `cluster2` because they join `cluster1` instead and the future is never computed.

The following lines need updating to choose the right scheduler.

https://github.com/dask/dask-kubernetes/blob/b042fbf345e8d375aa813c9ade86a2d852fe1296/dask_kubernetes/operator/operator.py#L264

https://github.com/dask/dask-kubernetes/blob/b042fbf345e8d375aa813c9ade86a2d852fe1296/dask_kubernetes/operator/operator.py#L288

jacobtomlinson:
Closes #449 

- Add `cluster` property to worker groups to make it easier for them to find their own scheduler. 
- Small hygiene tweaks

### DIFF
--- a/dask_kubernetes/experimental/kubecluster.py
+++ b/dask_kubernetes/experimental/kubecluster.py
@@ -203,7 +203,12 @@ class KubeCluster(Cluster):
 
     async def _add_worker_group(self, name, n=3):
         data = build_worker_group_spec(
-            f"{self.name}-cluster-{name}", self.image, n, self.resources, self.env
+            f"{self.name}-cluster-{name}",
+            f"{self.name}-cluster",
+            self.image,
+            n,
+            self.resources,
+            self.env,
         )
         async with kubernetes.client.api_client.ApiClient() as api_client:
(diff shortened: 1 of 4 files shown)


Matt711 (approved):
LGTM

PR #454, opened 2022-05-05
Fix worker group adoptions

jacobtomlinson:
When a workergroup is created it now patches itself with the ownership pointing to the cluster object in the create hook. This means that additional worker groups will also be correctly owned by the cluster. 

I also removed the pod creation logic from the create hook and just call the update hook instead to reduce duplication.


Matt711 (approved):
LGTM

PR #458, opened 2022-05-05
fix: environ may duplicate keys for str format

### LINKED ISSUE #453: TypeError: str.format() got multiple values for keyword argument 'user'
When an environment variable `user` is set, the example code from https://kubernetes.dask.org/en/latest/kubecluster.html fails with: 

```
  File "main.py", line 12, in <module>
    cluster = KubeCluster(pod_spec, name="trackingmigration")
  File "/lib/python3.10/site-packages/dask_kubernetes/core.py", line 496, in __init__
  File "/lib/python3.10/site-packages/distributed/deploy/spec.py", line 260, in __init__
    self.sync(self._start)
  File "/lib/python3.10/site-packages/distributed/utils.py", line 318, in sync
    return sync(
  File "/lib/python3.10/site-packages/distributed/utils.py", line 385, in sync
    raise exc.with_traceback(tb)
  File "/lib/python3.10/site-packages/distributed/utils.py", line 358, in f
    result = yield future
  File "/lib/python3.10/site-packages/tornado/gen.py", line 762, in run
    value = future.result()
  File "/lib/python3.10/site-packages/dask_kubernetes/core.py", line 572, in _start
    self._generate_name = self._generate_name.format(
TypeError: str.format() got multiple values for keyword argument 'user'
ERROR:asyncio:Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x119095f90>
ERROR:asyncio:Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x1190960b0>
```

This is because https://github.com/dask/dask-kubernetes/blob/main/dask_kubernetes/core.py#L573 doesn't check that it's expanded the same value >1x.

mirosval:
In an interesting edge-case, when you have `user` or `uuid` env variables defines str.format dies because of dupplicate keys. The fix is to filter the environ dictionary and remove those 2 keys

fixes #453 

I was unable to test this locally properly because tests were failing for me when I downloaded the project.

### DIFF
--- a/dask_kubernetes/core.py
+++ b/dask_kubernetes/core.py
@@ -569,8 +569,9 @@ class KubeCluster(SpecCluster):
         if self.namespace is None:
             self.namespace = namespace_default()
 
+        environ = {k:v for k, v in os.environ.items() if k not in ['user', 'uuid']}
         self._generate_name = self._generate_name.format(
-            user=getpass.getuser(), uuid=str(uuid.uuid4())[:10], **os.environ
+            user=getpass.getuser(), uuid=str(uuid.uuid4())[:10], **environ
         )
         self._generate_name = escape(self._generate_name)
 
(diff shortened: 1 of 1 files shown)


jacobtomlinson (approved):
Awesome many thanks.

PR #459, opened 2022-05-06
Only connect the RPC when we need it

jacobtomlinson:
Given that we only call the RPC once right now and we are shifting to the HTTP API soon I decided to simplify things a little and remove the RPC tracking singleton.

This PR just creates the RPC when we need it.


Matt711 (approved):
LGTM

PR #467, opened 2022-05-10
Put experimental tests with code instead of in operator

jacobtomlinson:
Added a quick try except to catch when `dask_kubernetes.experimental.KubeCluster` is used without the CRDs being installed. Then wanted to add a test for this and ended up refactoring the tests for the new `KubeCluster` to live alongside it instead of in the `operator` directory.


dask_kubernetes/conftest.py
```diff
@@ -1,15 +1,26 @@
...
 
+DIR = pathlib.Path(__file__).parent.absolute()
+
 check_dependency("helm")
 check_dependency("kubectl")
 check_dependency("docker")
 
 
+@pytest.fixture()
+async def kopf_runner(k8s_cluster):
```
graingert, line 20: why is this an `async def` fixture -

https://github.com/nolar/kopf/blob/ab53ace82e62a6fa709bf5a580007eac1273ac34/kopf/_kits/runner.py#L123

kopf_runner creates its own eventloop in its own thread

jacobtomlinson: This is likely a typo :)

PR #469, opened 2022-05-10
Fix docker image name in documentation

Matt711:
Quick fix for a docker image name in the documentation, so the `highmem-worker-group` workers run correctly.

### DIFF
--- a/doc/source/operator.rst
+++ b/doc/source/operator.rst
@@ -217,7 +217,7 @@ Let's create an example called ``highmemworkers.yaml`` with the following config
    spec:
       cluster: simple-cluster
       imagePullSecrets: null
-      image: "dask-kubernetes:dev"
+      image: "daskdev/dask:latest"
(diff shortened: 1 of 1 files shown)


doc/source/operator.rst
```diff
@@ -217,7 +217,7 @@ Let's create an example called ``highmemworkers.yaml`` with the following config
    spec:
       cluster: simple-cluster
       imagePullSecrets: null
-      image: "dask-kubernetes:dev"
+      image: "daskdev/dask:latest"
```
jacobtomlinson, line 220:
```suggestion
      image: "ghcr.io/dask/dask:latest"
```

jacobtomlinson (approved):
Good catch! But let's use the ghcr version as that is the main one we want folks using.

jacobtomlinson:
(Also yay for fast CI for docs changes!)

PR #478, opened 2022-05-12
Publish Helm Chart on release with chartpress

jacobtomlinson:
Removed the image config from the `chartpress.yaml` as we already build this image. Then added a workflow step to the release to publish the chart.


jacobtomlinson:
I tested this manually and released `2022.5.0-alpha.1` and things seem to work. I'm going to merge this and tag a release and see what happens.

Issue #479, opened 2022-05-12
Helm chart doesn't include any roles


jacobtomlinson:
The helm chart for the operator that was added in #452 has a service account but no roles. This means the chart doesn't work out of the box. The chart has been published in the latest release (`2022.5.0`) and is available at https://helm.dask.org. But we don't currently document it.

@samdyzon do you have thoughts about how the operator installed via the chart will authenticate with the cluster? Should we add roles or did you have something else in mind?

It would also be good to document using the helm chart.

samdyzon:
Looks like I've forgotten to add the service account roles to the helm chart - I totally thought I did that, sorry! To fix this, I can put together a PR which by default creates a service account which has the same roles and role bindings as the manifest in `dask_kubernetes/operator/deployment/manifests/operator.yaml` - if a user decides to override the `values.yaml` file to not create a serviceAccount by default, they will be required to provide their own roles/rolebindings.

jacobtomlinson:
Closed in #480

PR #480, opened 2022-05-12
feat: add service account roles/rolebinding to operator helm chart

samdyzon:
Add the roles/rolebindings to the default service account in the operator helm chart. 

Also added a bit of documentation for using the helm chart. The biggest issue to point out here is the lack of update/delete support for CRDs in helm. Maybe we could also add some doco about default values? What do you reckon?

Related to #479

### DIFF
--- a/doc/source/operator.rst
+++ b/doc/source/operator.rst
@@ -41,6 +41,25 @@ This will create the appropriate roles, service accounts and a deployment for th
    kube-system   dask-kubernetes-operator-775b8bbbd5-zdrf7   1/1     Running   0          74s
 
 
+Installing the operator with Helm
+---------------------------------
+
+Along with a set of kubernetes manifests, the operator has a basic Helm chart which can be used to manage the installation of the operator.
+The chart is published in the ``dask/helm-charts`` repository, and can be installed via:
+
+.. code-block:: console
+
+    $ helm repo add dask https://helm.dask.org
+    $ helm repo update
+    $ helm install --version 2022.5.0 myrelease dask/dask-kubernetes-operator
+
+This will install the custom resource definitions, service account, roles, and the operator deployment.
+
+.. warning::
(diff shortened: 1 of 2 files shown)


doc/source/operator.rst
```diff
@@ -41,6 +41,25 @@ This will create the appropriate roles, service accounts and a deployment for th
...
+---------------------------------
+
+Along with a set of kubernetes manifests, the operator has a basic Helm chart which can be used to manage the installation of the operator.
+The chart is published in the ``dask/helm-charts`` repository, and can be installed via:
+
+.. code-block:: console
+
+    $ helm repo add dask https://helm.dask.org
+    $ helm repo update
+    $ helm install --version 2022.5.0 myrelease dask/dask-kubernetes-operator
```
jacobtomlinson, line 54:
```suggestion
    $ helm install myrelease dask/dask-kubernetes-operator
```

jacobtomlinson: I'm going to remove the version pin here to make the docs more evergreen.

doc/source/operator.rst
```diff
@@ -41,6 +41,25 @@ This will create the appropriate roles, service accounts and a deployment for th
    kube-system   dask-kubernetes-operator-775b8bbbd5-zdrf7   1/1     Running   0          74s
 
 
+Installing the operator with Helm
+---------------------------------
+
+Along with a set of kubernetes manifests, the operator has a basic Helm chart which can be used to manage the installation of the operator.
+The chart is published in the ``dask/helm-charts`` repository, and can be installed via:
```
jacobtomlinson, line 48:
```suggestion
The chart is published in the `Dask Helm repo <https://helm.dask.org>`_ repository, and can be installed via:
```

I don't think folks will care too much about the specific of how the helm deployment works. Rather they will care about where to install it from.

jacobtomlinson (approved):
Thanks for picking this up so quickly!

Issue #474, opened 2022-05-12
Make more of the DaskCluster and DaskWorkerGroup specs optional


jacobtomlinson:
In #452 we merged new CRDs which are far more verbose and utilize the full `pod` and `scheduler` specs from Kubernetes. This is awesome because it enables users to have almost infinite flexibility when configuring pods and services. However, this does also give users far more rope and I'm nervous about how easy it is to misconfigure a cluster.

I'm totally sold on the verbosity of these manifests, it is the Kubernetes was and it is good to be explicit. However, I wonder if we can make more things optional to reduce the chance of error (like many settings in `Pod` are optional).

For example in the docs our example cluster looks like this:

```yaml
apiVersion: kubernetes.dask.org/v1
kind: DaskCluster
metadata:
  name: simple-cluster
spec:
  worker:
    replicas: 2
    spec:
      containers:
      - name: worker
        image: "ghcr.io/dask/dask:latest"
        imagePullPolicy: "IfNotPresent"
        args:
          - dask-worker
          # Note the name of the cluster service, which adds "-service" to the end
          - tcp://simple-cluster-service.default.svc.cluster.local:8786
  scheduler:
    spec:
      containers:
      - name: scheduler
        image: "ghcr.io/dask/dask:latest"
        imagePullPolicy: "IfNotPresent"
        args:
          - dask-scheduler
        ports:
          - name: comm
            containerPort: 8786
            protocol: TCP
          - name: dashboard
            containerPort: 8787
            protocol: TCP
        readinessProbe:
          tcpSocket:
            port: comm
            initialDelaySeconds: 5
            periodSeconds: 10
        livenessProbe:
          tcpSocket:
            port: comm
            initialDelaySeconds: 15
            periodSeconds: 20
    service:
      type: NodePort
      selector:
        dask.org/cluster-name: simple-cluster
        dask.org/component: scheduler
      ports:
      - name: comm
        protocol: TCP
        port: 8786
        targetPort: "comm"
      - name: dashboard
        protocol: TCP
        port: 8787
        targetPort: "dashboard"
```

In the majority of cases the scheduler config is going to be identical except for the `dask.org/cluster-name` selector which just needs to match the `metadata.name` set at the top. Of the users that will want to configure this I expect the majority of those will only want to change the service, not the pod.

I think we could make the following sections optional and if they are omitted just set them to the values from the example above in the operator:

- `spec.worker.spec`
- `spec.scheduler`
- `spec.scheduler.service`
[...]

PR #488, opened 2022-05-16
Add Istio support

### LINKED ISSUE #482: Support Istio in the operator
In order to support KubeFlow we need to support Istio. 

The advice we were given by @stephan-erb-by and @philipp-sontag-by is to put a service in front of every worker and have all inter-pod communication go via services. This should then play nicely with Istio.

Thanks to the changes added in #452 it should be straightforward to extend the CRDs to also include a service for the workers. We would just need to ensure that the workers advertise the service address to the scheduler instead of its own pod IP, this can probably be done by injecting environment variables.

cc @samdyzon

jacobtomlinson:
Closes #482.

The PR aims to ensure the operator will play nicely with Istio. As discussed in #482 the way to achieve this is for the workers to have a service each and ensure all network connections within the Dask cluster are going via services rather than directly pod to pod.

To kick things off in this PR I've added a fixture that installs Istio into the kind cluster that gets created, but it is opt-in. To enable it you need to set the `TEST_ISTIO=true` env var. This means we can enable Istio in the operator CI but not the Helm or classic `KubeCluster` CI.

I also had to shorten the `uuid` that was being appended to the worker name because Istio was reusing that name and appending more and was hitting a character limit. We should possibly consider reducing the length of some of our naming schemes.

I've then added a service in front of the workers and tried to get the Dask cluster to only speak via those. Things are not 100% working with this though.

I'm also doing some environment variable injection as discussed in #474. The use case here is so that the worker can correctly set its name.

### DIFF
--- a/dask_kubernetes/conftest.py
+++ b/dask_kubernetes/conftest.py
@@ -36,6 +36,16 @@ def k8s_cluster(kind_cluster, docker_image):
     del os.environ["KUBECONFIG"]
 
 
+@pytest.fixture(scope="session", autouse=True)
+def install_istio(k8s_cluster):
+    if bool(os.environ.get("TEST_ISTIO", False)):
+        check_dependency("istioctl")
+        subprocess.check_output(["istioctl", "install", "--set", "profile=demo", "-y"])
+        k8s_cluster.kubectl(
+            "label", "namespace", "default", "istio-injection=enabled", "--overwrite"
+        )
+
+
 @pytest.fixture(scope="session")
 def ns(k8s_cluster):
     return "default"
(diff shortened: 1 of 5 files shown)


rrmistry:
Bringing the comment here from https://github.com/dask/dask-kubernetes/commit/674d8bb6fa6fa23daa752cf1879f70fa5a4fda98#commitcomment-133495956

We cannot use Dask in production because the Dask operator is injecting pod labels `"sidecar.istio.io/inject": "false",` which disables Istio side-car injection. With this label, the mesh communication breaks down and the Dask cluster does not behave as expected with other components within the Kubernetes cluster.

@jacobtomlinson can we collaborate offline to discuss options? We can connect on [Gitter](https://app.gitter.im/) or [via email](mailto:rohit.mistry@righteous.ai)

rrmistry:
> @rrmistry Do you think this is something you have the time/interest to help out with?

Sure. I can invest some time for this as Dask is planned to be a major investment for us going forward.

I will take a look at any CONTRIBUTING docs or guidelines for best practices.

Thanks!

PR #489, opened 2022-05-16
Port Forward Dashboard in experimental and classic KubeCluster

### LINKED ISSUE #485: Port forward dashboard automatically
Currently when creating a `dask_kubernetes.experimental.KubeCluster` the scheduler comm is automatically port forwarded locally on a random high port with `kubectl` if the scheduler service type is set to `ClusterIP` (which is the default). 

```python
>>> from dask_kubernetes.experimental import KubeCluster
>>> cluster = KubeCluster(name="foo")
>>> cluster.scheduler_address
'tcp://localhost:49839'
```

This means we can pass the `KubeCluster` object to `Client` and it will connect, but we could also pass the localhost address and it will work as that port forward is kept open as long as the cluster object exists.

https://github.com/dask/dask-kubernetes/blob/8190529fc140b6ea2c345bde02aa1c647272eb98/dask_kubernetes/common/networking.py#L68-L88

Perhaps we should also port forward the dashboard and ensure the `cluster.dashboard_link` is set correctly as it currently returns the address relative to the scheduler pod.

As the port forwarding code is common this would require a slight tweak to the classic KubeCluster as well, but this feature could benefit both.

mmccarty:
closes #485

### DIFF
--- a/dask_kubernetes/classic/kubecluster.py
+++ b/dask_kubernetes/classic/kubecluster.py
@@ -30,7 +30,10 @@ from ..common.utils import (
     namespace_default,
     escape,
 )
-from ..common.networking import get_external_address_for_scheduler_service
+from ..common.networking import (
+    get_external_address_for_scheduler_service,
+    port_forward_dashboard,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -212,6 +215,10 @@ class Scheduler(Pod):
             self.service,
             service_name_resolution_retries=self._service_name_retries,
         )
+
+        await port_forward_dashboard(
+            self.service.metadata.name,
+            self.namespace)
 
         self.pdb = await self._create_pdb()
 
--- a/dask_kubernetes/common/networking.py
+++ b/dask_kubernetes/common/networking.py
@@ -89,14 +89,20 @@ async def port_forward_service(service_name, namespace, remote_port, local_port=
 
 
 async def is_comm_open(ip, port, retries=10):
+    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
     while retries > 0:
-        try:
-            async with Client(f"tcp://{ip}:{port}", asynchronous=True, timeout=2):
-                return True
-        except Exception:
-            time.sleep(0.5)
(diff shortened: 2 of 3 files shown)


dask_kubernetes/common/networking.py
```diff
@@ -89,16 +89,22 @@ async def port_forward_service(service_name, namespace, remote_port, local_port=
...
+        if result == 0:
+            return True
+        else:
+            time.sleep(2)
             retries -= 1
     return False
 
 
+async def port_forward_dashboard(service_name, namespace):
+    port = await port_forward_service(service_name, namespace, 8787, 8787)
```
jacobtomlinson, line 104: When we port forward the scheduler comm we put it on a random high port, I wonder if we should do the same here to avoid collisions?

jacobtomlinson: Would also need to update the `daskboard_link` property with this address.

jacobtomlinson (commented):
Thanks for raising this @mmccarty! Couple of comments.

jacobtomlinson:
Looks like `flake8` and `black` are unhappy too. Could you run the `pre-commit` hooks to get everything to stop complaining?

dask_kubernetes/common/networking.py
mmccarty: Good idea.

dask_kubernetes/classic/kubecluster.py
```diff
@@ -626,6 +629,9 @@ async def _start(self):
 
         await super()._start()
 
+        port = await port_forward_dashboard(self.name, self.namespace)
+        self.scheduler_info["services"]["dashboard"] = port
```
mmccarty, line 633: @jacobtomlinson - Do you think its okay to mutate `scheduler_info` here? It was the only way I found to update `dashboard_link` with the way its currently written in [distributed](https://github.com/dask/distributed/blob/main/distributed/deploy/cluster.py#L323).

jacobtomlinson:
@project-administrator I've opened #533 to discuss this

PR #493, opened 2022-05-17
support running experimental KubeClusters and HelmClusters on the same loop as another SyncMethodMixin instance

graingert:
see also https://github.com/dask/distributed/issues/6163

### DIFF
--- a/dask_kubernetes/classic/tests/test_async.py
+++ b/dask_kubernetes/classic/tests/test_async.py
@@ -1,3 +1,4 @@
+import pytest_asyncio
 import asyncio
 import base64
 import getpass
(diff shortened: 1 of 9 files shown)


graingert:
currently the `distributed.LoopRunner` grabs the event loop from the current thread and runs it in a different thread.

which results in pytest-asyncio trying to close it when run in `auto` mode:
https://github.com/dask/dask-kubernetes/runs/6472536358?check_suite_focus=true#step:5:155


once https://github.com/dask/distributed/issues/6163 is fixed this should go away

jacobtomlinson:
CI failures here seem to be related to `install_signal_handlers` being dropped without warning in dask/distributed/pull/6205.

jacobtomlinson (approved):

PR #507, opened 2022-05-30
fix: Preserve custom pod affinities

### LINKED ISSUE #506: Custom affinity is overridden by default affinity
**What happened**:

I am trying to set a custom affinity on my Dask worker pods. I realise there is a [default affinity](https://github.com/dask/dask-kubernetes/issues/146) already defined, but I would prefer to use my own.

My custom affinity is something like this:
```
affinity:
  nodeAffinity:
    requiredDuringSchedulingIgnoredDuringExecution:
      nodeSelectorTerms:
        - matchExpressions:
            - key: foo
              operator: In
              values:
                - bar
```

However when the pod starts up, it only contains the default affinity (output from `kubectl get pod -o yaml`:
```
  affinity:
    nodeAffinity:
      - preference:
          matchExpressions:
            operator: In
            values:
            - worker
        weight: 100
```

**What you expected to happen**:

I expected the affinity I specified to be used in the pod spec.

Looking at the [merged PR](https://github.com/dask/dask-kubernetes/pull/147) for this issue, the behaviour is driven by an argument `match_node_purpose` to the function `clean_pod_template()`. However, there is no code that sets `match_node_purpose` to any value other than the default of `"prefer"`.

**Suggestion**:

Allow `KubeCluster` to accept a new argument and use it to set the `match_node_purpose` value in the call to `clean_pot_template()`. 

I might further suggest modifying the arg name and values to make it a bit clearer, e.g. `apply_default_affinity` which could be `preferred`, `required` or `none` (there might be an even clearer approach here but can't think of one right now).

csymeonides-mf:
Closes #506

### DIFF
--- a/dask_kubernetes/classic/kubecluster.py
+++ b/dask_kubernetes/classic/kubecluster.py
@@ -351,6 +351,9 @@ class KubeCluster(SpecCluster):
     deploy_mode: str (optional)
         Run the scheduler as "local" or "remote".
         Defaults to ``"remote"``.
+    apply_default_affinity: str (optional)
+        Apply a default affinity to pods: "required", "preferred" or "none"
+        Defaults to ``"preferred"``.
     **kwargs: dict
         Additional keyword arguments to pass to SpecCluster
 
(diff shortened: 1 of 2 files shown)


jacobtomlinson (approved):
Thanks for this. Sorry you're having trouble with affinities here. I'm hoping the new operator will give folks more control over things like this. Happy to merge this into the classic cluster manager though.

The CI failures are related to some upstream asyncio changes happening in distributed. I'll looking to get them resolved now and then I'll go ahead and merge here.

csymeonides-mf:
Thank you, much appreciated!

PR #523, opened 2022-06-06
set cluster name via constructor

### LINKED ISSUE #516: HelmCluster cluster names get carried over to subsequent Clusters
```python
import sys

from distributed.deploy import Cluster
import dask_kubernetes


def main():

    print(
        Cluster()
    )  # Cluster(e174b1b2, '<Not Connected>', workers=0, threads=0, memory=0 B)
    try:
        dask_kubernetes.HelmCluster(
            release_name="example-release", namespace="example-namespace"
        )
    except RuntimeError:
        pass  # I don't have helm installed

    print(
        Cluster()
    )  # Cluster(example-release.example-namespace, '<Not Connected>', workers=0, threads=0, memory=0 B)
    return 0


if __name__ == "__main__":
    sys.exit(main())
```

see also https://github.com/dask/distributed/pull/5305#discussion_r886822632

graingert:
Fixes #516

### DIFF
--- a/dask_kubernetes/experimental/kubecluster.py
+++ b/dask_kubernetes/experimental/kubecluster.py
@@ -133,7 +133,6 @@ class KubeCluster(Cluster):
         shutdown_on_close=None,
         **kwargs,
     ):
-        self.name = name
         self.namespace = namespace or namespace_default()
         self.image = image
         self.n_workers = n_workers
@@ -146,7 +145,7 @@ class KubeCluster(Cluster):
 
         self._instances.add(self)
 
-        super().__init__(**kwargs)
+        super().__init__(name=name, **kwargs)
         if not self.asynchronous:
             self._loop_runner.start()
             self.sync(self._start)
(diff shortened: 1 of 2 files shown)


graingert:
I added 6f04d5e24d363d483d5ec98c83bb2dd235d76c59 to try and debug the TimeoutErrors in https://github.com/dask/dask-kubernetes/runs/6752846378?check_suite_focus=true but they didn't occur in the second run :(

graingert:
@jacobtomlinson I tried to move KubeCluster's `self.name =` into a `super().__init__` call however the name is only computed in `async def _start(`

https://github.com/dask/dask-kubernetes/blob/c64b102ea9f33ae022383b8d0e16f09f6ecef1bd/dask_kubernetes/classic/kubecluster.py#L643

I think because `self._fill_pod_templates(` can only be called after
https://github.com/dask/dask-kubernetes/blob/c64b102ea9f33ae022383b8d0e16f09f6ecef1bd/dask_kubernetes/classic/kubecluster.py#L582

graingert:
the `self.name` assignment in `KubeCluster` doesn't exhibit the problem in https://github.com/dask/dask-kubernetes/issues/516 because the assignment happens after `Cluster.__init__` has been called.

jacobtomlinson (approved):
Many thanks!

PR #525, opened 2022-06-07
Fix dashboard forwarding

### LINKED ISSUE #517: Adding argument to disable port forward for kubeCluster
With a recent PR https://github.com/dask/dask-kubernetes/pull/489, dashboard is always port forwarded when kubeCluster is created. [Link](https://github.com/dask/dask-kubernetes/blob/2022.5.2/dask_kubernetes/classic/kubecluster.py#L647).
This method calls `check_dependency("kubectl")` which checks via shutil that requires kubectl to be installed. However, there are use cases where the use of port forwarding the dashboard and having kubectl is not required.

Could we get an option to disable automatic port fowarding of dashboard as it used to be while initialising kubeCluster?
Thanks.

jacobtomlinson:
Closes #517.

In #509 we fixed the dashboard port forwarding in `experimental.KubeCluster` when running from within the Kubernetes cluster. This PR adds the same fix to `classic.KubeCluster`.

cc @mmccarty

### DIFF
--- a/dask_kubernetes/classic/kubecluster.py
+++ b/dask_kubernetes/classic/kubecluster.py
@@ -31,7 +31,7 @@ from ..common.utils import (
 )
 from ..common.networking import (
     get_external_address_for_scheduler_service,
-    port_forward_dashboard,
+    get_scheduler_address,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


mmccarty (approved):
:+1:

PR #530, opened 2022-06-29
Set operator to use host network

jacobtomlinson:


jacobtomlinson:
Can't remember why I did this. The branch name suggests it is something to do with AWS but I'm going to close this out until it comes up again.

PR #531, opened 2022-06-30
Updated requirements.txt with libraries due to which some tests were being skipped

shaswat-indian:
- Added libraries: `bokeh`, `ipywidgets` and `numpy`
- Corrected some typos in the operator documentation


requirements.txt
```diff
@@ -2,4 +2,7 @@ dask>=2021.02.0
 distributed>=2021.02.0
 kubernetes>=12.0.1
 kubernetes-asyncio>=12.0.1
-kopf>=1.35.3
\ No newline at end of file
+kopf>=1.35.3
+bokeh>=2.4.3
+ipywidgets>=7.7.1
+numpy==1.23.0
```
jacobtomlinson, line 8: I don't think we should be hard pinning this here. Probably needs to be a greater than.

```suggestion
numpy>=1.23.0
```

Typically installing `dask[complete]` with `pip` or `dask` with `conda` brings in these dependencies so I'm a little hesitant to add them here too.

jacobtomlinson (changes requested):
I'm guessing you installed `dask-kubernetes` without first installing `dask`. This resulted in just core `dask` being installed which has these packages as optional. I don't think we want to depend on these as being hard here either.

Given that the only problem here is that some tests are skipped due to missing dependencies I think it would be better to add them to `requirements-test.txt` rather than `requirements.txt` to make them test only dependencies.

https://github.com/dask/dask-kubernetes/blob/main/requirements-test.txt

requirements.txt
shaswat-indian: I've moved them to `requirements-test.txt` but let me know if we want to remove them altogether since they come along with `dask`.

jacobtomlinson: Yeah we could switch the dask dependency in `requirements-test.txt` to `dask[complete]` which should pull those in. That means we don't need to maintain the minimum versions in multiple places.

shaswat-indian: Added `dask[complete]` to `requirements-test.txt`.

jacobtomlinson (approved):

PR #538, opened 2022-07-29
Allow custom worker command

### LINKED ISSUE #537: Allow users to set the worker command in experimental.KubeCluster
Currently, it isn't possible to set the worker command via the Python API, only via the YAML.

#475 covers allowing users to pass a full spec to `KubeCluster`, which is one route to solving this, but insisting users set the whole pod spec is definitely a nuclear option.

Exposing some common options as kwargs makes sense, but perhaps we should consider allowing folks to patch the spec too instead of having to supply the whole thing.

jacobtomlinson:
Allow users to set a custom worker command in `dask_kubernetes.experimental.KubeCluster`.

```python
cluster = KubeCluster(name="customworkercommand",
                      worker_command=["python", "-m", "distributed.cli.dask_worker"])
```

Closes #537

### DIFF
--- a/ci/install-deps.sh
+++ b/ci/install-deps.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 curl -L https://istio.io/downloadIstio | sh -
 mv istio-*/bin/istioctl /usr/local/bin/istioctl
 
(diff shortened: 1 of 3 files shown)


jacobtomlinson:
Tes failures are unrelated and being investigated in #535

PR #545, opened 2022-08-16
Add a status to DaskCluster resources and use that to timeout the KubeCluster

### LINKED ISSUE #490: Cluster manager should check if operator is running
The new experimental cluster manager relies on the operator and CRDs to create the Dask cluster components for us. 

We need to ensure that the user experience is good for folks using the cluster manager but do not have the CRDs or operator running. 

If no CRDs are installed it should fail early and give a useful error message. 

If the operator is not running we need some way to detect that and also give an error message. Perhaps the operator should add a label to the DaskCluster resource to show it has been picked up. We could watch for that label and if it doesn't appear after a timeout we could show a useful error.

jacobtomlinson:
Closes  #490.

I've added a `status.phase` property to the `DaskCluster` resource and implemented two statuses `Created` and `Running`.

I've updated the controller so that the creation event on `DaskCluster` resources just sets the phase to `Created`. I've then created another hook that watches for that phase being set which then creates the downstream resources, once they have been created the phase is updated to `Running`.

Then the `experimental.KubeCluster` now has a `_wait_for_operator` method that watches for this `status.phase` property. If this doesn't get set after 60 seconds (configurable) the cluster creation will time out because the operator controller is likely not running or in a failed state. This is great because currently, the class will just hang if the controller isn't running.

### DIFF
--- a/dask_kubernetes/experimental/kubecluster.py
+++ b/dask_kubernetes/experimental/kubecluster.py
@@ -85,8 +85,11 @@ class KubeCluster(Cluster):
     shutdown_on_close: bool (optional)
         Whether or not to delete the cluster resource when this object is closed.
         Defaults to ``True`` when creating a cluster and ``False`` when connecting to an existing one.
-    operator_timeout: int (optional)
-        Time in seconds to wait for the operator to create cluster resources before giving up.
+    resource_timeout: int (optional)
+        Time in seconds to wait for the controller to take action before giving up.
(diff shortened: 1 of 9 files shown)


jacobtomlinson:
It would also be nice in the future if the `DaskCluster` went into a `Pending` phase once the other resources have been created and only move into the final `Running` phase when the scheduler readiness probe passes. I'll open an issue to track this as a follow up task.

PR #556, opened 2022-09-12
Test against multiple k8s versions

jacobtomlinson:
For a long time we've tested against the latest version of Kubernetes that `kind` uses by default. With the recent `1.25` release there seem to be some breakages, but fixing those could cause problems with older Kubernetes versions.

We don't pin this currently so CI failures started with the latest `kind` release which was outside of our control.

To make things clearer and easier to test I've explicitly set which Kubernetes version to test against and updated the CI matrix to also test against all supported Kubernetes versions and added a section to the docs to describe which versions we support.

The proposed support is for Python we follow [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) along with the rest of Dask and for Kubernetes we will follow the [yearly support KEP](https://kubernetes.io/releases/patch-releases/#support-period).

### DIFF
--- a/.github/workflows/helmcluster.yaml
+++ b/.github/workflows/helmcluster.yaml
@@ -2,12 +2,14 @@ name: "HelmCluster"
 on:
   pull_request:
     paths:
+      - ".github/workflows/helmcluster.yaml"
       - "ci/**"
       - "dask_kubernetes/helm/**"
       - "dask_kubernetes/common/**"
       - "dask_kubernetes/*"
   push:
     paths:
+      - ".github/workflows/helmcluster.yaml"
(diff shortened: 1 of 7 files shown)


jacobtomlinson:
I can't decide if I should test all supported Python versions against all supported Kubernetes versions, or just one of each.

**All**
image

**One of each**
image

I'll ask twitter https://twitter.com/_JacobTomlinson/status/1569321563912282114?s=20&t=gUTFyHaUgg39OnzUE2g3Pg

jacobtomlinson:
Going to merge as is and can always tweak the matrix again later.

PR #558, opened 2022-09-12
Try running tests in parallel

jacobtomlinson:
The CI is taking nearly 30 minutes at the moment, experimenting with running tests in parallel to bring that time down.

Initial timings:

| Suite | Time (approx) |
| --- | --- |
| `HelmCluster` | 5 mins |
| `KubeCluster` (classic) | 28 mins |
| `KubeCluster` (operator) | 21 mins |

### DIFF
--- a/.github/workflows/kubecluster.yaml
+++ b/.github/workflows/kubecluster.yaml
@@ -38,7 +38,7 @@ jobs:
       - name: Run tests
         env:
           KUBERNETES_VERSION: ${{ matrix.kubernetes-version }}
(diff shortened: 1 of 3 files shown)


jacobtomlinson:
Failures are happening because the first two tests in each process are both trying to create the session fixtures and are clobbering each other.

We lave the k8s cluster running after the tests have run, so one solution could be to run a single test without parallelism first which would cause the cluster to be created. Then run all the tests with some parallelism which would skip the cluster setup.

jacobtomlinson:
Not relevent any more as the classic and helm cluster managers have been removed.

PR #565, opened 2022-09-28
Fix potential typo in kubecluster.py

patrik93:
Based on contexts, it should be "deleted" instead of "deleated"


jacobtomlinson (approved):

jacobtomlinson:
ty

PR #568, opened 2022-10-03
Deprecate classic KubeCluster

jacobtomlinson:
Start switching out the classic `KubeCluster` for the operator-based implementation.

- [x] Graduate `dask_kubernetes.experimental.KubeCluster` to `dask_kubernetes.operator.KubeCluster`
- [x] Add deprecation warning to top-level import `from dask_kubernetes import KubeCluster`
- [x] Moved the operator controller into `dask_kubernetes.operator.controller` to make the naming more clear and to free up `dask_kubernetes.operator` for importing the new cluster manager from.
- [x] Updated top-level docs to describe new `KubeCluster` implementation first
- [x] Document migrating from the classic `KubeCluster`


mrocklin:
Fun!

jacobtomlinson:
I had originally intended to put a shim in place so that if folks tried to use `dask_kubernetes.operator` with kwargs from the classic implementation it would fall back cleanly. This way I would swap out the top-level import for the new one.

However, this has proved to be very brittle in my testing and was causing headaches with making class methods available. I was inspired by `SSHCluster` which replaced the whole class with a shim function, but `KubeCluster` seems to be complex enough that this would be too much of a compromise.

Instead, I've just added a deprecation warning to importing `from dask_kubernetes import KubeCluster` and am encouraging folks to explicitly import from `dask_kubernetes.classic` or `dask_kubernetes.operator`.

Issue #571, opened 2022-10-10
Ability to add annotations to dask k8s resources


bstadlbauer:
As discussed in [this discourse post](vhttps://dask.discourse.group/t/add-annotation-to-worker-pods-when-using-the-dask-k8s-operator/1175), it might be nice to have a possibility of adding annotations to `dask`-related resources (job-runner, scheduler as well as worker pods).

Per the discussion on discourse, this could happen by implicitly passing down annotations from the top-level `dask` custom resources (e.g. `DaskCluster`, `DaskWorkerGroup`, `DaskJob`) to the created k8s `Pod`s.

PR #572, opened 2022-10-10
Pass on annotations from dask custom resources

bstadlbauer:
As described in https://github.com/dask/dask-kubernetes/issues/571, it would be nice if there was a possibility to add annotations to `dask`-related pods.

This PR adds this capability by: 
- Passing on all annotations set on the `DaskCluster` CR to the scheduler `Pod` as well as the `DaskWorkerGroup` CR
- Passing on all annotations set on the `DaskWorkerGroup` CR to the worker `Pod`s
- Passing on all annotation set on the `DaskJob` CR to the job-runner `Pod`

A small issue I've noticed when implementing this was that there are already some "default" `kopf` and kubernetes related annotations on resources, which I've chosen to ignore: 
-   "kopf.zalando.org/last-handled-configuration"
-   "kubectl.kubernetes.io/last-applied-configuration"

Feel free to suggest alternatives, happy to change things up

Fixes https://github.com/dask/dask-kubernetes/issues/571

### DIFF
--- a/dask_kubernetes/operator/controller/controller.py
+++ b/dask_kubernetes/operator/controller/controller.py
@@ -109,7 +109,7 @@ def build_worker_pod_spec(
     return pod_spec
 
 
-def build_job_pod_spec(job_name, cluster_name, namespace, spec):
+def build_job_pod_spec(job_name, cluster_name, namespace, spec, annotations):
     pod_spec = {
         "apiVersion": "v1",
         "kind": "Pod",
@@ -120,6 +120,7 @@ def build_job_pod_spec(job_name, cluster_name, namespace, spec):
                 "dask.org/component": "job-runner",
                 "sidecar.istio.io/inject": "false",
             },
+            "annotations": annotations,
         },
         "spec": spec,
     }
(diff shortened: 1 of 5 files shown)


jacobtomlinson (commented):
This looks really great, thanks for raising this!

I'm happy with the way you have excluded some of the defaults, I think that makes sense. I'd maybe be tempted to go a step further and exclude the whole `kopf.zalando.org` and `kubectl.kubernetes.io` namespaces.

I also noticed you're not passing these annotations to the scheduler service. Any reason for this?

timblakely:
Just chiming in from the peanut gallery to say that I've patched this in to a custom operator image and it appears to work as intended. Adding `cluster-autoscaler.kubernetes.io/safe-to-evict` to the `custom_cluster_spec` when creating the cluster looks like it propagated correctly to the downstream `daskworkergroup`.

jacobtomlinson (approved):
Awesome thanks! CI failure is unrelated, merging.

(Thanks for trying it out @timblakely!)

PR #574, opened 2022-10-11
Fix string escaping in kubernetes.yaml

timblakely:
Looks like there's a missing `"` in the `image:` field of the Operator. This adds it :)


jacobtomlinson (commented):
Thanks for raising this! I'm afraid it's a duplicate of #569 though.

PR #576, opened 2022-10-12
Add daskautoscalers in the cluster role rules

john-jam:
This PR adds the `daskautoscalers` resource in the `clusterrole` rules.
Without it, the `dask-operator` pod fail to start with the error:
```
[2022-10-12 17:46:55,501] kopf._core.reactor.o [ERROR   ] Watcher for daskautoscalers.v1.kubernetes.dask.org@none has failed: ('daskautoscalers.kubernetes.dask.org is forbidden: User "system:serviceaccount:dask:dask-kubernetes-operator" cannot list resource "daskautoscalers" in API group "kubernetes.dask.org" at the cluster scope')
```

PR #578, opened 2022-10-18
Add missing `daskautoscalers` permission

timblakely:
Looks like `operator.yaml` was missing the `daskautoscalers` permission. This PR adds it :)


jacobtomlinson (approved):
Thanks, we should've included this in #576.

PR #580, opened 2022-10-20
Add status to dask job

### LINKED ISSUE #573: Add status to `DaskJob` CRD
Would it be possible to track the status of a job in the toplevel `DaskJob` CR? This would have the advantage of hiding the "implementation details" of the job from a user. I.e. instead of having to know about the state of a `job-runner` pod that gets created as part of a `DaskJob`, one could only look at the `DaskJob` resource to get that information.

A particular use case here would be the Flyte plugin I am currently working on. Most Flyte (backend) plugins work by creating a k8s resource which they continuously poll for updates. The way the plugin machinery is implemented does not allow to reach out to arbitrary k8s resources as the initially created resource is updated and passed into the plugin from the outside. 
The exact interface definition is [here](https://github.com/flyteorg/flyteplugins/blob/eed7326340039dbdfcfae75ae6911219ff49371d/go/tasks/pluginmachinery/k8s/plugin.go#L74-L89), where the `resource` argument in `Plugin.GetTaskPhase()` would correspond to the `DaskJob` CR created in `Plugin.BuildResource()`

Similar Job objects do also set a `Status` field on their Custom Resources, e.g.: 
- `ray`: [here](https://github.com/ray-project/kuberay/blob/b6b00c87722d7b85274889ffc0477f34b3ee6f76/ray-operator/apis/ray/v1alpha1/rayjob_types.go#L79-L85) and [here](https://github.com/ray-project/kuberay/blob/b6b00c87722d7b85274889ffc0477f34b3ee6f76/ray-operator/apis/ray/v1alpha1/rayjob_types.go#L57-L73)
- `spark`: [here](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/94775cd89ca0158e869fac39e1b097d1bf56a7e8/pkg/apis/sparkoperator.k8s.io/v1beta2/types.go#L90-L95) and [here](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/94775cd89ca0158e869fac39e1b097d1bf56a7e8/pkg/apis/sparkoperator.k8s.io/v1beta2/types.go#L137-L154)

Happy to contribute something in case we can align on a format of the status. My initial proposal would be: 
```go
type JobStatus string

const (
	JobStatusPending   JobStatus = "PENDING"
	JobStatusRunning   JobStatus = "RUNNING"
	JobStatusStopped   JobStatus = "STOPPED"
	JobStatusSucceeded JobStatus = "SUCCEEDED"
	JobStatusFailed    JobStatus = "FAILED"
)

type DaskJobStatus struct {
	DaskClusterName             string        `json:"rayClusterName,omitempty"`
	JobStatus                   JobStatus     `json:"jobStatus,omitempty"`
	StartTime                   *metav1.Time  `json:"startTime,omitempty"`
	EndTime                     *metav1.Time  `json:"endTime,omitempty"`
}
```

cc @hamersaw

bstadlbauer:
This PR aims at adding some `status` information to the `DaskJob` CRD.

In particular it adds the following fields to `status`: 
- `jobStatus`: Status of the job. string, required. Can by any of: 
    - `JobCreated`: Job CR was created
    - `ClusterCreated`: The `DaskCluster` CR associated with the job has been created
    - `Running`: The job-runner pod is in "Running" state
    - `Successful`: The job-runner pod is in `Succeeded` state
    - `Failed`: The job-runner pod is in the `Failed` statue
- `clusterName`: Name of the `DaskCluster` resorce to be created (string)
- `jobRunnerPodName`: The name of the job-runner pod
- `startTime`: Time when the job-runner pod started running
- `endTime`: Time when the job-runner pod went into the `Succeeded` state

@jacobtomlinson: Happy to change any of the stated/the behavior in general. 

The `test_job` test is also becoming quite large, let me know if I should pull some functions out or split it into multiple tests.

cc @hamersaw

**Edit 2022-10-21**: 
- Split the `Completed` state in into a `Successful` and `Failed` state
- Added `jobRunnerPodName`

Closes #573

### DIFF
--- a/dask_kubernetes/operator/controller/controller.py
+++ b/dask_kubernetes/operator/controller/controller.py
@@ -23,9 +23,11 @@ _ANNOTATION_NAMESPACES_TO_IGNORE = (
     "kubectl.kubernetes.io",
 )
 
+
 class DaskClusterPhase(enum.Enum):
     CREATED = "Created"
     RUNNING = "Running"
+
 
 # Load operator plugins from other packages
 PLUGINS = []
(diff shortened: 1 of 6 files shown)


jacobtomlinson (commented):
This looks really great I'm really happy with this functionality.

My only comment is that we are generally preferring `Literal` over `Enum` in other places in Dask, so we might want to do the same here.

Context https://github.com/dask/distributed/pull/6377#discussion_r926953704

bstadlbauer:
@jacobtomlinson Thank you for the review!

I've made some small changes: 
- Added a `Failed` state, in the case the `job-runner` pod fails (this was not being accounted for before, so the cluster would also not clean up). I've split the `Completed` state into a `Successful` and `Failed` state respectively. As part of this, I've also slightly refactored the job-test(s) to reduce code duplication.
- Added `jobRunnerPodName` to the status. This could come in hand when getting logs, etc. A few other frameworks do this aswell, so I thought it wouldn't hurt.
- Looked into converting the `enum`'s into  `Literal`s. I have never used `Literal`, but from my limited understanding, this would only help in case there is typing? 🤔 Converted things back to plain strings instead, but happy to change things up

jacobtomlinson (commented):
Thanks for fixing things up here, those changes sound great.

We should probably do better with type annotations in the controller here, but we can save that for a future PR.

The test failures seem to be related, I think one of the strings may be incorrect.

bstadlbauer:
@jacobtomlinson  Just had a look at the tests, I think what's happening is a racecondition there. The cluster has already gone from the `JobCreated` into `ClusterCreated` state before the status was checked - would you have an idea of how to better test this? 🤔 

Also, happy to follow up with type annotations for the controller if that's something that's on the roadmap

jacobtomlinson:
> would you have an idea of how to better test this?

I think the main thing being tested is that the states are progressing forwards. If the controller didn't pick it up correctly it would just not have a state. So maybe update the test to assert that state is set to anything.

> Also, happy to follow up with type annotations for the controller if that's something that's on the roadmap

Any code hygiene contributions like this would be appreciated.

bstadlbauer:
@jacobtomlinson Good point! Changed the test logic to only assert if `jobStatus` is present. For all other attributes the checks should be fine, as there will only be properties added and their value never changes. 

> Any code hygiene contributions like this would be appreciated.
Good to know! I'll try to get the Flyte plugin across the line, but would be open to some contribution. Will let you know before I would start

jacobtomlinson (approved):
Awesome many thanks for the efforts here!

Issue #582, opened 2022-10-23
Can't start cluster from a pod in the cluster


jadeidev:
**Describe the issue**: Cannot start KubeCluster (using the operator) from a pod that runs on the K8s cluster 

**Minimal Complete Verifiable Example**:

- Start new minikube cluster.
- Install Dask-Kubernetes helm per instructions.
- Start a python:3.8 Pod on the cluster with dask-kubernetes installed on it. 
- Start pod shell
within the pod shell run python and start KubeCluster:
```python
from dask_kubernetes.operator import KubeCluster
cluster = KubeCluster(name="mycluster", image='ghcr.io/dask/dask:latest')
```
Result:
getting permissions issue:
```python
Traceback (most recent call last):
  File "/opt/conda/lib/python3.8/site-packages/dask_kubernetes/operator/kubecluster/kubecluster.py", line 267, in _create_cluster
    await custom_objects_api.create_namespaced_custom_object(
  File "/opt/conda/lib/python3.8/site-packages/kubernetes_asyncio/client/api_client.py", line 192, in __call_api
    raise e
  File "/opt/conda/lib/python3.8/site-packages/kubernetes_asyncio/client/api_client.py", line 185, in __call_api
    response_data = await self.request(
  File "/opt/conda/lib/python3.8/site-packages/kubernetes_asyncio/client/rest.py", line 230, in POST
    return (await self.request("POST", url,
  File "/opt/conda/lib/python3.8/site-packages/kubernetes_asyncio/client/rest.py", line 187, in request
    raise ApiException(http_resp=r)
kubernetes_asyncio.client.exceptions.ApiException: (403)
Reason: Forbidden
HTTP response headers: <CIMultiDictProxy('Audit-Id': 'deleted', 'Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'X-Content-Type-Options': 'nosniff', 'X-Kubernetes-Pf-Flowschema-Uid': 'deleted', 'X-Kubernetes-Pf-Prioritylevel-Uid': 'deleted', 'Date': 'Sun, 23 Oct 2022 04:34:36 GMT', 'Content-Length': '377')>
[...]

PR #583, opened 2022-10-23
Update operator_kubecluster.rst

jadeidev:
update docs for starting KubeCluster from pod in the cluster

### DIFF
--- a/doc/source/operator_kubecluster.rst
+++ b/doc/source/operator_kubecluster.rst
@@ -155,8 +155,7 @@ following ClusterRole to that ServiceAccount via a ClusterRoleBinding:
        resources: [daskclusters, daskworkergroups, daskworkergroups/scale, daskjobs, daskautoscalers]
        verbs: [get, list, watch, patch, create, delete]
 
-     # Application: other resources it produces and manipulates.
-     # Here, we create/delete Pods.
+     # Application: other resources it needs to watch and get information from.
      - apiGroups:
        - ""  # indicates the core API group
        resources: [pods, pods/status]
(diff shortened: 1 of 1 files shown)


jadeidev:
Closes #582

doc/source/operator_kubecluster.rst
```diff
@@ -136,6 +136,62 @@ The ``cluster.add_worker_group()`` method also supports passing a ``custom_spec`
...
+       verbs: ["*"]
+
+     - apiGroups: 
+       - ""  # indicates the core API group
+       resources: [services]
+       verbs: ["*"]
+
+     - apiGroups: ["", events.k8s.io]
+       resources: [events]
+       verbs: ["*"]
```
jacobtomlinson, line 172: I don't think we should need any of these as the controller handles creating all of the objects.

jacobtomlinson (commented):
Many thanks for this, totally agree this is a good thing to document.

I don't think we need all of the permissions you have listed though. AFAIK we need the full `kubernetes.dask.org` resources and verbs and then a few get/list operations to get things like logs and statuses.

Could you double-check the permissions you have included here and remove any that are unnecessary?

doc/source/operator_kubecluster.rst
jadeidev: @jacobtomlinson I have updated the verbs and resources needed. I believe these are the minimum requirements to get things going.

doc/source/operator_kubecluster.rst
```diff
@@ -136,6 +136,55 @@ The ``cluster.add_worker_group()`` method also supports passing a ``custom_spec`
...
+   metadata:
+     name: dask-cluster-role
+   rules:
+     # Application: watching & handling for the custom resource we declare.
+     - apiGroups: [kubernetes.dask.org]
+       resources: [daskclusters, daskworkergroups, daskworkergroups/scale, daskjobs, daskautoscalers]
+       verbs: [get, list, watch, patch, create, delete]
+
+     # Application: other resources it produces and manipulates.
+     # Here, we create/delete Pods.
```
jacobtomlinson, line 159:
```suggestion
     # Application: other resources it needs to watch and get information from.
```

jacobtomlinson:
Many thanks!

Issue #592, opened 2022-10-27
Error: Dask Cluster resource not actioned after 60 seconds, is the Dask Operator running?


zoonman:
Hi guys. I am using v2022.10.1 and installed CRDs, I see the operator running. 

My code is a trivial example:
```python
import logging

from dask_kubernetes.operator import KubeCluster
from dask.distributed import Client

def square(x):
    return x ** 2

def neg(x):
    return -x

def main():
    logger = logging.getLogger("main")
    logger.setLevel('DEBUG')

    logger.info("Initializing cluster")
    with KubeCluster() as cluster:
        logger.info("Initializing client")
        with Client() as client:
            logger.info("Creating futures")

            A = client.map(square, range(10))
            B = client.map(neg, A)

            logger.info("Submitting the task...")
            total = client.submit(sum, B)

            logger.info("Getting result...")
            result = total.result()
            logger.info("Result: %s", result)


if __name__ == '__main__':
    main()
```
The code eventually fails with the following stack trace:
```
python -m my.dt
2022-10-27 12:56:49.486 INFO:main:main:Initializing cluster
2022-10-27 12:56:49.493 DEBUG:asyncio:__init__:Using selector: EpollSelector
2022-10-27 12:56:49.493 DEBUG:asyncio:__init__:Using selector: EpollSelector
2022-10-27 12:56:49.517 DEBUG:kubernetes_asyncio.client.rest:request:response body: b'{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"daskclusters.kubernetes.dask.org \\"dask-root-ba8bbba9-e\\" not found","reason":"NotFound","details":{"name":"dask-root-ba8bbba9-e","group":"kubernetes.dask.org","kind":"daskclusters"},"code":404}\n'
[...]

PR #593, opened 2022-10-27
Explicitly specify operator's image version

zoonman:
I know this is not a generic solution but it is better than a broken operator deployment.

Addresses https://github.com/dask/dask-kubernetes/issues/592

### DIFF
--- a/dask_kubernetes/operator/deployment/manifests/operator.yaml
+++ b/dask_kubernetes/operator/deployment/manifests/operator.yaml
@@ -26,7 +26,7 @@ spec:
       serviceAccountName: dask-kubernetes-operator
       containers:
         - name: operator
-          image: ghcr.io/dask/dask-kubernetes-operator:latest
+          image: ghcr.io/dask/dask-kubernetes-operator:2022.10.1
           imagePullPolicy: IfNotPresent
 ---
 apiVersion: v1


jacobtomlinson (commented):
Really appreciate you raising this but I'm working on a longer-term solution that I intend to raise today so I'm going to close this out.

zoonman:
You can move the `latest` tag. It will save the others from suffering ;-)

jacobtomlinson:
That's a good idea. I can't seem to figure out how to delete the `latest` tag on ghcr.io though...

jacobtomlinson:
Also as of #594 the only folks affected by this is anyone using the broken `:latest` referencing manifests that have now been deleted. This shouldn't be a problem going forwards.

zoonman:
If you pull the old image and then:

```shell

docker image tag ghcr.io/dask/dask-kubernetes-operator:2022.10.1  ghcr.io/dask/dask-kubernetes-operator:latest

docker image push ghcr.io/dask/dask-kubernetes-operator:latest
```

jacobtomlinson:
I don't want to update the tag, I want to delete it altogether. This is possible on Docker Hub and other registries but I can't seem to figure out how to do this on GitHub.

We will not need a `latest` tag going forwards because we are tightly pinning the chart/manifests to a specific version to ensure compatibility. So it makes sense to delete the old `latest` tags to avoid folks from falling into a trap.

PR #586, opened 2022-10-26
Consolidate install manifests

jacobtomlinson:
This PR reduces the number of kubectl commands needed to install the operator from five manifests to one.

```console
$ kubectl apply -f https://raw.githubusercontent.com/dask/dask-kubernetes/main/dask_kubernetes/operator/deployment/manifests/full.yaml
customresourcedefinition.apiextensions.k8s.io/daskautoscalers.kubernetes.dask.org created
customresourcedefinition.apiextensions.k8s.io/daskclusters.kubernetes.dask.org created
namespace/dask-operator created
deployment.apps/dask-kubernetes-operator created
serviceaccount/dask-kubernetes-operator created
clusterrole.rbac.authorization.k8s.io/dask-kubernetes-operator-role-cluster created
clusterrolebinding.rbac.authorization.k8s.io/dask-kubernetes-operator-rolebinding-cluster created
customresourcedefinition.apiextensions.k8s.io/daskjobs.kubernetes.dask.org created
customresourcedefinition.apiextensions.k8s.io/daskworkergroups.kubernetes.dask.org created
```

We currently store all our CRDs in separate YAML files and the controller resources in another, this is nice if you just want to install/upgrade different pieces. However, our documentation gets a new `kubectl apply` command every time we add a new CRD.

This PR adds a `pre-commit` hook that runs a bash script to concatenate all of the manifests into a single one called `full.yaml`. So we still have the separate manifests for those that want them, but now we also have a single manifest with everything in.

_Note this doesn't resolve the fact that the manifest method uses a `:latest` tag that we are trying to resolve in #508._

PR #594, opened 2022-10-28
Make Helm the only supported Operator install method

### LINKED ISSUE #508: Operator docker image didn't get latest tag on release
When `2022.5.1` was released the `ghcr.io/dask/dask-kubernetes-operator:2022.5.1` Docker image was built and published by Chartpress. However `ghcr.io/dask/dask-kubernetes-operator:latest` wasn't updated and continued to point to the previous release.

I've manually tagged `latest` and updated the registry so everything is ok now, but we need to fix this for the next release.

I'm guessing something needs changing in the Chartpress config to make this happen but looking at the [config reference](https://github.com/jupyterhub/chartpress#configuration) I can't immediately see what needs changing.

@consideratio could you point me in the right direction here?

jacobtomlinson:
To have a good installation experience we need to ensure the version of the operator manifests being installed matches the version of the controller docker image. We currently support two installation methods; static manifests and helm. When we package the helm chart we bake in the docker image tag so the chart version is tightly pinned to the container version.

However, the static manifests use `:latest` which doesn't get updated at all since we switched to using chartpress to package our container images too (xref #508). The Operator hasn't been around long but we are already seeing users being bitten by the "latest is not a version" anti-pattern.

To resolve this I've decided to remove the static manifests altogether as getting the pinning working with `pre-commit` hooks just seems like a minefield and installing YAML files from a `main` branch of a repo on GitHub doesn't seem to be a great default practice. Instead, I've made `helm` the primary install method but also documented using `helm template` to export the static manifests from the latest release for those that prefer not to use `helm`. 

**Install with Helm**
```console
$ helm install --repo https://helm.dask.org --create-namespace -n dask-operator --generate-name dask-kubernetes-operator
```

**Install with Kubectl**
```console
$ helm template --include-crds --repo https://helm.dask.org release dask-kubernetes-operator | kubectl apply -f -
```

This means everyone needs to have `helm` installed, but that seems like a fine compromise and should improve user experience and reduce maintainer overhead.

Closes #508.
Supersedes #586.


jacobtomlinson:
@consideratio I'd love your feedback on this approach

consideRatio:
I'm a very big +1 on centralizing on a single install method (not distinguishing helm temlate from helm install)! I think it is a win-win for maintainers and users.

Also, super plus on not referencing the gitrepo local helm chart when using helm template, but instead referencing the packaged helm chart in the helm chart repo - as already done in your example above. This is because the packaged helm chart should include values.schema.json and updated values.yaml, while updated values / schema are not available in the github repo without pre-processing.

PR #605, opened 2022-11-10
Ensure workers are recreated if pods are deleted/lost

### LINKED ISSUE #603: Worker group doesn't recover after pod deletion
**Describe the issue**:

```
> kubectl get pods -owide
```

```
> kubectl get pods -owide
```

**Environment**:

- Dask version: `2022.10.0`

`dask.yaml`:

```yaml
kind: DaskCluster
metadata:
  name: kai-dask
spec:
  scheduler:
    service:
      ports:
      - name: tcp-comm
        port: 8786
        protocol: TCP
        targetPort: tcp-comm
      - name: http-dashboard
        port: 8787
        protocol: TCP
        targetPort: http-dashboard
      selector:
        dask.org/component: scheduler
      type: ClusterIP
    spec:
      containers:
      - command:
        - dask
        - scheduler
        imagePullPolicy: Always
        livenessProbe:
          httpGet:
            path: /health
            port: http-dashboard
          initialDelaySeconds: 15
          periodSeconds: 20
        name: scheduler
        ports:
        - containerPort: 8786
          name: tcp-comm
          protocol: TCP
        - containerPort: 8787
          name: http-dashboard
          protocol: TCP
        readinessProbe:
          httpGet:
            path: /health
            port: http-dashboard
          initialDelaySeconds: 5
          periodSeconds: 10
        resources:
          limits:
            cpu: 500m
            memory: 512Mi
  worker:
    replicas: 3
    spec:
      containers:
      - args:
        - --name
        - $(DASK_WORKER_NAME)
        command:
        - dask
        - worker
        imagePullPolicy: Always
        name: worker
        resources:
          limits:
            cpu: "1"
            memory: 2Gi
```

jacobtomlinson:
Closes #603 

Added a timer to the `daskworkergroup_update` method so that the scaling logic is run every 5 seconds as well as when the group is updated. This function checks if the current number of pods matches the expected and creates/removes accordingly. 

If a node is terminated and pods are lost this timer should notice the mismatch and correct it.

Also added a test to check for this.

This does leave open the question of what about other resources like the scheduler pod or job runner pod. Should those also be recreated automatically, or should the parent resource go into an error state?

### DIFF
--- a/dask_kubernetes/operator/controller/controller.py
+++ b/dask_kubernetes/operator/controller/controller.py
@@ -362,6 +362,7 @@ async def get_desired_workers(scheduler_service_name, namespace, logger):
 
 
 @kopf.on.update("daskworkergroup")
+@kopf.timer("daskworkergroup", interval=5.0)
 async def daskworkergroup_update(spec, name, namespace, logger, **kwargs):
     async with kubernetes.client.api_client.ApiClient() as api_client:
         api = kubernetes.client.CoreV1Api(api_client)
--- a/dask_kubernetes/operator/controller/tests/test_controller.py
+++ b/dask_kubernetes/operator/controller/tests/test_controller.py
@@ -134,6 +134,29 @@ async def test_scalesimplecluster(k8s_cluster, kopf_runner, gen_cluster):
                     await client.wait_for_workers(3)
 
 
+@pytest.mark.asyncio
+async def test_group_recovers_from_pod_deletion(k8s_cluster, kopf_runner, gen_cluster):
+    with kopf_runner as runner:
+        async with gen_cluster() as cluster_name:
+            scheduler_pod_name = "simple-scheduler"
+            service_name = "simple-scheduler"
+            while "Running" not in k8s_cluster.kubectl(
+                "get", "pods", scheduler_pod_name
+            ):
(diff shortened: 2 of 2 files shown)


kwohlfahrt:
> Should those also be recreated automatically, or should the parent resource go into an error state?

IMO an error state makes more sense for these. I would expect a job to possibly have setup/teardown logic that isn't idempotent, so automatically restarting it seems potentially unsafe, an error state would be safer. Or possibly a configurable number of max retries?

kwohlfahrt:
To keep the load on the API server lower (especially with many worker pods), we should possibly use an [index](https://kopf.readthedocs.io/en/stable/indexing/) to track the live workers. I'd be happy to contribute a patch for this.

jacobtomlinson:
> IMO an error state makes more sense for these.

Appreciate the input on that. I opened #606 to track this conversation further.

> I'd be happy to contribute a patch for this.

Nice! I've not used `kopf` indexes before. Yeah if you want to submit a PR that supersedes this one that would be awesome!

jacobtomlinson:
@tasansal I expect this PR will be superseded by one that @Matt711 is working on this week.

jacobtomlinson:
Superseded

PR #611, opened 2022-11-15
Wait for LoadBalancer service readiness

hahrot:
The controller should not change the phase to Running until the LoadBalancer service is truly ready.

This requires the client to also check for actual value of the phase in the status.

### DIFF
--- a/dask_kubernetes/operator/controller/controller.py
+++ b/dask_kubernetes/operator/controller/controller.py
@@ -186,9 +186,19 @@ async def wait_for_service(api, service_name, namespace):
     """Block until service is available."""
     while True:
         try:
-            await api.read_namespaced_service(service_name, namespace)
-            break
+            service = await api.read_namespaced_service(service_name, namespace)
+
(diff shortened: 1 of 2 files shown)


jacobtomlinson (approved):
Thanks for raising this! I like this change and agree with the intent here. 

I have a concern that this extends the time spent in this loop to potentially a very long time. This function is called in `daskcluster_create_components` which should be short and snappy due to the way `kopf` is designed so we could run into edge cases where the controller is restarted while we are in this loop and I don't think things would end well.

I think the solution to that problem would be to refactor `daskcluster_create_components` into many functions which are sequentially triggered by events rather than us polling within loops. This would mean that `wait_for_service` is only used by `KubeCluster` and could maybe even move over there completely. That would also remove all of the imports that `KubeCluster` makes to the operator controller, a coupling I regret, so that's a nice benefit too.

I'm happy to merge this but I'll open a new issue to track resolving the underlying edge-case bug that this PR increases the chances of us running into.

PR #614, opened 2022-11-17
Handle cluster service status via the scheduler service status

jacobtomlinson:
Instead of waiting for the service within the cluster creation handler we handle the service status updates directly and only move the cluster into `Running` state if the service is up.

### DIFF
--- a/dask_kubernetes/operator/controller/tests/test_controller.py
+++ b/dask_kubernetes/operator/controller/tests/test_controller.py
@@ -181,7 +181,7 @@ async def test_simplecluster(k8s_cluster, kopf_runner, gen_cluster):
                     "jsonpath='{.items[0].metadata.annotations}'",
                 )[1:-1]
             )  # First and last char is a quote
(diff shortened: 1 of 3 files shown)


dask_kubernetes/operator/controller/tests/test_controller.py
```diff
@@ -193,7 +193,7 @@ async def test_simplecluster(k8s_cluster, kopf_runner, gen_cluster):
                     "jsonpath='{.items[0].metadata.annotations}'",
                 )[1:-1]
             )  # First and last char is a quote
-            assert service_annotations == _EXPECTED_ANNOTATIONS
+            assert _EXPECTED_ANNOTATIONS <= service_annotations
```
jacobtomlinson, line 196: Adding a handler to the service broke this test. Tweaking it to check that `_EXPECTED_ANNOTATIONS` are a subset of the annotations to avoid this kind of break in the future. cc @bstadlbauer

Issue #620, opened 2022-11-28
Dask Operator becomes unresponsive after ~1hr


baswelsh:
This one's a bit tricky.  I haven't been able to reproduce it in a kind or minikube cluster - it only happens in our hosted environments (i.e., AKS, EKS, etc.).  The dask operator deployment works great for a while and then becomes unresponsive.  I've seen it happen after being up for 1hr and 2hrs (pretty much on the nose).

The status of the pod is `running` but the logs are frozen and the operator is unresponsive to `KubeCluster` instantiation.  Has anyone seen this?  If not, I'd really appreciate any guidance on how to efficiently narrow in on the root cause (i.e., increasing log levels, inspecting heartbeats, operator health/status queries, etc.)

**Anything else we need to know?**:

Note that resetting the dask operator deployment via `kubectl rollout restart deployment dask-kubernetes-operator -n dask-operator` recovers the operator after about 60+s (which is how long it takes to terminate the pod once frozen).

Here are the final lines of the `dask-kubernetes-operator` pod logs:

```
[2022-11-28 00:45:46,745] kopf.objects         [DEBUG   ] [6556cd3b-4385-4d5d-98dd-af70b0ad8f80/sedaro-dask] Updating diff: (('change', ('status', 'phase'), 'Created', 'Running'),)
[2022-11-28 00:45:46,745] kopf.objects         [DEBUG   ] [6556cd3b-4385-4d5d-98dd-af70b0ad8f80/sedaro-dask] Patching with: <redacted>
[2022-11-28 00:45:46,754] kubernetes_asyncio.c [DEBUG   ] response body: <redacted>
[2022-11-28 00:45:46,764] kubernetes_asyncio.c [DEBUG   ] response body: <redacted>
[2022-11-28 00:45:46,764] kopf.objects         [INFO    ] [6556cd3b-4385-4d5d-98dd-af70b0ad8f80/sedaro-dask-default] Successfully adopted by sedaro-dask
[2022-11-28 00:45:46,784] kubernetes_asyncio.c [DEBUG   ] response body: <redacted>
[2022-11-28 00:45:46,801] kubernetes_asyncio.c [DEBUG   ] response body: <redacted>
[2022-11-28 00:45:46,805] kopf.objects         [INFO    ] [6556cd3b-4385-4d5d-98dd-af70b0ad8f80/sedaro-dask-default] Scaled worker group sedaro-dask-default up to 1 workers.
[...]

PR #621, opened 2022-11-28
Configure liveness probe for operator controller

jacobtomlinson:
xref #620 

Enabled the `kopf` liveness probe so that k8s can monitor controller health.


jacobtomlinson:
Lint failure fixed in #622

PR #624, opened 2022-11-29
set dask workergroup owner during creation

### LINKED ISSUE #623: dask worker group doesn't propagate labels into dask worker pod
**Bug details**

This issue is simple to reproduce. Simply create a dask cluster with labels in the metadata, i.e.

```
metadata:
  labels:
    test-label: "label-value"
```

We would expect this label to propagate into the dask worker group, dask scheduler, and dask worker pod. However these labels are only propagated into the dask worker group and the dask scheduler, **not** the dask worker pod.

**Explanation for bug**

We set the owner of the dask worker group during `daskworkergroup_create`, which happens after a worker group is created. This also kicks off `daskworkergroup_update` which is  where we create the worker pods and allow them to be adopted by the dask worker group. 

However in `daskworkergroup_update` when we set this pod/worker-group ownership, the labels aren't available yet, as the cluster/worker-group ownership isn't fully instantiated yet. 

Therefore the fix is to instantiate the cluster/worker-group ownership earlier so that the labels are available.

**Environment**:

- Dask version: 2022.11.1
- Python version: 3.8.5
- Operating System: Ubuntu 18.04
- Install method (conda, pip, source): pip

xWaita:
- Sets the dask workergroup owner during its creation instead of during `daskworkergroup_create` kopf call
- This allows labels to propagate correctly from the dask workergroup to the dask worker pod

closes #623

### DIFF
--- a/dask_kubernetes/operator/controller/controller.py
+++ b/dask_kubernetes/operator/controller/controller.py
@@ -1,21 +1,15 @@
+from contextlib import suppress
 from datetime import datetime
+from uuid import uuid4
 
 import aiohttp
-from contextlib import suppress
-
 import kopf
 import kubernetes_asyncio as kubernetes
 import pykube
-
-from uuid import uuid4
-
 from dask.compatibility import entry_points
-from distributed.core import rpc
-
 from dask_kubernetes.common.auth import ClusterAuth
-from dask_kubernetes.common.networking import (
-    get_scheduler_address,
-)
+from dask_kubernetes.common.networking import get_scheduler_address
+from distributed.core import rpc
 
 _ANNOTATION_NAMESPACES_TO_IGNORE = (
     "kopf.zalando.org",
@@ -397,9 +391,19 @@ async def get_desired_workers(scheduler_service_name, namespace, logger):
 @kopf.on.update("daskworkergroup.kubernetes.dask.org")
 async def daskworkergroup_update(spec, name, namespace, logger, **kwargs):
     async with kubernetes.client.api_client.ApiClient() as api_client:
-        api = kubernetes.client.CoreV1Api(api_client)
+        customobjectsapi = kubernetes.client.CustomObjectsApi(api_client)
+        corev1api = kubernetes.client.CoreV1Api(api_client)
 
-        workers = await api.list_namespaced_pod(
+        cluster = await customobjectsapi.get_namespaced_custom_object(
+            group="kubernetes.dask.org",
+            version="v1",
+            plural="daskclusters",
+            namespace=namespace,
+            name=spec["cluster"],
+        )
+        cluster_labels = cluster.get("metadata", {}).get("labels", {})
+
+        workers = await corev1api.list_namespaced_pod(
             namespace=namespace,
             label_selector=f"dask.org/workergroup-name={name}",
         )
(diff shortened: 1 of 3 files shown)


dask_kubernetes/operator/controller/controller.py
```diff
@@ -284,7 +279,6 @@ async def daskworkergroup_create(spec, name, namespace, logger, **kwargs):
             name=spec["cluster"],
         )
         new_spec = dict(spec)
-        kopf.adopt(new_spec, owner=cluster)
```
jacobtomlinson, line 287: We are doing this intentionally here so that new worker groups that get created manually are also adopted and cascade deleted when the cluster is deleted. Please restore this line.

dask_kubernetes/operator/controller/controller.py
```diff
@@ -260,6 +254,7 @@ async def daskcluster_create_components(spec, name, namespace, logger, patch, **
 
         worker_spec = spec.get("worker", {})
         data = build_default_worker_group_spec(name, worker_spec, annotations)
+        kopf.adopt(data)
```
jacobtomlinson, line 257: I removed this at some point because worker groups get adopted to their cluster after creation so that they are cascade deleted regardless of how they were created.

jacobtomlinson (commented):
Thanks for taking a stab at this. I'm afraid your changes only work for the default worker group and will break adoption for additional worker groups.

When a `DaskCluster` is created the controller creates the default `WorkerGroup`. But the user can create more `WorkerGroup` resources in Python or via `kubectl`. Those worker groups also need to be adopted by the cluster so that when the cluster is deleted the workers get cleaned up too.

I think instead we need to copy labels from the cluster resource onto the worker group when the worker group is created.

xWaita:
@jacobtomlinson thanks for the review. I've reverted the adopt change, and directly added the cluster's labels to each worker. How does the MR look now?

dask_kubernetes/operator/controller/tests/test_controller.py
```diff
@@ -227,6 +224,48 @@ async def test_simplecluster(k8s_cluster, kopf_runner, gen_cluster):
...
+                    "get",
+                    "daskworkergroups",
+                    "--selector=dask.org/component=workergroup",
+                    "-o",
+                    "jsonpath='{.items[0].metadata.labels}'",
+                )[1:-1]
+            )
+            for label in _EXPECTED_LABELS:
+                assert label in workergroup_labels
+                assert workergroup_labels[label] == _EXPECTED_LABELS[label]
```
jacobtomlinson, line 239: This can be simplified to
```suggestion
            assert _EXPECTED_LABELS.items() <= workergroup_labels.items()
```

jacobtomlinson (commented):
Looks great, one small nit.

dask_kubernetes/operator/controller/tests/test_controller.py
jacobtomlinson: Same for the others

xWaita: Thanks @jacobtomlinson , fixed up

xWaita: tests have just passed :)

xWaita:
Thanks @jacobtomlinson , what is the process for requesting a tagged release? It looks like I don't have perms to kick this off myself.

jacobtomlinson:
I just tagged `2022.12.0` :)

PR #627, opened 2022-11-29
Adds `pykube` as dependency

baswelsh:
Fixes https://github.com/dask/dask-kubernetes/issues/620#issuecomment-1330845244


baswelsh:
This is ready @jacobtomlinson 🎉

PR #630, opened 2022-12-04
Allow for the configuration of the scheduler dashboard forward port. 

### LINKED ISSUE #615: Ability to configure port for forwarding
Right now, the cluster scheduler's dashboard port is forwarded to a random system port. Ideally this port should be configurable for deployments where that makes sense. For example, I am running this in docker on windows and thus can't use host networking for the docker container and obviously can't forward every possible port from the container. 

I am willing to make the changes necessary, but I just wanted to see any feedback prior to submitting a PR.

patmagauran:
This adds the ability to configure the port for the scheduler dashboard. 
Fixes #615


dask_kubernetes/common/networking.py
```diff
@@ -76,12 +89,14 @@ def _random_free_port(low, high, retries=20):
 
 async def port_forward_service(service_name, namespace, remote_port, local_port=None):
     check_dependency("kubectl")
-    if not local_port:
+    if not local_port or _port_in_use(local_port):
```
jacobtomlinson, line 92: If the port is already in use we should raise a warning, otherwise the user may be confused why they've specified a port but it isn't using it.

jacobtomlinson (commented):
Generally this look great! Could I ask you to add a small test?

patmagauran:
> 

I added a test for the custom port as well as changing it so an exception is raised if the port is already in use. I figure that is better than printing a warning or randomly generating a port.

Issue #641, opened 2023-01-18
Add ability to control log level and other logging configs


jmif:
We've deployed the Dask operator via Helm and we're getting quite a few logs from the operator (especially when running auto scalers along with our Dask clusters), in production we don't need to see all of these logs.  I think it's coming from the `--verbose` line [here](https://github.com/dask/dask-kubernetes/blob/main/dask_kubernetes/operator/deployment/Dockerfile#L14).

Additionally, we're deployed on GKE and it GCP logging doesn't recognize the format of the output logs, which is causing everything to be marked as critical.  It'd be nice to be able to change the log output configuration as well.

Looking at the kopf docs, it looks like these are configurable via [command line flag](https://kopf.readthedocs.io/en/stable/configuration/#logging-formats-and-levels) or via [startup settings](https://kopf.readthedocs.io/en/stable/configuration/#startup-configuration), I don't see anything regarding alternative config methods (config file or env vars) though I have very little experience with kopf so I may be missing something.

Based on what I can find, I'd suggest either:
- modifying the helm template to support container args
- adding environment variables that then set config in the startup configuration

I'm happy to take a stab at a PR for this but would appreciate input on the approach before doing so.  If we think environment variables are best would love to hear thoughts on the environment variable naming.

Thanks!

jacobtomlinson:
This sounds like a great improvement. 

I did a little code spelunking into `kopf` and it definitely looks like CLI flags are the only way to configure the log level, format, etc and the startup settings are the only way to configure the k8s event API postings.

Making all of these configurable makes a ton of sense.

I think if we went down the environment variables road I would want to just use Dask's config system for consistency with other parts of the project. This would require a new Dask CLI which would load the config and then launch kopf in a subprocess with the appropriate flags depending on the config. That sounds like a bunch of extra work so I'm tempted to dismiss this solution.

Therefore perhaps the better solution would be to make these options configurable in the helm chart. Templating in the command line flags would be reasonably straight forward.

jmif:
Initial PR open!

PR #642, opened 2023-01-19
Add support for kopf CLI arguments in operator deployment

jmif:
Adds the ability to provide kopf CLI arguments via the Helm values.yaml file.  

Some callouts:
- I split the container `CMD` out into an `ENTRYPOINT` and `CMD` to make the `args` definition cleaner (allows consumers to just specify CLI args there and not worry about properly starting up container) 
- I kept the container default args the same for backwards compatibility and opted to override the default args via the helm chart
- I've updated the default helm deploy to remove the `--verbose` flag
- I've left the `--all-namespaces` flag as a `CMD` flag, so if users override this they'll need to include the `--all-namespaces` flag again.  I figure this provides the most flexibility (if we hard code it like we do liveness probe users can't change it because its a boolean flag).
- I tested both the docker container and the helm chart locally but haven't written any tests as I don't see any tests for this in the repo, let me know if I've missed anything here. I've outline what I've tested below.

**Tests:**
All tests were run with a local build of the included Dockerfile, run against minikube.

values.yml
```
kopfArgs: []
```

```
kopfArgs:
  - '--verbose'
  - --log-format
  - json
```

```
kopfArgs:
  - '--verbose'
  - --log-format
  - json
```

```
kopfArgs: [ '--verbose ']
```

Addresses https://github.com/dask/dask-kubernetes/issues/641

### DIFF
--- a/dask_kubernetes/operator/deployment/helm/dask-kubernetes-operator/values.yaml
+++ b/dask_kubernetes/operator/deployment/helm/dask-kubernetes-operator/values.yaml
@@ -42,3 +42,6 @@ nodeSelector: {}  # Node selector
 tolerations: []  # Tolerations
 
 affinity: {}  # Affinity
+
+kopfArgs: # Command line flags to pass to kopf on start up
+  - --all-namespaces
--- a/dask_kubernetes/operator/deployment/Dockerfile
+++ b/dask_kubernetes/operator/deployment/Dockerfile
@@ -11,4 +11,5 @@ WORKDIR /src/dask_kubernetes
 RUN pip install .
 
 # Start operator
-CMD kopf run -m dask_kubernetes.operator.controller --liveness=http://0.0.0.0:8080/healthz --verbose --all-namespaces
+ENTRYPOINT ["/usr/local/bin/kopf", "run", "-m", "dask_kubernetes.operator.controller"]
+CMD ["--liveness=http://0.0.0.0:8080/healthz", "--verbose", "--all-namespaces"]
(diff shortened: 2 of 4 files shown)


jmif:
Can someone help me understand the lint failure as well?  Looks like the linter wants arguments for the new values.yml field I added and it doesn't like what I've added:

```
 | `nodeSelector` | Node selector | `{}` |
 | `tolerations` | Tolerations | `[]` |
 | `affinity` | Affinity | `{}` |
-| `kopfArgs` | Command line flags to pass to kopf on start up  | `["--all-namespaces"]` |
+| `kopfArgs` |  | `["--all-namespaces"]` |
+
```

dask_kubernetes/operator/deployment/helm/dask-kubernetes-operator/values.yaml
```diff
@@ -42,3 +42,7 @@ nodeSelector: {}  # Node selector
 tolerations: []  # Tolerations
 
 affinity: {}  # Affinity
+
+# kopf command line arguments
+kopfArgs:
+  - --all-namespaces
```
jacobtomlinson, line 48: This *should* stop the linter from complaining. 

```suggestion
# Command line flags to pass to kopf on start up
kopfArgs:
  - --all-namespaces
```

jacobtomlinson (approved):
This looks great!

The linter uses frigate to auto-generate the table in the README from the comments in the `values.yaml`. It's generally less painful if you run pre-commit locally before pushing.

jmif:
Fixed!  Looks like Frigate wanted the comment on same line as the top level value key:

```
kopfArgs: # Command line flags to pass to kopf on start up
  - --all-namespaces
```

jacobtomlinson (approved):
Yup that makes sense. Thanks for fixing things up.

PR #643, opened 2023-01-21
Add services/status in the ClusterRole

john-jam:
This PR adds the missing `services/status` role in the `ClusterRole` resource to avoid errors like the following:
```
User "system:serviceaccount:dask:dask-kubernetes-operator" cannot patch resource "services/status" in API group ""
```


jacobtomlinson (approved):
Awesome thanks

PR #644, opened 2023-01-21
Separate the service account and the cluster role creation

john-jam:
This PR separates the service account and the cluster role creation with a new `rbac.create` option that creates the `ClusterRole` and the `ClusterRoleBinding` when set to `true`. It also define those resources in separate files.

This enables the possibility to create the service account outside of the Helm Chart and still create and bind the `ClusterRole` needed for the operator to work properly.


dask_kubernetes/operator/deployment/helm/dask-kubernetes-operator/README.md
```diff
@@ -18,6 +18,7 @@ The following table lists the configurable parameters of the Dask-kubernetes-ope
 | `nameOverride` | Override release name (not including random UUID) | `""` |
 | `fullnameOverride` | Override full release name | `""` |
 | `serviceAccount.create` | Create a service account for the operator to use | `true` |
+| `serviceAccount.createClusterRole` | Create a cluster role needed by the operator and bind it to the service account | `true` |
```
jacobtomlinson, line 21: I'm not sure about the namespacing on the config options here. Maybe there is something we could do to make them appear more consistent?

```suggestion
| `serviceAccount.createServiceAccount` | Create a service account for the operator to use | `true` |
| `serviceAccount.createClusterRole` | Create a cluster role needed by the operator and bind it to the service account | `true` |
```

or

```suggestion
| `serviceAccount.create` | Create a service account for the operator to use | `true` |
| `clusterRole.create` | Create a cluster role needed by the operator and bind it to the service account | `true` |
```

I'm not super confident in either of those suggestions. What do you think?

john-jam: I prefer the second option (having a separated `clusterRole` config option) because the `serviceAccount.create` option is usually common in helm charts so we better keep this one imho.
While we're at it, should we move the `ClusterRole` and `ClusterRoleBinding` into their own files `clusterrole.yml` and `clusterrolebinding.yml`?

jacobtomlinson: Ok sounds good, being consistent with user expectations is important. I'd be interested to see how other charts with Cluster Roles handles things, maybe there is a convention there too?

Happy to split the template files down further if you like. I think that's just a matter of preference though.

john-jam: According to the following items, I updated the PR and its description:
- [Structure of the templates folder](https://helm.sh/docs/chart_best_practices/templates/#structure-of-templates): _Each resource definition should be in its own template file._
- [Service Account and RBAC configuration](https://helm.sh/docs/chart_best_practices/rbac/#yaml-configuration): _RBAC and ServiceAccount configuration should happen under separate keys._
- Popular repositories using it:
  - [kubernetes/ingress-nginx](https://github.com/kubernetes/ingress-nginx/tree/main/charts/ingress-nginx)
  - [prometheus-community/helm-charts](https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus)
  - [cert-manager/cert-manager](https://github.com/cert-manager/cert-manager/tree/master/deploy/charts/cert-manager) (they use a single `*-rbac.yaml` if you prefer)

jacobtomlinson (approved):
Awesome this looks great.

PR #648, opened 2023-01-24
Add cool down period for autoscaling to reduce thrashing

jmif:
This PR is part of a two part set (#649 is the companion PR) that smooths out the autoscaling logic to reduce thrashing and hopefully smooth out autoscaling actions.   I've duplicated some of the context in both PRs so in case there are different reviewers. 

We've been using the dask operator for some long running workloads and have noticed that the autoscaling logic can result in thrashing in some of our workloads.  When this happened our we either experienced:
- A cluster left in an inconsistent state (no workers but work to be done and no actions to scale up being performed)
- Workers being spun up only to spin down, resulting in excessive data transfers and loss of work when the data could not be transferred within the pod termination grace period.

The second PR fixes a race condition in the worker group auto scaling process that should ensure that rapid updates to a worker group's scale will result in a consistent state for the worker group (no excess or missing workers).  With these two PRs combined, we've seen much smoother autoscaling and haven't run into missing workers (we've been running this in production for a couple weeks).

Some call outs:
- I've kept these two PRs separate as they can operate fully independently and are easier to grok independently but I'm happy to combine them if you'd prefer.
- I'm not sure where to store the "cooldown until" timestamp.  I'm currently using an annotation but I'm not sure what the best practice / recommended way to store this internal state is.  Happy to update this based on your feedback.  I've considered:
  - Annotations - went with this as it was the simplest way to get a PR open for feedback
  - Managed fields - not sure of operators are can (or are intended to) control these
  - A dedicated field in the CRD - seems like these are meant to be set by consumers and not used for internal state
  - Some kind of "replicas last updated at" field on the worker group - basically same thoughts as the above regarding where to store in the worker group object but this has the added benefit of being potentially useful state for other actors in the operator
- There are no tests yet, I'm opening this now to get feedback on the implementation / structure before adding tests.
- The autoscaling smoothing logic is pretty simple right now: it maintains a cooldown for scaling up and a cooldown for scaling down, the cooldown for scaling up is shorter, allowing for more aggressive scale ups and more conservative scale downs.
- I've added a max cool down rate of 25% of cluster size.  In one of our runs the scheduler scaled down from 6 to 1 instances right as the work was wrapping up and we ended up losing 75% of our computed work in the shutdown.  In addition to a cooldown, this would make sure that clusters scale down gradually to allow data to be rebalanced / work to be completed.
- These autoscaling patterns aren't perfect but I've seen them used successfully many times and I think they'll be a solid improvement on the existing autoscaling system.
- Added some other bug fixes and improvements that resolves errors we were seeing running the autoscaler in prod (ie missing scheduler or slow to spin up scheduler was throwing exceptions)
- Autoscaler stands down when it doesn't need to update replicas (resolves TODO)
- I bumped the autoscaler timer interval to 30 seconds as I don't think 5 seconds is necessary with the cooldowns

### DIFF
--- a/dask_kubernetes/operator/__init__.py
+++ b/dask_kubernetes/operator/__init__.py
@@ -1,5 +1,3 @@
-from .controller import *
-
 from .kubecluster import (
     KubeCluster,
     make_cluster_spec,
--- a/dask_kubernetes/operator/__init__.py
+++ b/dask_kubernetes/operator/__init__.py
@@ -1,3 +1,5 @@
+from .controller import *
+
 from .kubecluster import (
     KubeCluster,
     make_cluster_spec,
--- a/dask_kubernetes/operator/controller/controller.py
+++ b/dask_kubernetes/operator/controller/controller.py
@@ -1,3 +1,4 @@
+import time
 from contextlib import suppress
 from datetime import datetime
 from uuid import uuid4
@@ -7,6 +8,8 @@ import kopf
 import kubernetes_asyncio as kubernetes
 import pykube
 from dask.compatibility import entry_points
+from kubernetes_asyncio.client import ApiException
+
 from dask_kubernetes.common.auth import ClusterAuth
 from dask_kubernetes.common.networking import get_scheduler_address
 from distributed.core import rpc
@@ -18,6 +21,7 @@ _ANNOTATION_NAMESPACES_TO_IGNORE = (
 
 KUBERNETES_DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
 
+DASK_AUTOSCALER_COOLDOWN_UNTIL_ANNOTATION = "kubernetes.dask.org/cooldown-until"
 
 # Load operator plugins from other packages
 PLUGINS = []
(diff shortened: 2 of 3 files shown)


dask_kubernetes/operator/controller/controller.py
```diff
@@ -626,35 +630,121 @@ async def daskautoscaler_create(spec, name, namespace, logger, **kwargs):
...
         )
-        await customobjectsapi.patch_namespaced_custom_object_scale(
+
+        autoscaler_resource = await customobjectsapi.get_namespaced_custom_object(
+            group="kubernetes.dask.org",
+            version="v1",
+            plural="daskautoscalers",
+            namespace=namespace,
+            name=name,
+        )
```
jacobtomlinson, line 664: It looks like the only attribute being accessed on this object is `metadata`. This is already available in `kwargs["metadata"]`. I recommend updating the function signature to include a `metadata` kwarg and access that directly.

jacobtomlinson (commented):
Thanks for taking the time to dig into this. I've been bashing my head against this issue for a while. 

Adding some cooldowns seems totally reasonable, but I am worried they may be masking the underlying problem. The cooldowns added in this PR are also extremely long and may make clusters feel very unresponsive.

In the classic adaptive functionality the interval was 5 seconds and the scale down cooldown was 15 seconds. I would prefer things to be similar here. If the bugs are still presenting themselves at that speed then there is something deeper to investigate.

jmif:
I definitely agree that this is a bandaid and not an actual solution to the problem, I do think they’re a reasonable addition to at least smooth over some of the auto scaling issues even with that in mind if for no other reason that it may take the cluster some time to rebalance and adjust after scaling changes so it’s helpful to wait a bit before measuring again.

I also agree this makes the cluster feel less responsive, particularly on scale up.  I have a few suggestions based on this:

- We could significantly reduce the cool down period for scale ups and basically ensure that whenever the cluster wants to scale up, it waits a small number of seconds from the last auto scaling action
- I think we should keep the cool down period high to ensure that we don’t lose work.  I did some more digging into this and I think we were losing work because the cluster scaled down to 1 worker, requiring all closing workers to transfer their data to the remaining active worker.  Not all workers were able to do this at once and some were killed due to the pod termination period. 
- We could make the cool down periods configurable with sensible defaults

jmif:
Also - what are your thoughts on using the annotation to store the cooldown period?

jmif:
Update on this - after running for a couple weeks its clear the current cooldown restrictions are annoyingly slow and I think we've fixed the actual issue in #649.  I've updated this to match the classic functionality!

dask_kubernetes/operator/controller/controller.py
jacobtomlinson: I notice this still hasn't been addressed. I'd like to remove this unnecessary HTTP call if possible. Was there a reason why you couldn't use `kwargs["metadata"]` here?

jacobtomlinson (commented):
>  I've updated this to match the classic functionality!

Ok great. I'm excited about the progress in #649. 

> what are your thoughts on using the annotation to store the cooldown period?

I'm okay with adding annotations to allow the configuration of cooldowns.

I've nudged the comment that's still open from the first review. Other than that I think we are good to merge here.

jacobtomlinson:
Thanks! Looks like CI was hanging for some reason. Giving it a kick but if it times out again we may need to see what is going on.

dask_kubernetes/operator/controller/controller.py
jmif: Update on this @jacobtomlinson, it looks like we're getting `TypeError: daskautoscaler_adapt() missing 1 required positional argument: 'metadata'` with this setup.   I've double check that referencing correctly and it looks right to me so perhaps timers don't get the metadata of the resource?  May explain why I opted to do it this way in the first place, but can't remember :).

jacobtomlinson (commented):
This is looking great and CI seems happy now (there is a known intermittent hang that is unrelated to this PR).

jacobtomlinson:
Ok I removed the import and CI is passing so I'm going to merge this in.

dask_kubernetes/operator/__init__.py
```diff
@@ -1,3 +1,5 @@
+from .controller import *
```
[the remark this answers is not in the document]
bstadlbauer: @jacobtomlinson I think this might be related to https://github.com/dask/dask-kubernetes/pull/649#issuecomment-1439625841 as otherwise tests won't run on M1 for some strange reason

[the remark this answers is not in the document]
jacobtomlinson: Fair enough. Would you mind opening a separate issue to discuss M1 development though?

jmif:
Awesome!  Apologies if the aarch imports slipped in :)

Issue #633, opened 2022-12-11
Adaptive `KubeCluster` with Operator losing work and unexpectedly scaling up and down.


tasansal:
Hi everyone! Thanks for all the effort on the Kubernetes operator for Dask. It is definitely the way to go. However, we ran into a problem that negatively affects our workload. The same minimally reproducible code runs fine with the classic `KubeCluster` but fails with the new one. The culprit seems to be the `DaskAutoScaler`. More details are below.

**Describe the issue**

We are deploying the `Dask Kubernetes Operator` on our GKE cluster and turning on auto-scaling. However, with the simple example below, the auto scaler is having a very rough time. The pathology of issues we see:

Preface: I will use `adapt_min`, and `adapt_max` to denote min/max values for the adaptive scaler. 

1. Scheduler is idle and has `adapt_min` workers.
2. Workflow gets submitted.
3. New pods start spinning up, **sometimes existing workers that are ready to go die**
4. New pods start getting added to the scheduler. Sometimes it reaches `adapt_max`, sometimes it doesn't. Even with the same code.
5. Most of the time, the number of workers varies between `adapt_min` and `adapt_max`, even if the workload takes more than 30 seconds. In the middle of the progress, it can kill workers and restart them.
6. Sometimes, the compute finishes and takes much longer because the adaptive scaler killed and re-created workers with data.
7. Sometimes it doesn't finish with the error: Scheduler wited for `Key(...)`, but it wasn't available; reassigning.
8. Once workflow dies, it usually kills more workers than it should, and we end up with less than `adapt_min`.
9. Sometimes takes a long time to recover from bullet 8 and go back to `adapt_min` or even to scale again.

**Minimal Complete Verifiable Example**

Assuming `kubectl` is set up, and the operator is installed.

Creating the Adaptive Dask Cluster and connect from Client via:

```python
from dask_kubernetes.operator import KubeCluster
from distributed import Client

cluster = KubeCluster(
    name="simple", 
    image="ghcr.io/dask/dask:2022.12.0-py3.10",
    n_workers=4,
    worker_command=["dask", "worker", "--nthreads", "1"],
)

cluster.adapt(4, 30)

client = Client(cluster)
```

Workload; generate a random 3D array and take a slice from it.
This mimicks our workload of reading a slice of a large `Zarr` dataset:
```python
import dask.array as da

rand_vol = da.random.normal(
    size=(2048, 8192, 4096),
    chunks=(128, 512, 256),
).astype('float16')

[...]

PR #649, opened 2023-01-24
Resolve race condition on Dask worker group replica updates

jmif:
This PR is part of a two part set (#648 is the companion PR) that smooths out autoscaling logic to reduce thrashing and ensure a consistent state after scaling completed.  I've duplicated some of the context in both PRs in case there are separate reviewers.

We've been using the dask operator for some long running workloads and have noticed that the autoscaling logic can result in thrashing in some of our workloads. When this happened our we either experienced:

A cluster left in an inconsistent state (no workers but work to be done and no actions to scale up being performed)
Workers being spun up only to spin down, resulting in excessive data transfers and loss of work when the data could not be transferred within the pod termination grace period.

This PR fixes a race condition in the worker group auto scaling process that should ensure that rapid updates to a worker group's scale will result in a consistent state for the worker group (no excess or missing workers).  Combined with the autoscaling PR, we've seen much smoother autoscaling and haven't run into missing workers (we've been running this in production for a couple weeks).

We noticed that autoscaled clusters would often get stuck in the following state:
- Work available to be completed in the scheduler
- The default worker group set with a replicas value > 0
- No worker pods active and no worker pods being created

Looking at the code that manages the worker replica count, I noticed that it had a high probability for race conditions when multiple updates to the replica count were performed in quick succession.  Each event callback executes in parallel and the even callbacks are async, so we can get into a situation where two event callbacks are running to modify the number of replicas in a worker group.  The existing implementation makes async API calls to read state and then uses that to decide what scaling actions to take without locking, which was causing a race condition in deciding how many replicas to create and leaving us in a state with no replicas.  I haven't enumerated the specific steps to illustrate the race condition as I opted to try applying a lock on the worker group to see if it resolved the issue we were seeing.  We've been running this in production for a couple weeks and we haven't seen the issue reappear. 

Some call outs:
- I've kept these two PRs separate as they can operate fully independently and are easier to grok independently but I'm happy to combine them if you'd prefer.
- There are no tests yet, I'm opening this now to get feedback on the implementation / structure before adding tests.
- I've moved the worker group replica logic to a field event to focus the locking code to a single method (allow other code to be run in parallel related to the worker group if necessary in the future)
- After implementing this I noticed that worker pods were not being cleaned up on cluster shutdown.  Prior to these changes, worker groups were getting a replica count set to 0 when the cluster was deleted, which resulted in the worker pods being terminated and everything being cleaned up.    I haven't been able to identify the code path that was causing cleanup originally, so I'm not sure why these changes stopped the worker pods from happening.  However, I was able to resolve by having the worker group adopt the pods.  The worker group itself is owned by the cluster, so this results in a cascading delete.  Calling this out in case there is some assumption or flow I've broken here.


dask_kubernetes/operator/controller/controller.py
```diff
@@ -392,67 +403,92 @@ async def get_desired_workers(scheduler_service_name, namespace, logger):
...
+                for _ in range(workers_needed):
+                    data = build_worker_pod_spec(
+                        worker_group_name=name,
+                        namespace=namespace,
+                        cluster_name=cluster_name,
+                        uuid=uuid4().hex[:10],
+                        spec=spec["worker"]["spec"],
+                        annotations=annotations,
+                    )
+                    kopf.adopt(data, owner=body)
```
jacobtomlinson, line 459: FYI this is the line that causes pods tobe cleaned up. The `Pod` gets adopted by the `DaskWorkerGroup`, which in turn has been adopted by the `DaskCluster`. When a k8s resource is deleted it will cascade delete the child resources.

If this isn't happening that's an interesting but separate problem.

dask_kubernetes/operator/controller/controller.py
```diff
@@ -392,67 +403,92 @@ async def get_desired_workers(scheduler_service_name, namespace, logger):
         ) from e
 
 
-@kopf.on.update("daskworkergroup.kubernetes.dask.org")
-async def daskworkergroup_update(spec, name, namespace, logger, **kwargs):
-    async with kubernetes.client.api_client.ApiClient() as api_client:
-        customobjectsapi = kubernetes.client.CustomObjectsApi(api_client)
-        corev1api = kubernetes.client.CoreV1Api(api_client)
+worker_group_scale_locks = defaultdict(lambda: asyncio.Lock())
```
jacobtomlinson, line 406: Adding a lock like this makes a lot of sense. However, this does prevent the controller from scaling. What happens if there are two or more instances of the controller? They will have different locks.

We may need to think of some other way to ensure that only one scale call can be running at a time.

dask_kubernetes/operator/controller/controller.py
```diff
@@ -308,8 +313,14 @@ async def daskworkergroup_create(spec, name, namespace, logger, **kwargs):
         )
         logger.info(f"Successfully adopted by {spec['cluster']}")
 
-    await daskworkergroup_update(
-        spec=spec, name=name, namespace=namespace, logger=logger, **kwargs
+    del kwargs["new"]
```
jacobtomlinson, line 316: I'm curious what this does

jacobtomlinson (commented):
I'm really excited to see this. These race conditions are the "underlying problems" I was referring to in #648. 

I'm excited to hear this is running nicely for you in production, that gives me hope that we can get to the end of this bug.

I have a couple of concerns about the overall design here, particularly because the controller is tracking state of the locks which isn't how `kopf` is designed to be used. `kopf` supports HA so if you scale to multiple controllers this fix wont work.

jmif:
Yeah I had a feeling tracking the locks internally wasn't the kopf way, it certainly felt out of place when I was writing.  I've just done some more digging into this and I wasn't able to find any information on HA kopf deployments, can you point me to what you're referencing?  I did notice a couple other things while digging in:

- I came across [this example](https://github.com/nolar/kopf/blob/5c6f68c9dfd80a23dc54bc50a951c3c2535476f3/examples/13-hooks/example.py) which seems to be using a similar patter to us, so at least in a non HA setup we don't seem to be too far off?
- I was reading the [peering docs](https://github.com/nolar/kopf/blob/5c6f68c9dfd80a23dc54bc50a951c3c2535476f3/examples/13-hooks/example.py) and noticed in the section on priority they discuss ensuring two operators aren't activate at the same time but instead ensure that one is on standby ready to pick up if the other drops.  I'm wondering if this is the HA setup you were referencing?
- I noticed [this line](https://github.com/nolar/kopf/blob/5c6f68c9dfd80a23dc54bc50a951c3c2535476f3/examples/13-hooks/example.py) in the peering docs: `Usually, one and only one operator instance should be deployed for the resource. If that operator’s pod dies, the handling of the resource of this type will stop until the operator’s pod is restarted (and if restarted at all).` Which reads to me as if they intend only one operator instance to be running for a given resource.

Given the above, I'm wondering if our current setup is done in the way kopf intends?  I'll defer to you here as I'm brand new to k8s operators and kopf but calling out for discussion!

dask_kubernetes/operator/controller/controller.py
jmif: (this is a response to both this comment and the HA portion of your top level review comment)

Yeah I had a feeling tracking the locks internally wasn't the kopf way, it certainly felt out of place when I was writing. I've just done some more digging into this and I wasn't able to find any information on HA kopf deployments, can you point me to what you're referencing? I did notice a couple other things while digging in:

- I came across [this example](https://github.com/nolar/kopf/blob/5c6f68c9dfd80a23dc54bc50a951c3c2535476f3/examples/13-hooks/example.py) which seems to be using a similar patter to us, so at least in a non HA setup we don't seem to be too far off?
- I was reading the [peering docs](https://github.com/nolar/kopf/blob/5c6f68c9dfd80a23dc54bc50a951c3c2535476f3/examples/13-hooks/example.py) and noticed in the section on priority they discuss ensuring two operators aren't activate at the same time but instead ensure that one is on standby ready to pick up if the other drops. I'm wondering if this is the HA setup you were referencing?
- I noticed [this line](https://github.com/nolar/kopf/blob/5c6f68c9dfd80a23dc54bc50a951c3c2535476f3/examples/13-hooks/example.py) in the peering docs: Usually, one and only one operator instance should be deployed for the resource. If that operator’s pod dies, the handling of the resource of this type will stop until the operator’s pod is restarted (and if restarted at all). Which reads to me as if they intend only one operator instance to be running for a given resource.

Given the above, I'm wondering if our current setup is done in the way kopf intends? I'll defer to you here as I'm brand new to k8s operators and kopf but calling out for discussion!

dask_kubernetes/operator/controller/controller.py
jmif: Yeah it wasn't happening until I added the `owner=body` argument, which explicitly set the owner of the pod to the worker group.  I noticed the issue after I introduced this change and I haven't noticed issues in the current release with this, so its either that something else is causing the pods to be cleaned up (I traced through the flow and I couldn't find anything that would be cleaning pods up) or more likely `adopt` behaves differently when used in the field update callback?

jmif:
Updated with another fix for #633 (see comments for details)

jacobtomlinson:
Thanks @jmif. Apologies for being slow to respond here. I'll aim to get this reviewed and merged this week.

dask_kubernetes/operator/controller/controller.py
jacobtomlinson: > or more likely adopt behaves differently when used in the field update callback

Yeah you're probably right, although I'm a little surprised. Explicitly setting the owner seems like a great solution though.

dask_kubernetes/operator/controller/controller.py
jacobtomlinson: Thanks for the response. My assumption was that multiple instances of the controller would divide up events, but it looks like that assumption was wrong and the behaviour is more of an active-backup model which works in our favour here.

My worry is still valid because if it fails over to the backup controller, the locks will be lost. Then if it fails back the other way there will be old and potentially out-of-date locks.

There is also an assumption that the controller can be safely restarted at any time. So it shouldn't really be holding things like locks in memory.

I think my suggestion would be to use a Kubernetes resource to manage this state. Looking at [these docs](https://kopf.readthedocs.io/en/stable/continuity/) it seems the peering code uses annotations on a `KopfPeering` resource to track state. Given that we are tracking locks for scaling a `DaskWorkerGroup` perhaps we should be storing the lock in an annotation on that object?

dask_kubernetes/operator/controller/controller.py
```diff
@@ -392,67 +411,92 @@ async def get_desired_workers(scheduler_service_name, namespace, logger):
...
-                await corev1api.create_namespaced_pod(
+@kopf.on.field("daskworkergroup.kubernetes.dask.org", field="spec.worker.replicas")
+async def daskworkergroup_replica_update(
+    name, namespace, meta, spec, new, body, logger, **kwargs
+):
+    cluster_name = spec["cluster"]
+
+    # Replica updates can come in quick succession and the changes must be applied atomically to ensure
+    # the number of workers ends in the correct state
+    async with worker_group_scale_locks[f"{namespace}/{name}"]:
```
jacobtomlinson, line 425: I wonder if we should generate some UUID constant at the top of the file. Then to take the lock we set an annotation on the `DaskWorkerGroup` like `kubernetes.dask.org/scale-in-progress` and set the value to that UUID.

Then as part of trying to take the lock we check for that annotation. If it exists and the UUID matches we know the current controller is busy scaling in another coroutine, so we wait. If the UUID doesn't match we can assume the controller was restarted or failed over and the in-progress scale has been aborted. So we can update the annotation with our UUID and take the lock.

Then at the end, we delete the annotation again to release the lock.

This functionality could be factored into a small async context manager so that the code here stays basically the same.

jacobtomlinson (commented):
Thanks for really digging into what was going on here. 

I added a comment with a suggestion on how we could implement a lock in a "kopf-way" using an annotation.

dask_kubernetes/operator/controller/controller.py
jmif: Yup I think a kopf based annotation lock would work well.  I'll include some thoughts on how we should implement it here and I've also included some additional thoughts on the pure memory lock vs annotation lock.  Once we decide I'll implement and update!

Starting with thoughts on the method:
- The key question is what do we need to ensure is locked across function executions and processes.  In the case of `daskworkergroup_replica_update` we need to ensure that the (inclusive) block between calculating `current_workers` and making the API calls to k8s is executed serially.  We need to make sure that between the time we read the state of the cluster and apply changes, nothing else comes along and tries to read the current state.  
- I'm assuming that we can rely on kopf to ensure that only one operator is running at once, so a new operator wont start operating until the others are down.  If that isn't true then I think we need to go with the annotation approach.
- If the above assumption is true, then I think we can be sure that when the second operator starts running, the first one has stopped and thus the lock will be either cleanly or forcefully released by the first one.
  - If the lock is cleanly released then the scaling process completed (or at least finished in a steady state) and the next operator is free to pick it up
  - If the lock is forcefully released, the scaling process may be left in a intermittent state (half the pods terminated, half the pods created, etc).  This intermittent state is captured in the state of the pods and the operator is required to read the state of the pods (terminating, creating, etc) when it is determining what pods to create / destroy.  The operator already does this to make itself resilient to pod lifecycles states / changes. 
  - So in the case of this method I think the critical parts that must be executed serially are in fact executed serially, I can't think of a way to get the critical block to execute in parallel

Thoughts on how we can setup the annotation based approach:
- Use a UID for the operator instance (pull from kopf or generate one ourselve)
- Use two annotations (`owner_uid` and `heartbeat_timestamp`)
- To acquire the lock (following a similar pattern to distributed locking with DynamoDB / Redis):
  - Check if a lock exists
  - If a lock exists, wait for the lock to be released by looping to check, if the lock heartbeat goes stale for some period of time (ie the last time the `heartbeat_timestamp` is not updated for some number of minutes) then forcefully acquire the lock (this prevents a deadlock from a force kill of the operator)
  - Once the lock does not exist or we decide to force acquire, set the annotation of owner and timestamp and start an async timer that updates the heartbeat annotation every 10 seconds.
    - This needs to be done in an atomic "get / set" method  which essentially asserts that nobody has written the lock in parallel.  It ensures that if two processes are trying to acquire the same lock, only one wins.  In Dynamo I believe this is done with a condition expression that atomically ensures no lock record exists before writing the lock.  If another parallel write wins out, then the condition expression will fail the write and the lock won't be acquired.  I did some digging into the k8s API and it looks like if we include a resource version in the metadata and we use a replace call instead of an update (patch), the API server will reject the change if something else has written.
- To release the lock, delete the annotations
- Manage this with an async context manager as suggested

I think the timestamp / force acquire approach is necessary to ensure we don't step on another operator that is shutting down / finishing work before releasing the lock.  Without the timestamp, the new operator would need to forcefully acquire the lock when it starts up, which would mean we have a situation where the first operator could still be running thinking it has the lock.   As mentioned above, if we can rely on kopf to ensure that the new operator's callbacks don't run until the old one is shut down then I think this isn't necessary.  If we can't rely on kopf and forcefully acquire without the timestamp, there's a short time where two scaling processes could be executing together for the same worker group.  If we can rely on the operator we could still make an argument for future proofing things to account for logic that needs to be locked across processes, but I'm not sure the ROI on it is there.

Let me know what you think and I'll update!

dask_kubernetes/operator/controller/controller.py
jmif: This generally sounds good to me - I've responded to this in the other comment as both comments are related.

dask_kubernetes/operator/controller/controller.py
jacobtomlinson: > As mentioned above, if we can rely on kopf to ensure that the new operator's callbacks don't run until the old one is shut down then I think this isn't necessary.

My reading of the kopf peering docs makes me think we could rely on this.

I'll leave it to you how you want to implement it. I would be happy with either approach. If we agree that we can rely on kopf my advice would be to keep it as simple as possible.

jmif: I've given them another read and I'm in agreement we can rely on kopf so I suggest we stick with the current (simpler) implementation 👍

jmif:
Looks like CI is failing due to changes on this branch - I'm trying to get a local repro so I can fix but tests are failing for me I think because I'm on an M1 (basing this off note [here](https://kubernetes.dask.org/en/latest/testing.html)).  Any tips for getting this running short of spinning up a linux env?  I'm not seeing [_wait_for_controller](https://github.com/jmif/dask-kubernetes/blob/a24a3f9284043248e94816432abb9ece8e9be746/dask_kubernetes/operator/kubecluster/kubecluster.py#L377) get a running event (though it's getting and `ADDED` event) and the tests that rely on the cluster are timing out before the tests get started, so I'm assuming this is what is referenced in the testing docs page.

jacobtomlinson:
Sorry I don't really have any advice for running this on an M1.

bstadlbauer:
@jmif I've originally put in the note about the inability to run things on M1. Let me see whether this has changed since - will give this is try today

bstadlbauer:
@jmif I tried running tests and things definitely got farther than they did when I first tried this. I am seeing a few failed tests though (although the pods look alright in k8s). What test in particular are you interested in running?

jmif:
Thanks for taking a look!  Yeah I think we have a couple tests failing after these changes: `dask_kubernetes/operator/kubecluster/tests/test_kubecluster.py::test_adapt RERUN` is failing in #648 (happy to move that discussion there) and `dask_kubernetes/operator/kubecluster/tests/test_kubecluster.py::test_custom_spec RERUN [100%]` is failing here.

Are you able to get the tests running / succeeding on M1 on `main`?

bstadlbauer:
@jmif I've got the `dask_kubernetes/operator/kubecluster/tests/test_kubecluster.py::test_custom_spec` to work by adding `from .controller import *` to `dask_kubernetes/operator/__init__.py` as it seems like the `kopf_runner` does not actually start the controller on aarch. 

Currently working on a PR to get all tests to run locally.

jmif:
Awesome!  I added that and I'm able to get the tests progressing.  I've found the call that is hanging so now looking into why.

bstadlbauer:
Great! 
Sorry for not being super responsive right now, there's a lot going on everywhere. I couldn't get all test to run and haven't had time to look closely into why they're failing (IIRC I had about 4 or so failing)

jacobtomlinson:
I've pulled in from main and resolved the conflicts so let's see how CI goes now.

jacobtomlinson:
> @jmif I've got the `dask_kubernetes/operator/kubecluster/tests/test_kubecluster.py::test_custom_spec` to work by adding `from .controller import *` to `dask_kubernetes/operator/__init__.py` as it seems like the `kopf_runner` does not actually start the controller on aarch.
> 
> Currently working on a PR to get all tests to run locally.

Could you open any aarch compat stuff in a new PR? I'd like to review it separately from this PR.

jacobtomlinson:
It seems that CI is hanging on teardown pretty consistently for this PR now. I'm starting to pull out any orthogonal chunks that I can into separate PRs like #688 to try and reduce the scope of this PR to hopefully figure out what is causing the hang.

dask_kubernetes/operator/controller/controller.py
```diff
@@ -395,67 +412,92 @@ async def get_desired_workers(scheduler_service_name, namespace, logger):
...
+                    await corev1api.delete_namespaced_pod(
+                        name=wid,
+                        namespace=namespace,
+                    )
+                logger.info(
+                    f"Scaled worker group {name} down to {desired_workers} workers."
+                )
+
+
+@kopf.on.delete("daskworkergroup.kubernetes.dask.org")
```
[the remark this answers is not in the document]
jacobtomlinson: I'm going to try setting `optional=True` on the handler, which by my interpretation of [the docs](https://kopf.readthedocs.io/en/stable/handlers/#state-changing-handlers) means that Kubernetes will not block deletion of the resource on the handler completing, but the handler will still trigger.

jacobtomlinson:
Yay CI seems happy. Let's get this merged!

jmif:
Amazing!  Thank you for picking this up, been slowly trying to figure it out over the past few weeks when I've had a few spare minutes but had no luck.  Excited that it's merged!

jacobtomlinson:
Thank you for all your efforts on this! This was an unpleasant issue to dig through so I really appreciate you plowing forward with it.

PR #660, opened 2023-02-27
Add an asyncio shim for pykube-ng

### LINKED ISSUE #657: Noisy warnings on cluster creation
Whenever you create a cluster there are some noise warnings emitted. They aren't a problem but they are annoying.

```python
In [1]: from dask_kubernetes.operator import KubeCluster
   ...: cluster = KubeCluster(name="test", n_workers=1)
ERROR:asyncio:Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x7ff61fdf5340>
ERROR:asyncio:Unclosed connection
client_connection: Connection<ConnectionKey(host='127.0.0.1', port=46839, is_ssl=True, ssl=None, proxy=None, proxy_auth=None, proxy_headers_hash=None)>
```

I think they are coming from `kubernetes_asyncio` and the unclosed session is out of our control. So we probably need to find a way to suppress them rather than resolve them.

jacobtomlinson:
**TLDR**
- Add asyncio shim for `pykube-ng` because it is really nice to use but uses requests under the hood.
- Update the controller's usage of `pykube` to use the shim.
- Update `KubeCluster._wait_for_controller` to use this shim which fixes #657.

Over time I'd really like to replace all of our code that uses the swagger generated `kubernetes-asyncio` library with `pykube` to improve readability  and maintainability.

However, `pykube` is built on `requests` and doesn't natively support `asyncio`. To make it a little easier to work with this PR adds a `dask_kubernetes.aiopykube` submodule which provides an async shim that has the same API as `pykube` and runs all IO calls in a threadpool executor.

The shim also includes a `dask_kubernetes.aiopykube.dask` submodule with custom objects to represent [Dask custom resources](https://kubernetes.dask.org/en/latest/operator_resources.html).

There are a couple of design decisions in `pykube` which have limited how much of it can be converted to `asyncio`.
There are a few properties, such as `pukube.query.Query.query_cache`, that make HTTP calls. Using any kind of IO in a property or dunder like `__len__` is generally frowned upon and therefore there is no `asyncio` way to make or call a property asynchronously.

Instead of changing the API I've set these to raise `NotImplementedError` with useful exceptions that suggest an alternative way to achieve the same thing.

In theory we could pull the implementation out of `dask-kubernetes` into another dependency so that folks could reuse it. If there is demand from the community to do this we could definitely consider it but for now I want to keep complexity low and just add it as a submodule here. If we were to ever spin this shim out into a separate package we would probably advocate for a deeper rewrite using `aiohttp` instead of running calls that use `requests` in a threadpool executor.


jacobtomlinson:
I'm going to go ahead and merge this. If anyone has any feedback we can iterate in follow up PRs.

PR #666, opened 2023-03-07
Explore implementing a pure Python port forward

### LINKED ISSUE #663: Port forwards can timeout or close and not get recreated
When accessing a Dask cluster from outside of Kubernetes port forwards for the comm and dashboard will be set up automatically in a subprocess.

https://github.com/dask/dask-kubernetes/blob/6d54d275b706407d7c3037344bba534e80b0ce8e/dask_kubernetes/common/networking.py#L123-L147

However the `kubectl port-forward` process can close for a number of reasons, including timeouts or cluster autoscaling operations disrupting the stream.

Currently we do not track this process so if it exits prematurely the port forward will be lost.

To resolve this we could start our own thread/task and run the `kubectl port-forward` command in a loop, rerunning it if it exits, until it is cancelled.

jacobtomlinson:
Potentially closes #663.

Instead of calling `kubectl port-forward` in a subprocess I've been experimenting with implementing the port forward in pure Python.

I'm 90% of the way there but there are a few outstanding tasks:
- [x] Loop up a service and find the Pods. Choose one to open the port forward connection to.
- [x] Open the websocket stream with the Kubernetes API.
- [ ] Repeatedly open the port forward in a loop in a background task and cancel it when the cluster is closed. This will reconnect automatically and is the bit that closes #663.
- [ ] Figure out why the TCP sock and websocket seem to close prematurely.

### DIFF
--- a/dask_kubernetes/common/networking.py
+++ b/dask_kubernetes/common/networking.py
@@ -1,10 +1,9 @@
 import asyncio
+import aiohttp
 from contextlib import suppress
 import random
 import socket
-import subprocess
 import time
-from weakref import finalize
 import kubernetes_asyncio as kubernetes
 from tornado.iostream import StreamClosedError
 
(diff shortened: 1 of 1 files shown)


jacobtomlinson:
This functionality has been implemented in `kr8s` so we can close this here and just use that.

PR #668, opened 2023-03-08
Retire workers during scale down

jacobtomlinson:
I extracted these lines from #649. That PR seems to keep getting stuck in CI so I thought I'd break it down into some smaller chunks that we can merge more quickly and get to the bottom of the hang.


jacobtomlinson:
CI looks happy to going to merge this.

Issue #606, opened 2022-11-10
Recreate scheduler if lost


jacobtomlinson:
In #603 a bug was reported that workers are not automatically replaced if they are lost, e.g when a spot instance gets terminated. This opens the question of what should happen if the scheduler happens to be lost?

Losing the scheduler is a big deal because the client will need to reconnect and all work will be lost and need to be resubmitted. Therefore I would advise most folks to ensure the scheduler is on an on-demand node and only the workers are on the spot nodes.

Even so, the scheduler pod could be lost for any number of reasons from node failure to user error. 

Today the `DaskCluster` object would still exist and the workers would all go into a `CrashBackoffLoop` as they try and reconnect to the non-existing scheduler. The `DaskCluster` would never recover.

I suggest we update the controller to recreate the scheduler if it is ever lost so that the `DaskCluster` will eventually make it back to a healthy state.

webcoderz:
Ram into this in production on a client system the other day.. any chance you have a ETA when this fox will land? 
Thanks, you're doing great work on this operator! 😊

jacobtomlinson:
No ETA currently I'm afraid. Do you have any interest in contributing a PR?

PR #672, opened 2023-03-10
Auto shutdown idle clusters

### LINKED ISSUE #667: Add idle cluster cleanup
Inspired by https://github.com/dask/dask-gateway/issues/687 I think we should add a self cleanup for idle clusters here.

I expect the implementation would involve having some kind of configurable idle timeout in the `DaskCluster` resource and if this is set have the controller poll the scheduler via [a timer](https://kopf.readthedocs.io/en/stable/timers/) to find out if it is idle. If it is idle for longer than the timeout the controller would delete the `DaskCluster` resource.

jacobtomlinson:
Adds an optional `idleTimeout ` config option to the `DaskCluster` spec which instructs the controller to delete idle clusters after a certain timeout.

```yaml
kind: DaskCluster
metadata:
  name: foo
spec:
  idleTimeout: 60
  scheduler:
    ...
  worker:
    ...
```

Exposed via the Python API as the `idle_timeout` kwarg.

```python
cluster = KubeCluster(name="foo", n_workers=1, idle_timeout=60)
```

In the above examples if the scheduler is idle for more than 60 seconds the `DaskCluster` resource will be deleted automatically by the controller. This is regardless of whether the `KubeCluster` object still exists in Python.

~There is a challenge with this implementation which means I don't want to merge it in its current state~ Solved with https://github.com/dask/distributed/pull/7642

Closes #667

### DIFF
--- a/dask_kubernetes/operator/controller/controller.py
+++ b/dask_kubernetes/operator/controller/controller.py
@@ -383,7 +383,7 @@ async def retire_workers(
 
 
 async def check_scheduler_idle(scheduler_service_name, namespace, logger):
-    # Try gracefully retiring via the HTTP API
+    # Try getting idle time via HTTP API
     dashboard_address = await get_scheduler_address(
         scheduler_service_name,
         namespace,
@@ -404,7 +404,7 @@ async def check_scheduler_idle(scheduler_service_name, namespace, logger):
                 await resp.text(),
             )
 
-    # Otherwise try gracefully retiring via the RPC
+    # Otherwise try gracefully checking via the RPC
     logger.info(
         f"Checking {scheduler_service_name} idleness failed via the HTTP API, falling back to the Dask RPC"
     )
@@ -423,7 +423,7 @@ async def check_scheduler_idle(scheduler_service_name, namespace, logger):
 
     # Finally fall back to code injection via the Dask RPC for distributed<=2023.3.1
     logger.info(
-        f"Checking {scheduler_service_name} idleness failed via the Dask RPC, falling back to LIFO scaling"
+        f"Checking {scheduler_service_name} idleness failed via the Dask RPC, falling back to run_on_scheduler"
     )
 
     def idle_since(dask_scheduler=None):
@@ -858,13 +858,13 @@ async def daskautoscaler_adapt(spec, name, namespace, logger, **kwargs):
 
 @kopf.timer("daskcluster.kubernetes.dask.org", interval=5.0)
 async def daskcluster_autoshutdown(spec, name, namespace, logger, **kwargs):
-    if spec["autoCleanupTimeout"]:
+    if spec["idleTimeout"]:
         idle, idle_since = await check_scheduler_idle(
             scheduler_service_name=f"{name}-scheduler",
             namespace=namespace,
             logger=logger,
         )
-        if idle and time.time() > idle_since + spec["autoCleanupTimeout"]:
+        if idle and time.time() > idle_since + spec["idleTimeout"]:
             api = HTTPClient(KubeConfig.from_env())
             cluster = await DaskCluster.objects(api, namespace=namespace).get_by_name(
                 name
--- a/dask_kubernetes/operator/controller/controller.py
+++ b/dask_kubernetes/operator/controller/controller.py
@@ -394,10 +394,10 @@ async def check_scheduler_idle(scheduler_service_name, namespace, logger):
         url = f"{dashboard_address}/api/v1/check_idle"
         async with session.get(url) as resp:
             if resp.status <= 300:
-                idle, idle_since = await resp.json().values()
-                if idle:
+                idle_since = (await resp.json())["idle_since"]
+                if idle_since:
                     logger.debug("Scheduler idle since: %s", idle_since)
-                return idle, idle_since
+                return idle_since
             logger.debug(
                 "Received %d response from scheduler API with body %s",
                 resp.status,
@@ -416,10 +416,10 @@ async def check_scheduler_idle(scheduler_service_name, namespace, logger):
             allow_external=False,
         )
         async with rpc(comm_address) as scheduler_comm:
-            idle, idle_since = await scheduler_comm.check_idle()
-            if idle:
+            idle_since = await scheduler_comm.check_idle()
+            if idle_since:
                 logger.debug("Scheduler idle since: %s", idle_since)
-            return idle, idle_since
+            return idle_since
 
     # Finally fall back to code injection via the Dask RPC for distributed<=2023.3.1
     logger.info(
@@ -430,7 +430,7 @@ async def check_scheduler_idle(scheduler_service_name, namespace, logger):
         if not dask_scheduler.idle_timeout:
             dask_scheduler.idle_timeout = 300
         dask_scheduler.check_idle()
-        return dask_scheduler.idle_since is not None, dask_scheduler.idle_since
+        return dask_scheduler.idle_since
 
     comm_address = await get_scheduler_address(
         scheduler_service_name,
@@ -445,10 +445,10 @@ async def check_scheduler_idle(scheduler_service_name, namespace, logger):
             typ, exc, tb = clean_exception(**response)
             raise exc.with_traceback(tb)
         else:
-            idle, idle_since = response["result"]
-            if idle:
+            idle_since = response["result"]
+            if idle_since:
                 logger.debug("Scheduler idle since: %s", idle_since)
-            return idle, idle_since
+            return idle_since
 
 
 async def get_desired_workers(scheduler_service_name, namespace, logger):
@@ -859,12 +859,12 @@ async def daskautoscaler_adapt(spec, name, namespace, logger, **kwargs):
 @kopf.timer("daskcluster.kubernetes.dask.org", interval=5.0)
 async def daskcluster_autoshutdown(spec, name, namespace, logger, **kwargs):
     if spec["idleTimeout"]:
-        idle, idle_since = await check_scheduler_idle(
+        idle_since = await check_scheduler_idle(
             scheduler_service_name=f"{name}-scheduler",
             namespace=namespace,
             logger=logger,
         )
-        if idle and time.time() > idle_since + spec["idleTimeout"]:
+        if idle_since and time.time() > idle_since + spec["idleTimeout"]:
             api = HTTPClient(KubeConfig.from_env())
             cluster = await DaskCluster.objects(api, namespace=namespace).get_by_name(
                 name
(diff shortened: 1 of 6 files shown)


jacobtomlinson:
I wonder if another API here could be:

```yaml
kind: DaskCluster
metadata:
  name: foo
spec:
  idleTimeout: 60
  idleAction: "terminate"  # Or could be "scale" to just scale worker groups to zero
  scheduler:
    ...
  worker:
    ...
```

```python
cluster = KubeCluster(name="foo", n_workers=1, idle_timeout=60, idle_action="scale")
```

jacobtomlinson:
On second thoughts `idle_action="scale"` is pretty similar to adaptive scaling so probably not the best road to go down. Although renaming the spec item to `idleTimeout` to makes sense.

dask_kubernetes/operator/controller/controller.py
```diff
@@ -381,6 +382,75 @@ async def retire_workers(
...
+    )
+    # Dask version mismatches between the operator and scheduler may cause this to fail in any number of unexpected ways
+    with suppress(Exception):
+        comm_address = await get_scheduler_address(
+            scheduler_service_name,
+            namespace,
+            allow_external=False,
+        )
+        async with rpc(comm_address) as scheduler_comm:
+            idle, idle_since = await scheduler_comm.check_idle()
```
gjoseph92, line 419: I don't think this method on the scheduler is returning a tuple right now?

jacobtomlinson: Yeah I had done this in my local copy and then went down a different path. I need to update this PR to reflect the changes that landed in distributed. I'm partly waiting for a release to I can bump the minimum version.

dask_kubernetes/operator/controller/controller.py
```diff
@@ -381,6 +382,75 @@ async def retire_workers(
...
+async def check_scheduler_idle(scheduler_service_name, namespace, logger):
+    # Try getting idle time via HTTP API
+    dashboard_address = await get_scheduler_address(
+        scheduler_service_name,
+        namespace,
+        port_name="http-dashboard",
+        allow_external=False,
+    )
+    async with aiohttp.ClientSession() as session:
+        url = f"{dashboard_address}/api/v1/check_idle"
```
haddon-korinek-ent, line 394: The current `check_idle` impl on the scheduler looks like it will close itself via a background asyncio task if it passes the idle timeout from the regular dask config.

If a user happened to have `distributed.scheduler.idle-timeout` set, this might be confusing if their scheduler is exits due to idleness but the dask cluster resource isn't able to pick up on why it closed and then trigger the k8s dask cluster cleanup in response.

Another option could be to use the --death-timeout concept that workers have and auto-delete the dask cluster if the scheduler dies and stays dead for some period of time. This could help cleanup dask cluster k8s resources in cases the scheduler is unrecoverable and dask cluster operator won't be able to ever make a call to check idleness.

jacobtomlinson: I hear what you're saying. With this implementation, if the scheduler dies or becomes unresponsive we will never clean it up.

Today if the scheduler exits/dies it doesn't get recreated and the `DaskCluster` goes into a broken state. In #606 we discuss recreating the scheduler (probably using a `Deployment` or `ReplicaSet`). This is the opposite of what I understand you are proposing here which is cleaning up the whole `DaskCluster` if the scheduler is lost.

My view is that if a `DaskCluster` resource exists the controller should be working to get everything into a working state. We can achieve this via liveness/readiness probes and recreating missing `Pods`. If the scheduler or workers die it should be trying to reconcile and repair the cluster. We have some work to do to get to this place though.

I think actively reconciling the cluster and then asking the scheduler if it wants to time out is the right approach for this PR. But perhaps we also want some other timeout where if the controller can't get the cluster into a healthy state for a certain amount of time then it gives up and cleans it up.

haddon-korinek-ent: These are great points. The PR as is looks like it would alleviate nearly all of the manual work we're currently facing with regards to cleanup.

Very excited for it to land!

jacobtomlinson:
Tested this PR again today now that https://github.com/dask/distributed/pull/7781 has been merged and can confirm that it works as expected.

```python
cluster = KubeCluster(name="idle", n_workers=10, idle_timeout=120, env={"EXTRA_PIP_PACKAGES": "git+https://github.com/dask/distributed.git"})

# Cluster is deleted automatically after 2 mins
```

This PR will have to wait until the next Dask release but then we should be good to go.

vitawasalreadytaken:
Hi, may I ask if there's an ETA for merging and releasing this change? We were so happy to find that you were working on this and now we're really looking forward to this feature 😊  Thank you so much for your great work!

HynekBlaha:
Amazing :) 🤞

preneond:
@jacobtomlinson can we get some estimate when it will be done?

Artimi:
Hello, thanks for your work on this feature. We're looking forward to it.

I was trying to test it. This is my setup:
* Built dask kubernetes operator image from this branch, deployed it to kubernetes
* Using dask and distributed version `2023.3.2`
* Set `idle_timeout` to `10` both to `KubeCluster` init and `cluster_spec` which `KubeCluster` takes as an argument.
* Deploy a dask cluster which works about a minute and then stops

Findings:

1. When I start the cluster I can see this error in dask operator log

```
[2023-06-01 07:09:45,811] kopf.objects         [ERROR   ] [ana-ixian/ixian-petrsebek-36-violent-otheyms] Timer 'daskcluster_autoshutdown' failed with an exception. Will retry.
Traceback (most recent call last):                                                                                                                                                                          File "/usr/local/lib/python3.8/site-packages/kopf/_core/actions/execution.py", line 276, in execute_handler_once
    result = await invoke_handler(                                                                                                                                                                          File "/usr/local/lib/python3.8/site-packages/kopf/_core/actions/execution.py", line 371, in invoke_handler
    result = await invocation.invoke(                                                                                                                                                                       File "/usr/local/lib/python3.8/site-packages/kopf/_core/actions/invocation.py", line 116, in invoke                                                                                                         result = await fn(**kwargs)  # type: ignore                                                                                                                                                             File "/usr/local/lib/python3.8/site-packages/dask_kubernetes/operator/controller/controller.py", line 936, in daskcluster_autoshutdown                                                                      idle_since = await check_scheduler_idle(                                                                                                                                                                File "/usr/local/lib/python3.8/site-packages/dask_kubernetes/operator/controller/controller.py", line 433, in check_scheduler_idle
    dashboard_address = await get_scheduler_address(                                                                                                                                                        File "/usr/local/lib/python3.8/site-packages/dask_kubernetes/common/networking.py", line 177, in get_scheduler_address
    service = await api.read_namespaced_service(service_name, namespace)                                                                                                                                    File "/usr/local/lib/python3.8/site-packages/kubernetes_asyncio/client/api_client.py", line 192, in __call_api
    raise e                                                                                                                                                                                                 File "/usr/local/lib/python3.8/site-packages/kubernetes_asyncio/client/api_client.py", line 185, in __call_api
    response_data = await self.request(                                                                                                                                                                     File "/usr/local/lib/python3.8/site-packages/kubernetes_asyncio/client/rest.py", line 193, in GET                                                                                                           return (await self.request("GET", url,                                                                                                                                                                  File "/usr/local/lib/python3.8/site-packages/kubernetes_asyncio/client/rest.py", line 187, in request                                                                                                       raise ApiException(http_resp=r)                                                                                                                                                                       kubernetes_asyncio.client.exceptions.ApiException: (404)
Reason: Not Found                                                                                                                                                                                         HTTP response headers: <CIMultiDictProxy('Audit-Id': 'e926835b-5361-4f25-b05a-6fedae1d547e', 'Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'X-Kubernetes-Pf-Flowschema-Uid': '
a3ef744b-c3e6-4bf2-9c05-b8decfa0b75a', 'X-Kubernetes-Pf-Prioritylevel-Uid': '2d88b8ba-ee58-4770-a7b4-2342c139b0cc', 'Date': 'Thu, 01 Jun 2023 07:09:45 GMT', 'Content-Length': '264')>                    HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"services \"ixian-petrsebek-36-violent-otheyms-scheduler\" not found","reason":"NotFound","details":{"na
me":"ixian-petrsebek-36-violent-otheyms-scheduler","kind":"services"},"code":404}
```

This is probably because we are trying to get the scheduler address before the dask cluster is started.  It is not a big issue but it'd be nice to show it to user because it can be misleading.

2. After the dask cluster is running I can see that the `daskcluster_autoshutdown` is called repeatedly. However, the dask cluster is not stopped after it finishes its work and no new jobs are assigned. I can see this log

```
[2023-06-01 07:18:23,624] kopf.objects         [INFO    ] [ana-ixian/ixian-petrsebek-36-violent-otheyms] Checking ixian-petrsebek-36-violent-otheyms-scheduler idleness failed via the HTTP API, falling back to the Dask RPC                                                                                                                                                                                       [2023-06-01 07:18:23,636] kopf.objects         [INFO    ] [ana-ixian/ixian-petrsebek-36-violent-otheyms] Timer 'daskcluster_autoshutdown' succeeded.
```

So it seems that the HTTP API method fails, but the fallback method using Dask RPC is successful. But the cluster is never shut down nonetheless.

3. When I port-forward the port 8786 and I try to do the HTTP API request it returns an empty pickle file instead of some json with idle time as I'd expect. Am I doing it right?

```
$ curl --http0.9 http://localhost:8786/api/v1/check_idle --output check_idle
$ xxd check_idle
00000000: 3a00 0000 0000 0000 0100 0000 0000 0000  :...............
00000010: 2a00 0000 0000 0000 83ab 636f 6d70 7265  *.........compre
00000020: 7373 696f 6ec0 a670 7974 686f 6e93 030b  ssion..python...
00000030: 03af 7069 636b 6c65 2d70 726f 746f 636f  ..pickle-protoco
00000040: 6c05                                     l.
```

jacobtomlinson:
Thanks @Artimi.

> When I start the cluster I can see this error in dask operator log

It would be nice to clean that up. The controller will automatically retry but it's untidy and we should fix that.

> but the fallback method using Dask RPC is successful

The HTTP API is not enabled by default, so falling back to the RPC API is expected. 

> When I port-forward the port 8786 and I try to do the HTTP API request it returns an empty pickle file instead of some json with idle time as I'd expect. Am I doing it right?

The scheduler has two ports, it looks like you've forwarded the TCP comm instead of the web server.

jacobtomlinson:
Ok it looks like I've managed to correct the CI problems that were blocking this PR. Tests are passing which gives me confidence we can merge this. I'll give this another look over at the start of next week and check that it is working as expected and then hopefully hit the green button.

jacobtomlinson:
I've run through and tested this locally again. Things seem to be working as expected and I'm seeing clusters get cleaned up.

I made a couple of tweaks to the logging to reduce noise, but I think this should be good to merge on passing CI.

PR #676, opened 2023-03-14
Reduce local test flakyness

bstadlbauer:
I've originally started this PR in an effort to enable testing on aarch (https://github.com/dask/dask-kubernetes/pull/649#issuecomment-1439625841). Throughout fixing things, I've noticed that most changes are not directly related to aarch, so this should generally help folks with slower machines to run tests.
I ran all tests on a 2020 13" MBP (M1) with 16Gb ram.

Overall, I've made the following changes: 
- Increased timeouts throughout, as I've seen that performance is quite bad so the timeouts have been a bit too tight
- A small UTC time fix as `test_auto_refresh` failed for me in CET.
- Switched `test_pod_create_and_delete` to use `docker_image` instead of `gcr.io/google_containers/pause` as that's not available on ARM
- Made the `chart_repo` fixture more resilient
- Checked for pod readyness instead of "Running" as that should be more resilient (a pod can be `Running` but not ready)
- Removed the note about tests not working on M1 

@jacobtomlinson When I run all tests sequentially, all tests after the `helm` tests fail with `RuntimeError: Event loop is closed`. However, as CI never runs them sequentially I'm not sure whether it's worth fixing.

Note: @jmif while I originally required `from .controller import *` in `dask_kubernetes/operator/__init__.py` I could not reproduce this any more. Would you mind testing again with a freshly created environment? I suspect this might be due to a dependency upgrade, but without a lockfile this is hard to say. I've tested that it works for me with both Python 3.8 as well as Python 3.9.

### DIFF
--- a/dask_kubernetes/operator/controller/tests/test_controller.py
+++ b/dask_kubernetes/operator/controller/tests/test_controller.py
@@ -116,11 +116,13 @@ async def test_scalesimplecluster(k8s_cluster, kopf_runner, gen_cluster):
                 await asyncio.sleep(0.1)
             while worker_pod_name not in k8s_cluster.kubectl("get", "pods"):
                 await asyncio.sleep(0.1)
-            while "Running" not in k8s_cluster.kubectl(
-                "get", "pods", scheduler_pod_name
-            ):
-                await asyncio.sleep(0.1)
-            await asyncio.sleep(10)  # Wait for all components to be ready
+            k8s_cluster.kubectl(
+                "wait",
+                "pods",
+                "--for=condition=Ready",
+                scheduler_pod_name,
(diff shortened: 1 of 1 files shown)


jacobtomlinson:
Thanks so much for taking the time to look into this. 

>  However, as CI never runs them sequentially I'm not sure whether it's worth fixing.

The long-term plan is to remove the classic `KubeCluster` implementation and the `HelmCluster` class once the operator gains more traction. I'd like to update the helm chart to use a `DaskCluster` instead of a `Deployment`. So users of both will be able to use `dask_kubernetes.operator.KubeCluster` so I wouldn't invest any effort there now.

PR #689, opened 2023-03-27
Add prometheus monitors for schedulers and workers

### LINKED ISSUE #687: Add optional ServiceMonitor
On the Dask Forum user `preneond` asked if the operator has the ability to include a `ServiceMonitor` resource as the [helm chart does](https://github.com/dask/helm-chart/blob/40e7b688d474969af2e1ee14f2fedeb4afa2d0cb/dask/values.yaml#L37-L51). Right now the answer is no.

https://dask.discourse.group/t/deploying-dask-kubernetes-operator-with-enabled-service-monitor/1692

It would be nice to add this.

jacobtomlinson:
Add optional `ServiceMonitor` and `PodMonitor` resources to enable Prometheus scraping of Dask components.
As not all clusters have the Prometheus operator installed these are disabled by default. They can be enabled with the following comfig options.

```yaml
metrics:
   scheduler:
      enabled: true
      serviceMonitor:
         enabled: true
   worker:
      enabled: true
      serviceMonitor:
         enabled: true
```

You'll also need to ensure the container images you choose for your Dask components have the `prometheus_client` library installed.
If you're using the official Dask images you can install this at runtime.

```python
cluster = KubeCluster(name="monitored", env={"EXTRA_PIP_PACKAGES": "prometheus_client"})
```

Closes #687

### DIFF
--- a/dask_kubernetes/operator/deployment/helm/dask-kubernetes-operator/README.md
+++ b/dask_kubernetes/operator/deployment/helm/dask-kubernetes-operator/README.md
@@ -30,6 +30,24 @@ The following table lists the configurable parameters of the Dask-kubernetes-ope
 | `tolerations` | Tolerations | `[]` |
 | `affinity` | Affinity | `{}` |
 | `kopfArgs` | Command line flags to pass to kopf on start up | `["--all-namespaces"]` |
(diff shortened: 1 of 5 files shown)


preneond:
@jacobtomlinson  Tested it out and it worked well without any issues. It helped me so much, thank you!  🎆 🚀

jacobtomlinson:
Thank you so much for testing it, it's really helpful when folks do that.

Issue #536, opened 2022-07-20
Operator fails if nodeport out of allowed range


jacobtomlinson:
If a `DaskCluster` is configured with a `NodePort` service but the ports are out of range the `DaskCluster` will be created but the controller logs will error repeatedly.

```
HTTP response headers: <CIMultiDictProxy('Audit-Id': '8949180b-0ce2-45f7-bf41-8974bb2d6dbf', 'Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'X-Kubernetes-Pf-Flowschema-Uid': '8f735b53-3ddf-42b8-9a01-100a585f89dd', 'X-Kubernetes-Pf-Prioritylevel-Uid': 'a1296d7b-f6e8-49d6-86b1-398564e988c4', 'Date': 'Wed, 20 Jul 2022 14:34:11 GMT', 'Content-Length': '546')>
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Service \"rapids-dask-cluster-service\" is invalid: spec.ports[0].nodePort: Invalid value: 38786: provided port is not in the valid range. The range of valid ports is 30000-32767","reason":"Invalid","details":{"name":"rapids-dask-cluster-service","kind":"Service","causes":[{"reason":"FieldValueInvalid","message":"Invalid value: 38786: provided port is not in the valid range. The range of valid ports is 30000-32767","field":"spec.ports[0].nodePort"}]},"code":422}


[...]

PR #706, opened 2023-05-02
Add validation check for out-of-range nodeport values

skirui-source:
Fixes https://github.com/dask/dask-kubernetes/issues/536

### DIFF
--- a/dask_kubernetes/operator/controller/controller.py
+++ b/dask_kubernetes/operator/controller/controller.py
@@ -87,6 +87,15 @@ def build_scheduler_service_spec(cluster_name, spec, annotations, labels):
             "dask.org/component": "scheduler",
         }
     )
+
+    if spec.get("type") == "NodePort":
+        try:
+            for port in spec["ports"]:
+                node_port = port.get("nodePort")
+                assert node_port in range(30000, 32768)
+        except ValueError as e:
+            raise ValueError("Invalid nodePort value") from e
+
     return {
         "apiVersion": "v1",
         "kind": "Service",
--- a/dask_kubernetes/operator/controller/tests/test_controller.py
+++ b/dask_kubernetes/operator/controller/tests/test_controller.py
@@ -433,3 +433,7 @@ async def test_failed_job(k8s_cluster, kopf_runner, gen_job):
 
     assert "A DaskJob has been created" in runner.stdout
     assert "Job failed, deleting Dask cluster." in runner.stdout
+
+
+# def test_node_port_out_of_range():
+#     with kopf_runner as runner:
--- a/dask_kubernetes/operator/kubecluster/kubecluster.py
+++ b/dask_kubernetes/operator/kubecluster/kubecluster.py
@@ -230,6 +230,19 @@ class KubeCluster(Cluster):
         if isinstance(self.worker_command, str):
             self.worker_command = self.worker_command.split(" ")
 
+        if self.n_workers is not None and not isinstance(self.n_workers, int):
+            raise TypeError(f"n_workers must be an integer, got {type(self.n_workers)}")
+
+        try:
+            # Validate input resources
+            assert "limits" in resources
+            assert isinstance(resources["limits"], dict)
+            assert "CPU" in resources["limits"]
+            assert isinstance(resources["limits"]["CPU"], str)
+        except AssertionError:
+            print("Invalid resource limits format")
+            raise
+
         name = name.format(
             user=getpass.getuser(), uuid=str(uuid.uuid4())[:10], **os.environ
         )
--- a/dask_kubernetes/operator/kubecluster/kubecluster.py
+++ b/dask_kubernetes/operator/kubecluster/kubecluster.py
@@ -243,9 +243,8 @@ class KubeCluster(Cluster):
                     assert isinstance(self.resources[field], dict)
                 else:
                     raise ValueError(f"Unknown field '{field}' in resources")
-        except TypeError:
-            print(f"invalid '{type(resources)}' for resources type")
-            raise
+        except TypeError as e:
+            raise TypeError(f"invalid '{type(resources)}' for resources type") from e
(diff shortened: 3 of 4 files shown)


skirui-source:
minor icky thing-- why does this PR track changes that were already merged into main branch?
#weird

dask_kubernetes/operator/controller/controller.py
```diff
@@ -87,6 +87,17 @@ def build_scheduler_service_spec(cluster_name, spec, annotations, labels):
...
+
+    # Check if NodePort is out of range
+    try:
+        for port in spec.get("ports", []):
+            if port.get("nodePort", None) and (
+                port["nodePort"] < 30000 or port["nodePort"] > 32767
+            ):
+                raise ValueError("NodePort out of range")
+    except ValueError as e:
+        raise ValueError("Invalid nodePort value") from e
```
jacobtomlinson, line 99:
```suggestion
    for port in spec.get("ports", []):
        if port.get("nodePort", None) and (
            port["nodePort"] < 30000 or port["nodePort"] > 32767
        ):
            raise ValueError("NodePort out of range")
```

jacobtomlinson (commented):
Thanks @skirui-source. I don't think we need the outer try/catch as it just reraises the same exception. We can probably just let that bubble up to the user.

Could you please add a test to verify this functionality?

dask_kubernetes/operator/controller/controller.py
```diff
@@ -87,6 +87,14 @@ def build_scheduler_service_spec(cluster_name, spec, annotations, labels):
             "dask.org/component": "scheduler",
         }
     )
+
+    # Check if NodePort is out of range
+    for port in spec.get("ports", []):
+        if port.get("nodePort", None) and (
+            port["nodePort"] < 30000 or port["nodePort"] > 32767
+        ):
+            raise ValueError("NodePort out of range")
```
jacobtomlinson, line 96: Thinking about this more deeply I don't think we want to check this at this point. The resource has already been created and we are in an unschedulable state because the resource cannot be created.

I think we want to add a validation hook to check this at resource creation time.

https://kopf.readthedocs.io/en/stable/admission/#validation-handlers

```python
@kopf.on. validate("daskcluster.kubernetes.dask.org")
async def daskcluster_validate_nodeport(spec, warnings, **kwargs):
    for port in spec.get("ports", []):
        if port.get("nodePort", None) and (
            port["nodePort"] < 30000 or port["nodePort"] > 32767
        ):
            raise kopf.AdmissionError(f"nodePort must be between 30000 and 32767.")

**Comment by skirui-source** [comment]
Not sure why these tests are getting canceled..

**Comment by skirui-source** [comment]
> Not sure why the CI tests are getting canceled.. but seeing this error (locally). unsure how to resolve:
> 
>



In response to this, I followed these [instructions](https://kopf.readthedocs.io/en/stable/admission/#servers-and-tunnels ) to configure the admission  tunnel/server by adding this line [here]( ):
- `settings.admission.server = kopf.WebhookServer()  # defaults to localhost::port 9443`


Then i re ran the tests, but still failing as such: 
```
(2 durations < 0.005s hidden.  Use -vv to show these durations.)
================================================================================== short test summary info ==================================================================================
FAILED dask_kubernetes/operator/controller/tests/test_controller.py::test_nodeport_valid - asyncio.exceptions.TimeoutError: Dask Cluster resource not actioned after 60 seconds, is the Dask Operator running?
```


even though I can confirm that dask operator is running:
```
(kind-k8) skirui@dt08:~/dev/dask-kubernetes$ kubectl get pods
NAME                                                   READY   STATUS    RESTARTS   AGE
dask-kubernetes-operator-1683188683-7d754f579d-7ztxs   1/1     Running   0          3h57m
```

jacobtomlinson:
Tests get cancelled if they hang for more than 45 mins. 

It looks like you're on the right track with configuring the tunnel/server. I must admit this is more complicated than I expected. When the tests run the controller it outside of the Kubernetes cluster, so the Kubernetes API will not be able to communicate with the webhook. It looks like they have docs on using minikube or k3s but not kind. It might be best to leave this one with me for now and switch to another issue, I'll have a think about this.

Also, you shouldn't have a controller pod running like that, the tests will start/stop controllers so if you have a "production" one running too it's likely you will get conflicts. Do you have the operator help chart installed? If so you need to uninstall that for development.

skirui-source:
> Tests get cancelled if they hang for more than 45 mins.
> 
> It looks like you're on the right track with configuring the tunnel/server. I must admit this is more complicated than I expected. When the tests run the controller it outside of the Kubernetes cluster, so the Kubernetes API will not be able to communicate with the webhook. It looks like they have docs on using minikube or k3s but not kind. It might be best to leave this one with me for now and switch to another issue, I'll have a think about this.
> 
> Also, you shouldn't have a controller pod running like that, the tests will start/stop controllers so if you have a "production" one running too it's likely you will get conflicts. Do you have the operator help chart installed? If so you need to uninstall that for development.

Still confused about the difference between operator, controller and Kubernetes API server? 
Also if the tests automatically start/stop "controller", how can i check the status of said controller at any given time?

jacobtomlinson:
> Still confused about the difference between operator, controller and Kubernetes API server?

**Operator** is a design pattern where you extend the Kubernetes API with Custom Resource Definitions and a custom controller. You effectively treat Kubernetes like a database and have a controller reconsile the state of the world with what that database describes.

**Custom Resource Definition** is a description of a new object type for Kubernetes to track. It basically creates a new table in the Kubernetes database and describes the data structure for your custom resource type.

**Controller** is the process that runs on the Kubernetes cluster and listens for API events and takes action (e.g on `DaskCluster` create event set up the cluster resources).

**Kubernetes API server** the API that `kubectl`, `kubernetes_asyncio`, `kubelet`, etc interacts with. Is backed by a database (usually etcd).

Issue #698, opened 2023-04-19
Invalid KubeCluster kwargs raises confusing exception


jacobtomlinson:
Today when creating a cluster I accidentally passed `n_workers` as a string instead of an int and got a confusing error message about CRDs not being installed.

```python
In [1]: from dask_kubernetes.operator import KubeCluster
   ...: cluster = KubeCluster(name="idle", n_workers="10")

RuntimeError: Failed to create DaskCluster resource. Are the Dask Custom Resource Definitions installed? 
```

<details>
<summary>Full traceback</summary>

```pytb
---------------------------------------------------------------------------
ApiException                              Traceback (most recent call last)
~/Projects/dask/dask-kubernetes/dask_kubernetes/operator/kubecluster/kubecluster.py in _create_cluster(self)
    325                 self._log("Creating DaskCluster object")
--> 326                 await custom_objects_api.create_namespaced_custom_object(
    327                     group="kubernetes.dask.org",

~/.local/lib/python3.8/site-packages/kubernetes_asyncio/client/api_client.py in __call_api(self, resource_path, method, path_params, query_params, header_params, body, post_params, files, response_types_map, auth_settings, _return_http_data_only, collection_formats, _preload_content, _request_timeout, _host, _request_auth)
    191             e.body = e.body.decode('utf-8') if six.PY3 else e.body
--> 192             raise e
    193 

~/.local/lib/python3.8/site-packages/kubernetes_asyncio/client/api_client.py in __call_api(self, resource_path, method, path_params, query_params, header_params, body, post_params, files, response_types_map, auth_settings, _return_http_data_only, collection_formats, _preload_content, _request_timeout, _host, _request_auth)
    184             # perform request and return response
--> 185             response_data = await self.request(
    186                 method, url, query_params=query_params, headers=header_params,

~/.local/lib/python3.8/site-packages/kubernetes_asyncio/client/rest.py in POST(self, url, headers, query_params, post_params, body, _preload_content, _request_timeout)
    229                    _request_timeout=None):
--> 230         return (await self.request("POST", url,
    231                                    headers=headers,

~/.local/lib/python3.8/site-packages/kubernetes_asyncio/client/rest.py in request(self, method, url, query_params, headers, body, post_params, _preload_content, _request_timeout)
    186             if not 200 <= r.status <= 299:
--> 187                 raise ApiException(http_resp=r)
    188 

ApiException: (422)
Reason: Unprocessable Entity
[...]

PR #708, opened 2023-05-04
Add correct handling for invalid `KubeCluster` kwargs

skirui-source:
Fixes: https://github.com/dask/dask-kubernetes/issues/698

Tasks
 

- [x] Update error handling to only raise CRD error if CRDs are missing
- [x]  Update error to raise a helpful message when a 422 is returned to point the user towards which kwarg is invalid

### DIFF
--- a/dask_kubernetes/operator/kubecluster/kubecluster.py
+++ b/dask_kubernetes/operator/kubecluster/kubecluster.py
@@ -230,6 +230,19 @@ class KubeCluster(Cluster):
         if isinstance(self.worker_command, str):
             self.worker_command = self.worker_command.split(" ")
 
+        if self.n_workers is not None and not isinstance(self.n_workers, int):
+            raise TypeError(f"n_workers must be an integer, got {type(self.n_workers)}")
+
+        try:
+            # Validate input resources
+            assert "limits" in resources
+            assert isinstance(resources["limits"], dict)
+            assert "CPU" in resources["limits"]
+            assert isinstance(resources["limits"]["CPU"], str)
+        except AssertionError:
+            print("Invalid resource limits format")
+            raise
+
         name = name.format(
             user=getpass.getuser(), uuid=str(uuid.uuid4())[:10], **os.environ
         )
--- a/dask_kubernetes/operator/kubecluster/kubecluster.py
+++ b/dask_kubernetes/operator/kubecluster/kubecluster.py
@@ -233,15 +233,18 @@ class KubeCluster(Cluster):
         try:
             # Validate `resources` param is a dictionary whose
             # keys must either be 'limits' or 'requests'
-            assert isinstance(self.resources, dict)
-
+            assert isinstance(
+                self.resources, dict
+            ), f"resources must be dict type, found {type(resources)}"
             for field in self.resources:
                 if field in ("limits", "requests"):
-                    assert isinstance(self.resources[field], dict)
+                    assert isinstance(
+                        self.resources[field], dict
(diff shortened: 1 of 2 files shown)


dask_kubernetes/operator/kubecluster/kubecluster.py
```diff
@@ -339,11 +339,16 @@ async def _create_cluster(self):
...
+                if e.status == 422:
+                    # unsure how to extract the value of the invalid kwarg
+                    # from the API response to populate below
+                    raise ValueError("unable to process invalid kwarg(s)")
+                elif e.status == 404:
+                    raise RuntimeError(
+                        "Failed to create DaskCluster resource. "
+                        "Are the Dask Custom Resource Definitions installed? "
+                        "https://kubernetes.dask.org/en/latest/operator.html#installing-the-operator"
+                    ) from e
```
jacobtomlinson, line 351: This is interesting. Does it always give a 404 if the CRDs are missing? And is that the only situation that it would raise a 404?

If so I would just simplify this right down and let the Kubernetes error bubble up to the user.

```
if e.status == 404:
    raise RuntimeError(...) from e
else:
    raise e
```

skirui-source:
BLOCKED : Refer to [comment ](https://github.com/dask/dask-kubernetes/issues/698#issuecomment-1541421885) -- unable to reproduce the error

dask_kubernetes/operator/kubecluster/kubecluster.py
skirui-source: Also I am confused, in what case would the CRD be missing? 

**IF**  the code automatically generate a custom resource  spec with default values via **make_cluster_spec()**  if one isn't passed to `KubeCluster()` ..

then creates the CRD via **create_namespaced_custom_object()**  ?

jacobtomlinson: I mean the case where the user has skipped installing the operator and just tries to use `KubeCluster`. When I say CRD I mean Custom Resource **Definition** which describes the custom resource. These get installed by the helm chart. Pytest also installs/removes these via a fixture for testing.

What you are referring to that `create_namespaced_custom_object` creates is an instance of that CRD, usually just referred to as a *resource*. Without the CRDs Kubernetes has no idea what a `DaskCluster` or `DaskWorkerGroup` even is, and it looks like it returns a 404 in this case.

So my guess is that it returns a 404 if the CRDs are missing, and a 422 if you try and create something that doesn't meet the CRDs expectations.

skirui-source:
Ready for Review / MERGE

dask_kubernetes/operator/kubecluster/kubecluster.py
```diff
@@ -230,21 +230,21 @@ def __init__(
...
+                    assert isinstance(
+                        self.resources[field], dict
+                    ), f"key of '{field}' must be dict type"
                 else:
-                    raise ValueError(f"Unknown field '{field}' in resources")
-        except TypeError as e:
-            raise TypeError(f"invalid '{type(resources)}' for resources type") from e
+                    raise ValueError(f"resources has unknown field '{field}'")
+        except AssertionError as e:
+            raise e
```
jacobtomlinson, line 247: I think we should be raising a `TypeError` here instead or simply reraising the `AssertionError`

jacobtomlinson (commented):
One small comment

jacobtomlinson (approved):
This look great thanks @skirui-source.

I think the CI failure is unrelated, I've triggered it again, hopefully it will pass and then we can get this merged.

PR #711, opened 2023-05-11
Create scheduler pod through a Deployment

### LINKED ISSUE #606: Recreate scheduler if lost

Matt711:
Closes #606. Currently, if the scheduler pod is lost, the user will lose all of their work because the Client will need to be connected. One way we can prevent this is by creating the scheduler pod through a deployment in stead of creating the pod directly. The advantage of a deployment is that we can define the number of replicas (1 replica/scheduler in this case) and Kubernetes will recreate the scheduler pod if it's lost. This, way, the workers can reconnect to the scheduler.

### DIFF
--- a/dask_kubernetes/common/networking.py
+++ b/dask_kubernetes/common/networking.py
@@ -196,8 +196,8 @@ async def get_scheduler_address(
 async def wait_for_scheduler(api, cluster_name, namespace, timeout=None):
     pod_start_time = None
     while True:
-        pod = await Pod.objects(api, namespace=namespace).get_by_name(
-            cluster_name + "-scheduler"
+        pod = await Pod.objects(api, namespace=namespace).get(
+            **{"selector" : {"dask.org/component":"scheduler-deployment"}}
         )
         phase = pod.obj["status"]["phase"]
         if phase == "Running":
--- a/dask_kubernetes/operator/controller/controller.py
+++ b/dask_kubernetes/operator/controller/controller.py
@@ -60,19 +60,20 @@ def _get_labels(meta):
     }
 
 
-def build_scheduler_pod_spec(cluster_name, spec, annotations, labels):
+def build_scheduler_deployment_spec(cluster_name, spec, annotations, labels):
     labels.update(
         **{
             "dask.org/cluster-name": cluster_name,
-            "dask.org/component": "scheduler",
+            "dask.org/component": "scheduler-deployment",
             "sidecar.istio.io/inject": "false",
         }
     )
+    spec["replicas"]=1
     return {
         "apiVersion": "v1",
-        "kind": "Pod",
+        "kind": "Deployment",
         "metadata": {
-            "name": f"{cluster_name}-scheduler",
+            "name": f"{cluster_name}-scheduler-deployment",
             "labels": labels,
             "annotations": annotations,
         },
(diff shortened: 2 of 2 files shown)


Matt711:
The deployment is being created. But I'm getting `ConnectionError: kubectl port forward failed`. 
1. `python -m kopf run dask_kubernetes/operator/controller/controller.py`
2. 
```python
>>> from dask_kubernetes.operator import KubeCluster
>>> cluster = KubeCluster(name="foo")
```

dask_kubernetes/common/networking.py
```diff
@@ -196,8 +196,14 @@ async def get_scheduler_address(
 async def wait_for_scheduler(api, cluster_name, namespace, timeout=None):
     pod_start_time = None
     while True:
+        async with kubernetes.client.api_client.ApiClient() as api_client:
+            k8s_api = kubernetes.client.CoreV1Api(api_client)
+            pods = await k8s_api.list_namespaced_pod(
+                namespace=namespace,
+                label_selector="dask.org/component=scheduler-deployment",
+            )
```
jacobtomlinson, line 204: We probably also need to add another selector for the `dask.org/cluster={cluster_name}` to make sure we are selecting just the scheduler we want.

dask_kubernetes/operator/controller/controller.py
```diff
@@ -271,13 +272,40 @@ async def daskcluster_create_components(
...
+                metadata = body_metadata,
+                spec = kubernetes.client.V1PodSpec(
+                    containers=[container],
+                )
+            ),
+        )
+        body = kubernetes.client.V1Deployment(
+            metadata=body_metadata,
+            spec=body_spec,
+        )
```
jacobtomlinson, line 305: Can we not create this directly from `data`?

tasansal:
This is awesome. We are losing the scheduler randomly all the time on multiple production deployments. Any ETA on the release?

The logs don't have any information, the scheduler all of a sudden receives a SIGTERM. We don't know why yet.

image

jacobtomlinson:
@tasansal we are blocked on some CI problems which I'm exploring now

dask_kubernetes/common/networking.py
```diff
@@ -196,8 +196,14 @@ async def get_scheduler_address(
 async def wait_for_scheduler(api, cluster_name, namespace, timeout=None):
     pod_start_time = None
     while True:
+        async with kubernetes.client.api_client.ApiClient() as api_client:
+            k8s_api = kubernetes.client.CoreV1Api(api_client)
+            pods = await k8s_api.list_namespaced_pod(
+                namespace=namespace,
+                label_selector=f"dask.org/component=scheduler,dask.org/cluster-name={cluster_name}",
+            )
```
Matt711, line 204: This lines are used in a few places, so I could replace with a helper function. But it'll be replaced with [Kr8s](https://github.com/kr8s-org/kr8s) soon

Matt711:
image

jacobtomlinson (changes requested):
This generally looks great! I tried this out locally and things seem to mostly work as expected.

I noticed the widget didn't seem to work as expected, the values never update.

image

I also noticed that if you delete the scheduler Pod it gets recreated, and the workers restart and reconnect. However, if you have a client connected it never reconnects. That's probably not something to worry about here but we might want to open an issue on distributed to discuss this functionality.

It would also be nice to add a test which creates a cluster, deletes the scheduler Pod and asserts that it gets replaced.

dask_kubernetes/operator/controller/tests/test_controller.py
```diff
@@ -360,6 +336,34 @@ async def test_scalesimplecluster_from_cluster_spec(
...
+                "delete",
+                "pods",
+                "-l",
+                "dask.org/cluster-name=simple,dask.org/component=scheduler",
+                "-n",
+                ns,
+            )
+            assert scheduler_deployment_name in k8s_cluster.kubectl(
+                "get", "pods", "-n", ns
+            )
```
jacobtomlinson, line 364: This could potentially pick up the deleted Pod which may be in a `Terminating` phase. We might want to wait for the new Pod to be ready.

```suggestion
            k8s_cluster.kubectl(
                "wait",
                "--for=condition=Ready",
                "-l",
                "dask.org/cluster-name=simple,dask.org/component=scheduler",
                "pod",
                "-n",
                ns,
                "--timeout=60s",
            )
            assert scheduler_deployment_name in k8s_cluster.kubectl(
                "get", "pods", "-n", ns
            )
```

jacobtomlinson (approved):
Tested locally and things seem to be working great. Happy to merge on passing CI.

bstadlbauer:
Quick update in case anyone runs into this. We've just tried deploying `2023.6.1` (which includes this fix), but our Flyte plugin ran into: 

```
https://github.com/flyteorg/flyteplugins/pull/367
```
spec.template.spec.restartPolicy: Unsupported value: "Never": supported values: Always
```

I've opened https://github.com/flyteorg/flyteplugins/pull/367 accordingly, so in case anyone runs into this, the latest Flyte release should have a fix.

@jacobtomlinson Do you happen to know what will happen in case the scheduler fails as part of a job? I'm assuming the job will still be terminated as expected?

jacobtomlinson:
If the scheduler fails it will be restarted. I would expect that to cause the job Pod to fail (although users could write their code in a way that handles this). When the job Pod fails the `DaskJob` is marked as failed and the `DaskCluster` is cleaned up.

PR #713, opened 2023-05-11
DOC: Add documentation for contributing to dask kubernetes

Matt711:
It would be helpful to have a contributing page in the documentation for those who want set up their development environments correctly and contribute to Dask Kubernetes. I'll start with information on the Dask Operator (Installing the Custom Resource Definitions, running the controller, etc.)

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -20,7 +20,11 @@
 import os
 import sys
 
-sys.path.insert(0, os.path.abspath(".."))
+try:
+    from dask_kubernetes import __version__
+except ImportError:
+    sys.path.insert(0, os.path.abspath("../.."))
+    from dask_kubernetes import __version__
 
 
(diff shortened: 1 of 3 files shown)


jacobtomlinson:
There is an existing `testing` page which has some of this information. Perhaps we want to replace that with something more useful?

Matt711:
> There is an existing `testing` page which has some of this information. Perhaps we want to replace that with something more useful?

Yes, I was thinking of deleting it and incorporating it into this new page. I would like "both" methods of testing (pytest and running the controller directly) to be on one page. I think it makes more sense for both of them to be on the Developer page.

jacobtomlinson:
Sounds great!

jacobtomlinson:
Enough time has passed that I'm tempted to close this out. We can always use this for further docs improvement inspiration in the future.

PR #715, opened 2023-05-16
Unblock classic KubeCluster CI

jacobtomlinson:
In #711 the classic CI is failing. The change to `dask_kubernetes/common/networking.py` is triggering the CI, but the function that is modified doesn't appear to be used in classic so it should be causing the CI to fail.

Opened this PR with a superficial fix in `dask_kubernetes/classic/kubecluster.py` to trigger the CI independently to see if it fails here.

### DIFF
--- a/dask_kubernetes/classic/tests/test_async.py
+++ b/dask_kubernetes/classic/tests/test_async.py
@@ -825,6 +825,7 @@ async def test_adapt_delete(cluster, ns):
 
 
 @pytest.mark.asyncio
+@pytest.mark.xfail(reason="Failing in CI with FileNotFoundError")
 async def test_auto_refresh(cluster):
     config = {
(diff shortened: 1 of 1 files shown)


jacobtomlinson:
Yup this is the same failure we are seeing in #711.

```
FAILED dask_kubernetes/classic/tests/test_async.py::test_auto_refresh - FileNotFoundError: [Errno 2] No such file or directory: 'python /home/runner/work/dask-kubernetes/dask-kubernetes/dask_kubernetes/classic/tests/fake_gcp_auth.py'
```

jacobtomlinson:
Given that the plan is to retire the classic `KubeCluster` in the next few months, and that GCP auth is now deprecated I'm tempted to skip the failing test for now to unblock CI.

PR #720, opened 2023-05-23
Scale Dask worker group from Cluster spec

### LINKED ISSUE #636: k8s Operator: DaskCluster spec updates do not propagate to pods
We’ve deployed the disk-operator and custom resources (DaskCluster, DaskAutoscaler) via helm. 

When updating DaskCluster specs (e.g. scheduler container resources requests - cpu and memory) via helm upgrade, DaskCluster custom resource reflects the new configurations correctly; however, the pods (scheduler and workers) are not re-deployed with the new changes.

Matt711:
Closes #636. This PR adds another handler that scales the default worker group when the `worker.replicas` field of the cluster spec is changed.

### DIFF
--- a/dask_kubernetes/operator/controller/controller.py
+++ b/dask_kubernetes/operator/controller/controller.py
@@ -481,6 +481,29 @@ async def get_desired_workers(scheduler_service_name, namespace, logger):
 worker_group_scale_locks = defaultdict(lambda: asyncio.Lock())
 
 
+@kopf.on.field("daskcluster.kubernetes.dask.org", field="spec.worker.replicas")
+async def daskcluster_default_worker_group_replica_update(
+    name, namespace, meta, spec, old, new, body, logger, **kwargs
+):
+    if old is None:
+        return
+    worker_group_name = f"{name}-default"
+
+    async with kubernetes.client.api_client.ApiClient() as api_client:
+        custom_objects_api = kubernetes.client.CustomObjectsApi(api_client)
+        custom_objects_api.api_client.set_default_header(
+            "content-type", "application/merge-patch+json"
+        )
+        await custom_objects_api.patch_namespaced_custom_object_scale(
+            group="kubernetes.dask.org",
+            version="v1",
+            plural="daskworkergroups",
+            namespace=namespace,
+            name=worker_group_name,
+            body={"spec": {"replicas": new}},
+        )
+
+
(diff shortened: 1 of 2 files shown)


dask_kubernetes/operator/controller/controller.py
```diff
@@ -481,6 +481,29 @@ async def get_desired_workers(scheduler_service_name, namespace, logger):
...
+            "content-type", "application/merge-patch+json"
+        )
+        await custom_objects_api.patch_namespaced_custom_object_scale(
+            group="kubernetes.dask.org",
+            version="v1",
+            plural="daskworkergroups",
+            namespace=namespace,
+            name=worker_group_name,
+            body={"spec": {"replicas": new}},
+        )
```
jacobtomlinson, line 504: I'm excited about the `kr8s` changes in #696 because this will be hugely simplified.

```suggestion
    worker_group = await DaskWorkerGroup.get(worker_group_name)
    await worker_group.scale(new)
```

Happy to merge this as-is though and I'll update it in a follow-up once `kr8s` is integrated nicely.

dask_kubernetes/operator/controller/tests/test_controller.py
```diff
@@ -143,6 +143,48 @@ async def test_scalesimplecluster(k8s_cluster, kopf_runner, gen_cluster):
...
+                        cluster_name,
+                    )
+                    await client.wait_for_workers(5)
+                    k8s_cluster.kubectl(
+                        "scale",
+                        "--replicas=3",
+                        "daskcluster.kubernetes.dask.org",
+                        cluster_name,
+                    )
+                    await client.wait_for_workers(3)
```
jacobtomlinson, line 185: This wont actually test anything because `client.wait_for_workers(3)` will just pass when there are 5 workers.

I opened https://github.com/dask/distributed/pull/6377 a while ago to allow you to modify the behaviour for cases like this. I should nudge that PR along.

jacobtomlinson (approved):
This is a really nicely written PR! Thanks for picking this up.

dask_kubernetes/operator/controller/controller.py
Matt711: Oh, wow! Definitely looking forward to that!

jacobtomlinson (approved):

Issue #670, opened 2023-03-09
Flaky test blocking CI


jacobtomlinson:
CI seems to be intermittently hanging some of the time and timing out at the 45 minute limit. It generally happens to at leats one check every run. Rerunning the failed check usually gets it to pass.

The pytest timeouts don't seem to behaving any effect and it's hanging until the workflow timeout. This usually indicates the hang is in a fixture not a test itself.

Looking at the logs the stuck test seems to be `dask_kubernetes/operator/controller/tests/test_controller.py::test_simplecluster`.

Example:
https://github.com/dask/dask-kubernetes/actions/runs/4373514542/jobs/7651785461

```
2023-03-09T10:35:58.8324020Z ============================= test session starts ==============================
2023-03-09T10:35:58.8325891Z platform linux -- Python 3.10.10, pytest-7.2.2, pluggy-1.0.0 -- /opt/hostedtoolcache/Python/3.10.10/x64/bin/python
2023-03-09T10:35:58.8327067Z cachedir: .pytest_cache
2023-03-09T10:35:58.8328416Z rootdir: /home/runner/work/dask-kubernetes/dask-kubernetes, configfile: setup.cfg
2023-03-09T10:35:58.8329088Z plugins: rerunfailures-11.1.2, kind-22.11.1, timeout-2.1.0, asyncio-0.20.3
2023-03-09T10:35:58.8329443Z timeout: 300.0s
2023-03-09T10:35:58.8329712Z timeout method: signal
2023-03-09T10:35:58.8329964Z timeout func_only: False
2023-03-09T10:35:58.8330226Z asyncio: mode=strict
2023-03-09T10:35:58.9779759Z collecting ... collected 29 items
2023-03-09T10:35:58.9780032Z 
2023-03-09T10:38:41.7319744Z dask_kubernetes/common/tests/test_kind.py::test_config_detection PASSED  [  3%]
2023-03-09T10:38:41.7953196Z dask_kubernetes/common/tests/test_kind.py::test_auth XPASS (Has asyn...) [  6%]
2023-03-09T10:38:41.7978304Z dask_kubernetes/common/tests/test_objects.py::test_make_pod_from_dict PASSED [ 10%]
2023-03-09T10:38:41.8004372Z dask_kubernetes/common/tests/test_objects.py::test_make_pod_from_dict_default_container_name PASSED [ 13%]
[...]

PR #721, opened 2023-05-24
Replace aiopykube with kr8s in controller

jacobtomlinson:
Taking a more methodical approach to #696 and replacing one thing at a time.

This first PR adds `kr8s` as a dependency, defines the custom resource objects in the controller and updates `handle_scheduler_service_status()` to use `kr8s` for setting the `DaskCluster` status.


jacobtomlinson:
All of the failures here appear to be the same as in #670

PR #726, opened 2023-05-29
Add extra config arguments into pods.  Helpful for taints, affinity, nodeSelector

danielhstahl:
Currently I have to create a yaml file to enable taints, affinity, and nodeSelector, bypassing the convenience of the high-level python API.

This PR addresses this shortcoming by allowing a dictionary to be passed to the `extra_config_args` parameter.

### DIFF
--- a/dask_kubernetes/kubernetes.yaml
+++ b/dask_kubernetes/kubernetes.yaml
@@ -4,6 +4,7 @@ kubernetes:
 
   # Shared options
   name: "dask-{user}-{uuid}"
+  extra_pod_config: {}
   namespace: null
   env: {}
   count:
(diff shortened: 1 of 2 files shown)


jacobtomlinson (commented):
Thanks for taking the time to raise this. However I'm afraid we can't accept this contribution ase already have an API for customizing Pods which you can find [documented here](https://kubernetes.dask.org/en/latest/operator_kubecluster.html#customising-your-cluster).

The high-level API allows you to generate the cluster spec and create the cluster in two separate steps. This allows you to tweak the spec however you like while still benefiting from the high-level API.

For example if you wanted to add a `nodeSelector` you would do it like this.

```python
from dask_kubernetes.operator import KubeCluster, make_cluster_spec

spec = make_cluster_spec(name="selector-example", n_workers=2)
spec["spec"]["worker"]["spec"]["nodeSelector"] = {"disktype": "ssd"}

cluster = KubeCluster(custom_cluster_spec=spec)
```

PR #727, opened 2023-05-30
KubeCluster.shutdown_on_close should be False when created from_name

### LINKED ISSUE #725: KubeCluster.shutdown_on_close defaults to True
**Describe the issue**:
The [docstring](https://github.com/dask/dask-kubernetes/blob/main/dask_kubernetes/operator/kubecluster/kubecluster.py#L102-L104) says 
```
    shutdown_on_close: bool (optional)
        Whether or not to delete the cluster resource when this object is closed.
        Defaults to ``True`` when creating a cluster and ``False`` when connecting to an existing one.
```
which is the behavior I'd expect. But in practice it's just `True` even for an existing cluster, which means when you close your interpreter/notebook kernel you kill your cluster (found this out the hard way!).

# Created a cluster `brett` w/ `kubectl`
image

**Environment**:

- Dask version:
- Python version:
- Operating System:
- Install method (conda, pip, source):

Matt711:
Closes #725 .

### DIFF
--- a/dask_kubernetes/operator/kubecluster/kubecluster.py
+++ b/dask_kubernetes/operator/kubecluster/kubecluster.py
@@ -867,7 +867,12 @@ class KubeCluster(Cluster):
         --------
         >>> cluster = KubeCluster.from_name(name="simple-cluster")
         """
-        return cls(name=name, create_mode=CreateMode.CONNECT_ONLY, **kwargs)
+        return cls(
+            name=name,
+            create_mode=CreateMode.CONNECT_ONLY,
+            shutdown_on_close=False,
+            **kwargs,
+        )
 
 
 def make_cluster_spec(
--- a/dask_kubernetes/operator/kubecluster/kubecluster.py
+++ b/dask_kubernetes/operator/kubecluster/kubecluster.py
@@ -867,10 +867,10 @@ class KubeCluster(Cluster):
         --------
         >>> cluster = KubeCluster.from_name(name="simple-cluster")
         """
+        defaults = {"create_mode": CreateMode.CONNECT_ONLY, "shutdown_on_close": False}
+        kwargs = defaults | kwargs
         return cls(
             name=name,
-            create_mode=CreateMode.CONNECT_ONLY,
-            shutdown_on_close=False,
             **kwargs,
         )
 
--- a/dask_kubernetes/operator/kubecluster/tests/test_kubecluster.py
+++ b/dask_kubernetes/operator/kubecluster/tests/test_kubecluster.py
@@ -5,6 +5,7 @@ from dask.distributed import Client
 from distributed.utils import TimeoutError
 
 from dask_kubernetes.aiopykube.dask import DaskCluster
+from dask_kubernetes.aiopykube import HTTPClient, KubeConfig
 from dask_kubernetes.operator import KubeCluster, make_cluster_spec
 from dask_kubernetes.exceptions import SchedulerStartupError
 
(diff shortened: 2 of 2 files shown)


dask_kubernetes/operator/kubecluster/kubecluster.py
```diff
@@ -867,7 +867,12 @@ def from_name(cls, name, **kwargs):
         --------
         >>> cluster = KubeCluster.from_name(name="simple-cluster")
         """
-        return cls(name=name, create_mode=CreateMode.CONNECT_ONLY, **kwargs)
+        return cls(
+            name=name,
+            create_mode=CreateMode.CONNECT_ONLY,
+            shutdown_on_close=False,
+            **kwargs,
+        )
```
jacobtomlinson, line 875: _Pseudo code and probably wont pass linting. But we could use the dict pipe merge now that 3.9 is our minimum!_

```suggestion
        defaults = {"create_mode": CreateMode.CONNECT_ONLY, "shutdown_on_close": False}
        kwargs = defaults | kwargs
        return cls(
            name=name,
            **kwargs,
        )
```

jacobtomlinson (commented):
Thanks for raising this. I wonder if we can somehow allow folks to specify these options if they really want to.

For example with this change a user wouldn't be able to do the following

```python
cluster = KubeCluster.from_name("foo", shutdown_on_close=True)
```

jacobtomlinson (commented):
Thanks for raising this. I wonder if we can somehow allow folks to specify these options if they really want to.

For example with this change a user wouldn't be able to do the following

```python
cluster = KubeCluster.from_name("foo", shutdown_on_close=True)
```

jacobtomlinson (commented):
Looks like CI is passing here. Could I ask you to add/update a test to cover this functionality?

dask_kubernetes/operator/kubecluster/tests/test_kubecluster.py
```diff
@@ -4,6 +4,8 @@
 from dask.distributed import Client
 from distributed.utils import TimeoutError
 
+from dask_kubernetes.aiopykube.dask import DaskCluster
+from dask_kubernetes.aiopykube import HTTPClient, KubeConfig
```
jacobtomlinson, line 8:
```suggestion
from dask_kubernetes.operator.objects import DaskCluster
```

dask_kubernetes/operator/kubecluster/tests/test_kubecluster.py
```diff
@@ -95,13 +97,25 @@ def test_multiple_clusters_simultaneously_same_loop(kopf_runner, docker_image):
...
         ) as firstcluster:
-            with KubeCluster.from_name("abc", namespace=ns) as secondcluster:
+            with KubeCluster.from_name(
+                "abc", namespace=ns, asynchronous=True
+            ) as secondcluster:
                 assert firstcluster == secondcluster
+            k8s_api = HTTPClient(KubeConfig.from_env())
+            cluster = await DaskCluster.objects(k8s_api, namespace=ns).get_by_name(
+                "abc"
+            )
```
jacobtomlinson, line 117:
```suggestion
            cluster = await DaskCluster.get("abc", namespace=ns)
```

dask_kubernetes/operator/kubecluster/tests/test_kubecluster.py
```diff
@@ -95,13 +97,25 @@ def test_multiple_clusters_simultaneously_same_loop(kopf_runner, docker_image):
...
+            assert cluster.obj["status"]["phase"] == "Running"
```
jacobtomlinson, line 118:
```suggestion
            assert cluster.status["phase"] == "Running"
```

jacobtomlinson (changes requested):
Awesome thanks!

The first `kr8s` PR has landed with #721 so we can start using it 🎉.

I made some quick suggestions here that you should be able to use to swap them out.

Matt711:
> Awesome thanks!
> 
> The first `kr8s` PR has landed with #721 so we can start using it 🎉.
> 
> I made some quick suggestions here that you should be able to use to swap them out.

Sweet! thanks

jacobtomlinson (approved):

PR #728, opened 2023-05-30
Update scheduler_info across cluster instances

Matt711:
I noticed that the scheduler_info is not updated accross all cluster instances. So if you scale one `KubeCluster` instance, the other instances will not show the new number of workers. This PR updates all instances of `KubeCluster.scheduler_info` when it's scaled.

### DIFF
--- a/dask_kubernetes/operator/kubecluster/kubecluster.py
+++ b/dask_kubernetes/operator/kubecluster/kubecluster.py
@@ -792,7 +792,8 @@ class KubeCluster(Cluster):
                 body={"spec": {"replicas": n}},
             )
             for instance in self._instances:
-                instance.scheduler_info = self.scheduler_info
+                if instance.name == self.name:
+                    instance.scheduler_info = self.scheduler_info
 
     def adapt(self, minimum=None, maximum=None):
         """Turn on adaptivity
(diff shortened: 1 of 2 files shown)


Matt711:
```python
c1 = KubeCluster(name="foo")
c2 = KubeCluster.from_name("foo")
print(c1)
print(c2)
```
```
KubeCluster(foo, 'tcp://localhost:54057', workers=0, threads=0, memory=0 B)
KubeCluster(foo, 'tcp://localhost:63098', workers=0, threads=0, memory=0 B)
```
```python
c2.scale(5)
print(c1)
print(c2)
```
```
KubeCluster(foo, 'tcp://localhost:54057', workers=5, threads=80, memory=77.43 GiB)
KubeCluster(foo, 'tcp://localhost:63098', workers=5, threads=80, memory=77.43 GiB)
```

jacobtomlinson (changes requested):
Good idea!

This implementation needs tweaking a little because `self._instances` is ALL instances of `KubeCluster`. So there probably needs to be something like `if instance.name == self.name` in there.

Also could you add a small test for this?

jacobtomlinson (approved):

PR #736, opened 2023-06-13
Reinstate test_job

### LINKED ISSUE #735: Reinstate job test
In #729 I added a skip to `dask_kubernetes/operator/controller/tests/test_controller.py::test_job` because it was failing intermittently.

We should get it passing again and remove the skip.

jacobtomlinson:
Closes #735

### DIFF
--- a/dask_kubernetes/operator/controller/tests/test_controller.py
+++ b/dask_kubernetes/operator/controller/tests/test_controller.py
@@ -475,8 +475,8 @@ async def test_job(k8s_cluster, kopf_runner, gen_job):
             # job_status = _get_job_status(k8s_cluster, ns)
             # _assert_final_job_status(job, job_status, "Successful")
 
-    assert "A DaskJob has been created" in runner.stdout
-    assert "Job succeeded, deleting Dask cluster." in runner.stdout
+    # assert "A DaskJob has been created" in runner.stdout
+    # assert "Job succeeded, deleting Dask cluster." in runner.stdout
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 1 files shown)


bstadlbauer:
@jacobtomlinson Do you happen to have a link to failed run? I've had a look at the job code and might have found a race condition (that I think we've also seen on our production clusters). I'm not too sure about the exact inner working of kopf/k8s though so unsure. 

One thing that I could see being problematic is that we create the `job-runner` pod here: 
https://github.com/dask/dask-kubernetes/blob/6970a8c75cef0c866de939ad04ee6ddce44e3458/dask_kubernetes/operator/controller/controller.py#L728-L732

Which would, once it becomes `"Running"`, set the `JobStatus` to `Running` here: 
https://github.com/dask/dask-kubernetes/blob/6970a8c75cef0c866de939ad04ee6ddce44e3458/dask_kubernetes/operator/controller/controller.py#L738-L761

In case this happens before the `daskjob_create_components()` finishes, it could be that the `JobsStatus` is overwritten to `ClusterCreated` here: 
https://github.com/dask/dask-kubernetes/blob/6970a8c75cef0c866de939ad04ee6ddce44e3458/dask_kubernetes/operator/controller/controller.py#L733-L735

jacobtomlinson:
Yeah that's interesting! Perhaps we need some check of before and after to ensure we only transition in the right direction. Do you have any interest in taking this over?

bstadlbauer:
I can, but at the moment I'm not sure whether I can commit to a timeline. I would like to get #731 finished first but could have a look at this one afterwards

bstadlbauer:
@jacobtomlinson I've thought a bit about this and I think there might be no good way to resolve the race condition between `ClusterCreated` and `Running`, as we would need an atomic patch of the form "Only update to Cluster Created if it is not Running" where the if cannot be realized in a patch afaik. 
What about removing the `ClusterCreated` state entirely? 🤔

jacobtomlinson:
> We could also create the job runner pod after the cluster moved to ClusterCreated 

I think that might make sense.

jacobtomlinson:
Closing in favour of #739

PR #738, opened 2023-06-22
Work around dask/distributed#7940

jacobtomlinson:
In https://github.com/dask/distributed/issues/7940 we are discussing how calling `self.asynchronous` now raises an exception if there is no loop running because it tried to access the `LoopRunner.loop` before the loop is created. We do this in `KubeCluster` to check whether we need to start the loop.

I've opened https://github.com/dask/distributed/pull/7941 to fix things but also replacing that check with a check suggested by @graingert which means we don't try and use `self.loop` until after the loop as been created anyway. 

This PR means we can move forward whether https://github.com/dask/distributed/pull/7941 gets accepted and we aren't waiting on a release, it should also unblock the current CI failures.


jacobtomlinson:
Looks like CI on main last passed with dask/distributed@74a1bcdc45ab8f1c4a33a8ae0a41de5cb2e5b832

PR #739, opened 2023-06-22
Fix job test

bstadlbauer:
The job tests have been disabled as they've been disabled in CI as they've been failing. This PR brings those back. More context in https://github.com/dask/dask-kubernetes/pull/736

@jacobtomlinson turns out they've been failing because the namespace has not been passed correctly to all `kubectl` commands - should be good now 👍 

As for the race condition in [this message](https://github.com/dask/dask-kubernetes/pull/736#issuecomment-1595819375), I'll open a separate PR.

Closes https://github.com/dask/dask-kubernetes/issues/735


jacobtomlinson (approved):
Awesome! Nice catch.

PR #745, opened 2023-06-27
Replace DaskJob pod with Kubernetes Job

### LINKED ISSUE #695: Add backoffLimit to DaskJobs
Would it be possible to add backoffLimit to DaskJobs? Kubernetes jobs have this argument so that the job is reported as failed only it the pod fails a certain number of times (see below). Could we add these to DaskJobs as well? I have been using this argument in jobs  because Dask sometimes "just hangs/crashes" in very long jobs and restarting the job  fixes that.

```yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: pi
spec:
  template:
    spec:
      containers:
      - name: pi
        image: perl:5.34.0
        command: ["perl",  "-Mbignum=bpi", "-wle", "print bpi(2000)"]
      restartPolicy: Never
  backoffLimit: 4
```

Matt711:
Closes #695. This PR replaces the Pod created from a DaskJob CRD with a Job. This allows us to use some of the existing functionality of K8s Jobs.

### DIFF
--- a/dask_kubernetes/operator/controller/controller.py
+++ b/dask_kubernetes/operator/controller/controller.py
@@ -162,11 +162,11 @@ def build_worker_deployment_spec(
     return deployment_spec
 
 
-def get_job_runner_pod_name(job_name):
+def get_job_runner_job_name(job_name):
     return f"{job_name}-runner"
 
 
-def build_job_pod_spec(job_name, cluster_name, namespace, spec, annotations, labels):
+def build_job_spec(job_name, cluster_name, namespace, spec, annotations, labels):
     labels.update(
         **{
             "dask.org/cluster-name": cluster_name,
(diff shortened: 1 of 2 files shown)


bstadlbauer:
Hey @Matt711! 
Would you mind if I try whether things still work with the `Flyte` `dask` plugin ([this thing](https://github.com/flyteorg/flyteplugins/blob/master/go/tasks/plugins/k8s/dask/dask.go)) once you're ready? 
It makes some assumptions around the job here, in particular the `job-runner` pod name, etc. So would be nice if this wouldn't be breaking the plugin

Matt711:
> Hey @Matt711! Would you mind if I try whether things still work with the `Flyte` `dask` plugin ([this thing](https://github.com/flyteorg/flyteplugins/blob/master/go/tasks/plugins/k8s/dask/dask.go)) once you're ready? It makes some assumptions around the job here, in particular the `job-runner` pod name, etc. So would be nice if this wouldn't be breaking the plugin
> 
> cc @jacobtomlinson

Yeah, no problem!

dask_kubernetes/operator/controller/controller.py
```diff
@@ -162,40 +162,46 @@ def build_worker_deployment_spec(
...
+    metadata = {
+        "name": get_job_runner_job_name(job_name),
+        "labels": labels,
+        "annotations": annotations,
+    }
+    job_spec = {
+        "apiVersion": "batch/v1",
+        "kind": "Job",
+        "metadata": metadata,
+        "spec": {
```
jacobtomlinson, line 186: I think we probably want our spec to replace this, rather than the `"spec"` within `"template"`.

That does mean we need to modify our CRDs to ref a `JobSpec` instead. That is a breaking change to our CRDs so we might need to think about how to bump the version there.

https://github.com/dask/dask-kubernetes/blob/2aa43eda392e8de899993ca019bbf1da11b6c303/dask_kubernetes/operator/customresources/daskjob.yaml#L65

PR #748, opened 2023-06-27
Add batch worker provisioning capability to the Operator

### LINKED ISSUE #733: Add batched worker provisioning to Dask Cluster spawning on Kubernetes
In Spark on Kubernetes, there is an option to set a batch size for the number of workers that are spun up simultaneously within a larger job. An example of this would be with 100 workers and batch size of 10, the distributed scheduler will spin up 10 gangs of 10 workers each sequentially to lessen load on the kubernetes scheduler as it provisions each gang. There is also an option to set a time delay, so instead of waiting for each batch to complete, instead it waits for n seconds to make the next request. 

It would be nice if Dask could implement similar functionality for batched worker provisioning. We find that in our use case, we start to have problems with the IAM provider in AWS above 40 simultaneous worker requests, and especially in large clusters of 100 or more. We have worked around it by adding sleeps or exponential backoffs to the container images in use, but if it could be adjusted at the cluster level, it would smooth out large cluster provisioning without needing kludges. 

Thanks - DH

Matt711:
Closes #733. This PR adds provides the ability for the operator deploy workers in batches of a given `size` with a cooldown `delay` between deployments.

### DIFF
--- a/dask_kubernetes/operator/controller/controller.py
+++ b/dask_kubernetes/operator/controller/controller.py
@@ -635,7 +635,7 @@ async def daskworkergroup_replica_update(
 
             SIZE = dask.config.get("kubernetes.controller.worker-allocation.batch-size")
             DELAY = dask.config.get("kubernetes.controller.worker-allocation.delay")
-            batch_size = min(workers_needed, SIZE)
+            batch_size = min(workers_needed, SIZE) if SIZE else workers_needed
             if workers_needed > 0:
                 for _ in range(batch_size):
                     data = build_worker_deployment_spec(
--- a/dask_kubernetes/operator/controller/controller.py
+++ b/dask_kubernetes/operator/controller/controller.py
@@ -3,7 +3,6 @@ from collections import defaultdict
 import time
 from contextlib import suppress
 from datetime import datetime
-import os
 from uuid import uuid4
 
 import aiohttp
@@ -22,6 +21,7 @@ from dask_kubernetes.common.auth import ClusterAuth
 from dask_kubernetes.common.networking import get_scheduler_address
 from distributed.core import rpc, clean_exception
 from distributed.protocol.pickle import dumps
+import dask.config
 
 _ANNOTATION_NAMESPACES_TO_IGNORE = (
     "kopf.zalando.org",
@@ -261,9 +261,6 @@ async def startup(settings: kopf.OperatorSettings, **kwargs):
     # The default timeout is 300s which is usually to long
     # https://kopf.readthedocs.io/en/latest/configuration/#networking-timeouts
     settings.networking.request_timeout = 10
-    global SIZE, DELAY
-    SIZE = int(os.getenv("WORKER_BATCH_SIZE"))
-    DELAY = int(os.getenv("WORKER_DELAY")[:-1])
 
 
 # There may be useful things for us to expose via the liveness probe
@@ -627,8 +624,6 @@ async def daskworkergroup_replica_update(
             )
             desired_workers = new
             workers_needed = desired_workers - current_workers
-            deployment_groups = workers_needed // SIZE
-            excess_workers = workers_needed % SIZE
             labels = _get_labels(meta)
             annotations = _get_annotations(meta)
             worker_spec = spec["worker"]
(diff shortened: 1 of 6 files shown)


dask_kubernetes/operator/controller/controller.py
```diff
@@ -260,6 +261,9 @@ async def startup(settings: kopf.OperatorSettings, **kwargs):
     # The default timeout is 300s which is usually to long
     # https://kopf.readthedocs.io/en/latest/configuration/#networking-timeouts
     settings.networking.request_timeout = 10
+    global SIZE, DELAY
+    SIZE = int(os.getenv("WORKER_BATCH_SIZE"))
+    DELAY = int(os.getenv("WORKER_DELAY")[:-1])
```
jacobtomlinson, line 266: It would be nice if we could use the Dask config system instead as that supports environment variables.

dask_kubernetes/operator/controller/controller.py
```diff
@@ -632,24 +638,65 @@ async def daskworkergroup_replica_update(
...
+                            )
+                            kopf.adopt(data, owner=body)
+                            kopf.label(data, labels=cluster_labels)
+                            await kubernetes.client.AppsV1Api(
+                                api_client
+                            ).create_namespaced_deployment(
+                                namespace=namespace,
+                                body=data,
+                            )
+                        await asyncio.sleep(DELAY)
```
jacobtomlinson, line 661: Instead of having this concept of deployment groups and batching things like this, we could do 

```python
for _ in range(min(workers_needed, SIZE)):
``` 

and then if we don't have enough workers we could 

```python
if workers_needed > SIZE:
    raise kopf.TemporaryError(
        "Add maximum number of workers for this batch but still need to create more workers, " 
        f"waiting for {DELAY} seconds before continuing.", 
        delay=DELAY)
```

This will cause `kopf` to stop executing `daskworkergroup_replica_update`, wait for the delay period and then call it again.

https://kopf.readthedocs.io/en/stable/errors/#temporary-errors

I think this would simplify things a lot here. We just let `kopf` keep calling `daskworkergroup_replica_update` until we have all the workers we requested. It also means if the autoscaler reduces the number of workers in the meantime we will stop scaling up.

dask_kubernetes/operator/deployment/helm/dask-kubernetes-operator/values.yaml
```diff
@@ -85,3 +85,7 @@ metrics:
         - dask.org/cluster-name
         - dask.org/workergroup-name
       metricRelabelings: [] # MetricRelabelConfigs to apply to samples before ingestion.
+
+workerAllocationSettings:
+  size: "5"
+  delay: 1s
```
jacobtomlinson, line 91: Can we default these to `nil` so that we don't batch by default? And add a comment to the end of the line explaining how to use each option.

jacobtomlinson (commented):
This looks like a great start. I have a few thoughts about the design.

If you added a `controller` section to [kubernetes.yaml](https://github.com/dask/dask-kubernetes/blob/main/dask_kubernetes/kubernetes.yaml) we could use the Dask config system instead of using global constants.

```yaml
kubernetes:
  ...
  controller:
    worker-allocation:
      batch-size: null
      delay: null
```

Then you can use the environment variables `DASK_KUBERNETES__CONTROLLER__WORKER_ALLOCATION__BATCH_SIZE` and `DASK_KUBERNETES__CONTROLLER__WORKER_ALLOCATION__DELAY` in the chart to set them and you can access the values in Python.

```python
import dask.config

...

dask.config.get("kubernetes.controller.worker-allocation.batch-size")
dask.config.get("kubernetes.controller.worker-allocation.delay")
```

(You might find [this utility](https://docs.dask.org/en/latest/configuration.html#conversion-utility) useful for converting between the different formats.)

I'm also keen to let `kopf` help us as much as possible. When batching we effectively want to call the scale workers handler but only scale up by the max batch size, then we want to wait for some period of time and then call the scale workers handler again until we have eventually scaled to our desired number. There is a way to do this by raising exceptions in `kopf`, see my comment below for more information.

dask_kubernetes/operator/controller/controller.py
```diff
@@ -631,8 +632,12 @@ async def daskworkergroup_replica_update(
                     annotations.update(**worker_spec["metadata"]["annotations"])
                 if "labels" in worker_spec["metadata"]:
                     labels.update(**worker_spec["metadata"]["labels"])
+
+            SIZE = dask.config.get("kubernetes.controller.worker-allocation.batch-size")
+            DELAY = dask.config.get("kubernetes.controller.worker-allocation.delay")
+            batch_size = min(workers_needed, SIZE)
```
jacobtomlinson, line 638: This will raise an exception if `SIZE` is `None` (the default)
```suggestion
            batch_size = min(workers_needed, SIZE) if SIZE else workers_needed
```

dask_kubernetes/operator/deployment/helm/dask-kubernetes-operator/values.yaml
```diff
@@ -85,3 +85,7 @@ metrics:
         - dask.org/cluster-name
         - dask.org/workergroup-name
       metricRelabelings: [] # MetricRelabelConfigs to apply to samples before ingestion.
+
+workerAllocationSettings:
```
jacobtomlinson, line 89: Maybe we just call this `workerAllocation` to be in line with `kubernetes.controller.worker-allocation` in the Dask config?

jacobtomlinson (commented):
Thanks for making these changes so quickly. I've added a couple of nits but otherwise this looks good.

The CI is still failing to I would suggest you start the controller locally in a terminal and manually create a cluster and watch the log output from the controller to see what is happening.

jacobtomlinson (approved):

PR #764, opened 2023-07-11
Use kr8s in networking functions

Matt711:
Seperate operator networking functions and use kr8s api in them.

### DIFF
--- a/dask_kubernetes/operator/controller/networking.py
+++ b/dask_kubernetes/operator/controller/networking.py
@@ -9,7 +9,6 @@ from distributed.core import rpc
 
 from dask_kubernetes.aiopykube.objects import Pod
 from dask_kubernetes.exceptions import CrashLoopBackOffError
-from kr8s.portforward import PortForward
 from kr8s.asyncio.objects import Service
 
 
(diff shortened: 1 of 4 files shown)


dask_kubernetes/operator/controller/networking.py
```diff
@@ -0,0 +1,176 @@
...
+        except socket.gaierror as e:
+            if i >= retries - 1:
+                raise e
+            await asyncio.sleep(0.5)
+
+
+async def port_forward_service(service, remote_port, local_port=None):
+    async with PortForward(service, remote_port, local_port) as port:
+        print(f"Forwarding to port {port}")
+        return port
```
Matt711, line 95: This should be `port.local_port`

jacobtomlinson, line 95: If you do `async with` the port forward will close when the function returns.

```suggestion
    pf = service.portforward(remote_port, local_port)
    port = await pf.start()
    print(f"Forwarding to port {port}")
    return port
```

We might also want to return `pf` and hold a reference to it somewhere but I'm not sure.

jacobtomlinson:
This was superseded

PR #768, opened 2023-07-14
Add smoke test for helm deployment

bstadlbauer:
This PR adds a smoke test that installs the helm chart into the local kind cluster to check whether the deployment is working correctly and a simple cluster can come up. This should minimize releasing broken versions.

### DIFF
--- a/dask_kubernetes/conftest.py
+++ b/dask_kubernetes/conftest.py
@@ -26,7 +26,9 @@ def kopf_runner(k8s_cluster):
 @pytest.fixture(scope="session")
 def docker_image():
     image_name = "dask-kubernetes:dev"
-    subprocess.run(["docker", "build", "-t", image_name, "./ci/"], check=True)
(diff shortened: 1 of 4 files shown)


bstadlbauer:
@jacobtomlinson Seems like the test setup is slow and I would expect this to be due to the docker build that's happening. 

Should I try to improve the build time there by leveraging caching?

bstadlbauer:
I've also added a second tests with `--set=rbac.create=false` which seems to be failing - is this expected?

jacobtomlinson:
> Seems like the test setup is slow and I would expect this to be due to the docker build that's happening.

I'm curious why the existing docker image fixture isn't being used here?

> I've also added a second tests with --set=rbac.create=false which seems to be failing - is this expected?

We should look into this. Do you want to xfail the test for now and open an issue to track it?

bstadlbauer:
Closing this as I've had not had time to get back to this

PR #770, opened 2023-07-14
Fix worker batching by ensuring env vars are passed correctly

### LINKED ISSUE #769: Helm chart failing to create workers with the new release.
Actually @jacobtomlinson it's still failing to create workers with the new release. 

I believe because [here](https://github.com/dask/dask-kubernetes/commit/35d42e5d74ded4c940d3ba4c23650da47c3e02c5#diff-fa353c03afcff8463787e5b6ffc87cf63d70b7d7030263085aacaf3cfc706ca6R38) the env variable must be a string. So need to cast [these](https://github.com/dask/dask-kubernetes/commit/35d42e5d74ded4c940d3ba4c23650da47c3e02c5#diff-7cb42253590c235aed32d49bd4ea3cebbbbec7877213a4969e9f4a880d47c8f4R637-R638)

_Originally posted by @dyu-bot in https://github.com/dask/dask-kubernetes/issues/766#issuecomment-1634771138_

jacobtomlinson:
Updated the helm chart to ensure we only pass the worker batching env vars if they are set, and set them as a string because env vars expect that.

Then updated the controller to ensure we cast those back to integers.

Closes #769

### DIFF
--- a/dask_kubernetes/operator/controller/controller.py
+++ b/dask_kubernetes/operator/controller/controller.py
@@ -634,8 +634,12 @@ async def daskworkergroup_replica_update(
                 if "labels" in worker_spec["metadata"]:
                     labels.update(**worker_spec["metadata"]["labels"])
 
-            SIZE = dask.config.get("kubernetes.controller.worker-allocation.batch-size")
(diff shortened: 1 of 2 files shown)


jacobtomlinson:
cc @dyu-bot

PR #772, opened 2023-07-14
Tell pytest to fail fast

jacobtomlinson:
We keep seeing hanging CI and it's hard to debug what is going on. Many of these runs have many test failures before locking up. This PR adds the `-x` flag to pytest which will cause it to exit after the first failure. Hopefully this should help get to the bottom of the hangs.


jacobtomlinson:
Not working perfectly but is an improvement so going to merge

Issue #776, opened 2023-07-21
Run Dask Operator container as non-root


droctothorpe:
Many clusters prohibit running containers as root to optimize security. The Dask operator container runs as root. Any particular reason why? Would creating and switching to user 1000 break the application code in any way?

jacobtomlinson:
No reason at all. I think we would just need to tweak the `Dockerfile` and `deployment.yaml` to make this the default.

jsonmp-k8:
The deployment has the option to pass runasuser param in the values file. The Dockerfile needs to be adjusted to make sure the image has the right user.

droctothorpe:
Closed by https://github.com/dask/dask-kubernetes/pull/777.

PR #777, opened 2023-07-24
Run operator container as non-root

droctothorpe:
This PR updates the operator container `Dockerfile` and operator deployment manifest (via the values file) to run as non-root, i.e. user 1000 (username `toor`). 

Relevant [issue](https://github.com/dask/dask-kubernetes/issues/776).

I confirmed the image builds, shelled into the running container, and confirmed that I can execute the `kopf` binary as the `dask` user.

### DIFF
--- a/dask_kubernetes/operator/deployment/Dockerfile
+++ b/dask_kubernetes/operator/deployment/Dockerfile
@@ -10,6 +10,12 @@ WORKDIR /src/dask_kubernetes
 # Install dependencies
 RUN pip install .
 
+# Add non-root user
+RUN useradd -s /bin/bash dask
+
+# Run container as non-root
+USER 1000
+
 # Start operator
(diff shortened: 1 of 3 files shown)


droctothorpe:
Hmm,, `SSL: CERTIFICATE_VERIFY_FAILED`.

droctothorpe:
> Thanks for raising this. CI failures are unrelated to this change.

Phewph. I spent way too long making sure it wasn't PEBCAK, heh.

> One small nit, see my comment below.

Not a nit at all. I was surprised to learn that there isn't a standardized default name for the first non-root user. The closest thing is `toor`, which is used in a couple of distros, but it's far from being a standard. Updated to `dask`.

droctothorpe:
Thanks for the rapid review and merge, @jacobtomlinson! 🙏

PR #780, opened 2023-07-26
Switch pytest-asyncio to use anyio

jacobtomlinson:
In https://github.com/dask/distributed/issues/7984 @fjetter and @graingert suggested moving away from `pytest-asyncio`.

This PR updates all tests to use `anyio` but with just the `asyncio` backend.


jacobtomlinson:
Well that was pretty successful. Thanks for the tips.

PR #783, opened 2023-07-26
Fix extensibility section link on kubecluster.rst

### LINKED ISSUE #782: Extensibility link broken on kubecluster.rst DOCS
Hi there, I was going over some docs and I found that in this section of the docs https://github.com/dask/dask-kubernetes/blob/main/doc/source/kubecluster.rst#docker-images the link to "Extensibility Section" leads to a 404. 

When I check the source docs code it points to this link: https://docs.dask.org/en/latest/setup/docker.html#extensibility

I think it should point to this page: https://docs.dask.org/en/latest/deploying-docker.html#extensibility

Opening a follow-up PR with the fix, if it's the wrong page, happy to fix.

ncclementi:
Closes #782


jacobtomlinson (approved):
Thanks @ncclementi!

PR #801, opened 2023-08-10
Convert oid expiration timestamp to datetime.datetime

droctothorpe:
After enabling refresh tokens on our oidc backend (shout out to @harrisonfritz),  dask-kubernetes still failed. I narrowed down the problem to the fact that `parse_rfc3339` is not equipped to handle an `expiration_timestamp` in epoch string format, which is what the OIDC protocol spec specifies (shout out to @DWSR for pointing us to the relevant [docs](https://openid.net/specs/openid-connect-core-1_0.html#IDToken)):

> exp
REQUIRED. Expiration time on or after which the ID Token MUST NOT be accepted for processing. The processing of this parameter requires that the current date/time MUST be before the expiration date/time listed in the value. Implementers MAY provide for some small leeway, usually no more than a few minutes, to account for clock skew. **Its value is a JSON number representing the number of seconds from 1970-01-01T0:0:0Z as measured in UTC until the date/time.** See [RFC 3339](https://openid.net/specs/openid-connect-core-1_0.html#RFC3339) [RFC3339] for details regarding date/times in general and UTC in particular.

Emphasis mine. I manually validated that this works on our end. Not sure why no one has run into it before. Perhaps the oidc auth path is not widely adopted or most people are invoking it in-cluster (where the service account token is consumed).

### DIFF
--- a/dask_kubernetes/common/auth.py
+++ b/dask_kubernetes/common/auth.py
@@ -162,7 +162,9 @@ class AutoRefreshKubeConfigLoader(KubeConfigLoader):
             self.token_expire_ts <= datetime.datetime.now(tz=tzUTC)
         ):
             await self._refresh_oidc(provider)
-            expires = self.extract_oid_expiration_from_provider(provider=provider)
+            expires = datetime.datetime.fromtimestamp(
+                self.extract_oid_expiration_from_provider(provider=provider)
+            )
 
             await self.create_refresh_task_from_expiration_timestamp(
                 expiration_timestamp=expires


jacobtomlinson:
Thanks for raising this. The code you've changed here is actually part of the classic code that is in the process of being stripped out so this will be removed again in the very near future. However, given that it's such a small fix I'm happy to merge it for now.

> Not sure why no one has run into it before. Perhaps the oidc auth path is not widely adopted or most people are invoking it in-cluster (where the service account token is consumed).

I think you're right.

jacobtomlinson (approved):

droctothorpe:
Thanks, @jacobtomlinson! This clears the last obstacle for us to adopt the operator pattern. Really appreciate the merge.

jacobtomlinson:
That's great! I'll get a patch release out with this in.

Issue #806, opened 2023-08-25
Dask can not connect to scheduler using NodePort when node has multiple addresses


dbalabka:
**Describe the issue**:

If we try to create a scheduler with service type `NodePort`, dask will try to connecting to scheduler forever:
```
╭─────────────────── Creating KubeCluster 'dmitryb-cluster' ────────────────────╮
│                                                                              │
│   DaskCluster                                                      Running   │
│   Scheduler Pod                                                    Running   │
│   Scheduler Service                                                Created   │
│   Default Worker Group                                             Created   │
│                                                                              │
│ ⠼ Connecting to scheduler                                                    │
╰──────────────────────────────────────────────────────────────────────────────╯
```
It seems that `get_external_address_for_scheduler_service` returns internal port `tcp://172.16.18.191:8786`, instead of which is being mapped in services:
```
❯ kubectl get services
NAME                       TYPE       CLUSTER-IP      EXTERNAL-IP   PORT(S)                         AGE
dmitryb-cluster-scheduler   NodePort   10.105.154.53   <none>        8786:32167/TCP,8787:30729/TCP   8m45s
```
https://github.com/dask/dask-kubernetes/blob/7c09b57b05bfccc3d69ddc553d0ddf73a8ef063f/dask_kubernetes/common/networking.py#L86

As a solution, we have to use `node_port` instead of `port` in case of using `NodePort` service:
https://github.com/dask/dask-kubernetes/blob/7c09b57b05bfccc3d69ddc553d0ddf73a8ef063f/dask_kubernetes/common/networking.py#L60

So, the correct connection URI should be: `tcp://172.16.18.191:32167`

**Minimal Complete Verifiable Example**:

```python
from dask_kubernetes.operator import KubeCluster, make_cluster_spec

spec = make_cluster_spec(
    name="dmitryb-cluser", 
    image='ghcr.io/dask/dask:2023.8.1-py3.10',
    n_workers=1, 
    scheduler_service_type="NodePort",
)
 
cluster = KubeCluster(
    namespace="dask-operator", 
    custom_cluster_spec=spec, 
)
```

**Environment**:

```
dask-kubernetes = "~2023.8.0"
dask = "~2023.8.0"
```

jacobtomlinson:
I expect the bug is on this line. If your node has two addresses only the first is returned. We will need some way to figure out which is public and which is private.

[...]

PR #808, opened 2023-08-29
Fix port number when use NodePort (#806)

dbalabka:
Here is an explanation why we have to use `node_port` instead of `port` when service type is `NodePort`: https://stackoverflow.com/a/41510604/983577


dbalabka (commented):
Fix port number when use nodeport (#806)

jacobtomlinson (approved):
This looks great many thanks!

PR #811, opened 2023-09-08
Add more deprecation warnings ahead of removing old code

jacobtomlinson:
For some time we've had a deprecation warning when users do `from dask_kubernetes import KubeCluster`. 

Now that the operator is more stable and is fully migrated to kr8s it's time for the classic `KubeCluster` and the `HelmCluster` to go away. This PR adds a more forceful deprecation warning to those classes ahead of full removal.

### DIFF
--- a/dask_kubernetes/__init__.py
+++ b/dask_kubernetes/__init__.py
@@ -23,15 +23,6 @@ del get_versions
 
 def __getattr__(name):
     if name == "KubeCluster":
-        warn(
-            "It looks like you are using the classic implementation of KubeCluster. "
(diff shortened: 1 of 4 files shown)


karlkovaciny:
If HelmCluster is deprecated, should it be removed from the docs at https://kubernetes.dask.org/en/latest/helmcluster.html and [docs.dask.org](https://docs.dask.org/en/stable/deploying-kubernetes.html?highlight=helmcluster#single-cluster-helm-chart)?

jacobtomlinson:
@karlkovaciny yes it will be. Although the chart will still exist, but the Python tooling around it will be going away.

PR #813, opened 2023-09-11
Add type hints to KubeCluster constructor

droctothorpe:
This PR adds type hints to the `operator.KubeCluster` constructor. I ran `isort` since there were a few additional imports. Hope you don't mind the nit, but I also corrected two minor typos in the docstring. 

There's some type hint redundancy between the parameters and the docstring. I usually exclude type hints from docstrings to avoid that, but TBH I'm not sure if that's idiomatic. [asyncio](https://github.com/python/cpython/blob/0b6b05391bcc5893d2b71032cb704995696ecd0b/Lib/asyncio/staggered.py#L16) in the stdlib seems to adhere to that pattern fwiw. Happy to strip the type hints from the docstring if you think it makes sense.

If this PR seems reasonable and gets merged, happy to add type hints elsewhere in the codebase and run mypy and stuff.

Lastly, it might make sense to add [isort to the pre-commit hooks](https://pycqa.github.io/isort/docs/configuration/pre-commit.html). Happy to submit that in a discrete PR.

### DIFF
--- a/dask_kubernetes/operator/kubecluster/kubecluster.py
+++ b/dask_kubernetes/operator/kubecluster/kubecluster.py
@@ -2,48 +2,43 @@ from __future__ import annotations
 
 import asyncio
 import atexit
-from contextlib import suppress
-from enum import Enum
 import getpass
 import logging
 import os
 import time
-from typing import ClassVar
+import uuid
 import warnings
 import weakref
-import uuid
-
-from rich import box
-from rich.live import Live
-from rich.table import Table
-from rich.console import Group
-from rich.panel import Panel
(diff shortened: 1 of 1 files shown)


jacobtomlinson (approved):
This looks great thanks @droctothorpe.

I've been using `ruff` in `kr8s` for a while to handle the `black`/`isort`/`flake8` stuff and I've been very happy with it, maybe we should introduce that here.

droctothorpe:
Nice! Haven't tried ruff yet. Will try adding it to the pre-commit hooks and seeing what happens when I run `pre-commit run --all-files`.

PR #815, opened 2023-09-11
Add ruff to pre-commit and run against all files

droctothorpe:
This PR adds [ruff](https://github.com/astral-sh/ruff) to the pre-commit hooks and resolves linting violations listed by an execution of `pre-commit run --all-files`. A few errors are ignored, namely:

- [F401](https://www.flake8rules.com/rules/F401.html) since it complains about unused imports in all of the `__init__.py` files.
- [E402](https://www.flake8rules.com/rules/E402.html) since the auto-generated `conf.py` begins with comments rather than imports.
- [E501](https://www.flake8rules.com/rules/E501.html) since there are so many violations (around 70), particularly in `controller.py`. If this is enforced, it might make sense to handle the remediation in a discrete PR. 

Issue: https://github.com/dask/dask-kubernetes/issues/814

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,15 +1,15 @@
 repos:
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.0.287
+    hooks:
+      - id: ruff
+        args: [--fix, --exit-non-zero-on-fix]
   - repo: https://github.com/psf/black
     rev: 22.3.0
     hooks:
       - id: black
         language_version: python3
         exclude: versioneer.py
-  - repo: https://github.com/pycqa/flake8
-    rev: 3.9.2
-    hooks:
-      - id: flake8
-        language_version: python3
   - repo: https://github.com/rapidsai/frigate/
     rev: v0.4.0 #  pre-commit autoupdate  - to keep the version up to date
     hooks:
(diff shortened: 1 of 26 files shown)


jacobtomlinson (commented):
Thanks for this.

It looks like this doesn't take into account the existing flake8 config, so that will need to be ported over. I expect that will at least solve the line length issues.

https://github.com/dask/dask-kubernetes/blob/92714da5785709726f85c4c6ec92451f5c23ad04/setup.cfg#L1-L28

Also for the unused imports I would add `#noqa` comments in the file instead of disabling the check.

Also it looks like the import sorting has introduced a circular import and is causing CI to fail.

jrbourbeau (commented):
Nice. @jacobtomlinson I'd be curious to hear what the `ruff` experience is like here at some point. Might be worth doing something similar in other projects if you think it improves the developer experience (xref https://github.com/dask/dask/issues/10196)

droctothorpe:
Congrats on being on the front page of HN, @jacobtomlinson!

jacobtomlinson (approved):
Thanks for the nudge, this generally looks great.

I'm always kinda sad about how `pytest` get mixed into the list of imports, I usually prefer it to be the 1st import line in a test file with a blank line below it. But that's the point of these tools I guess to remove these personal preferences in favour of consistency.

Issue #817, opened 2023-09-16
Cleanup pending pods on scale down


BitTheByte:
Currently, the operator retires workers using the HTTP or RPC APIs however those only control the connected dask workers, the operator should take in count pods that are in a pending state as those will cause useless a Kubernetes cluster scale-up then connect to dask and get retired thus pending pods should be removed before they enter running state

jacobtomlinson:
Agreed.

We could add a check here for any Pods that aren't in a Running phase and delete those before calling `retire_workers` (if that's even necessary any more).

https://github.com/dask/dask-kubernetes/blob/92714da5785709726f85c4c6ec92451f5c23ad04/dask_kubernetes/operator/controller/controller.py#L599-L600

BitTheByte:
Looks good to me, but I really don't know if we should also subtract pending workers from the number of workers passed to `retire_workers`

PR #820, opened 2023-09-18
Retire pending workers

BitTheByte:
Based on discussion at https://github.com/dask/dask-kubernetes/issues/817

### DIFF
--- a/dask_kubernetes/operator/controller/controller.py
+++ b/dask_kubernetes/operator/controller/controller.py
@@ -609,7 +609,9 @@ async def daskworkergroup_replica_update(
         if workers_needed < 0:
             worker_ids = await retire_workers(
                 n_workers=-workers_needed,
-                scheduler_service_name=SCHEDULER_NAME_TEMPLATE.format(cluster_name),
+                scheduler_service_name=SCHEDULER_NAME_TEMPLATE.format(
+                    cluster_name=cluster_name
+                ),
                 worker_group_name=name,
                 namespace=namespace,
                 logger=logger,


BitTheByte:
Everything looks good at least for now. I deployed the change to our production cluster and will provide updates if needed. Hopefully, everything works well 😄

BitTheByte:
Surprisingly I had the most stable run more than ever. One note to mention is if a pod is restarting which means it's `deployment` in an unready state a small possibility might happen:

1. The operator gets the unready state
2. pod restarts fast enough and starts executing work
3. The operator kills the pod mid-run

So I was thinking of a way to only execute the logic only on deployment that have pods in a `pending` state however I can't find a way to do that using `kr8s`

BitTheByte:
Done! now the operator takes actions based on pending pods rather than the deployments

dask_kubernetes/operator/controller/controller.py
```diff
@@ -597,17 +606,51 @@ async def daskworkergroup_replica_update(
...
+                    continue
+
+                worker_pod = kube_pods[0]
+                if (
+                    worker_pod.status.phase
+                    not in [
+                        "ContainerCreating",
+                        "Initializing",
+                        "Terminating",
+                        "Running",
```
jacobtomlinson, line 630: I think you might be confusing the Pod phases with the status that `kubectl` displays. The possible values for phase are `Pending`, `Running`, `Succeeded`, `Failed` and `Unknown`.

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#podstatus-v1-core

dask_kubernetes/operator/controller/controller.py
```diff
@@ -549,15 +560,13 @@ async def daskworkergroup_replica_update(
...
-        current_workers = len(
-            await kr8s.asyncio.get(
-                "deployments",
-                namespace=namespace,
-                label_selector={"dask.org/workergroup-name": name},
-            )
+        current_workers = await kr8s.asyncio.get(
+            "deployments",
+            namespace=namespace,
+            label_selector={"dask.org/workergroup-name": name},
```
jacobtomlinson, line 566: Instead of listing all the worker deployments maybe we just want to list the Pods and use the field selector to get the ones that are Pending?

```python
current_workers = await kr8s.asyncio.get(
    "pods", 
    namespace=namespace, 
    label_selector={"dask.org/workergroup-name": name},
    field_selector="status.phase=Pending",
)
```

Then use the `dask.org/object-name` label (which I would maybe rename to `dask.org/component-name`) to delete the Deployment object later.

jacobtomlinson (commented):
Thanks for taking the time here. A few thoughts.

dask_kubernetes/operator/controller/controller.py
BitTheByte: Yup, you're correct I've been debugging this using k9s

dask_kubernetes/operator/controller/controller.py
BitTheByte: > Instead of listing all the worker Deployments maybe we just want to list the Pods and use the field selector to get the ones that are Pending?

Couldn't that cause a time of check time of use bug?

Changing `dask.org/object-name` -> `dask.org/component-name` seems good to me

jacobtomlinson: > Couldn't that cause a time of check time of use bug?

Only if you delete the Pods sequentially. You could delete them concurrently in a task group.

```python
async with anyio.create_task_group() as tg:
    for pod in pods:
        tg.call_soon(pod.delete)
```

BitTheByte:
Sorry for the long delay, unfortunately, I don't have much time to work on this. meanwhile, I'll close this PR to allow someone else to finish this

BitTheByte:
@jacobtomlinson I believe this is ready to merge

jacobtomlinson (commented):
The CI appears to be hanging as a result of these changes. This needs looking into before we can merge.

BitTheByte:
It is something related to building some go code. I tried to check what is wrong but didn't figure it out. it appears to be something related to the CI itself as I see most of the PRs are falling too.

BitTheByte:
@jacobtomlinson If possible can you suggest any solution to solve this issue?

jacobtomlinson:
Thanks for being patient here. I've nudged the CI back into a happy state and pulled main into this PR. So hopefully now any failures will only be related to this PR.

jacobtomlinson:
Ok the tests are still failing here so they are certainly hanging due to the changes in this PR.

jacobtomlinson:
Sure but I expect the CI is timing out because the tests are hanging. The tests on `main` do not hang, which suggests the changes here cause the tests to hang.

PR #823, opened 2023-09-27
Clarify name parameter docstring

droctothorpe:
When `custom_cluster_spec` is passed, `name` is no longer required. This PR updates the docstring to avoid confusion on this point.

### DIFF
--- a/dask_kubernetes/operator/kubecluster/kubecluster.py
+++ b/dask_kubernetes/operator/kubecluster/kubecluster.py
@@ -60,8 +60,10 @@ class KubeCluster(Cluster):
 
     Parameters
     ----------
-    name: str (required)
-        Name given the Dask cluster.
+    name: str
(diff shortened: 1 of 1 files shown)


dask_kubernetes/operator/kubecluster/kubecluster.py
```diff
@@ -60,8 +60,9 @@ class KubeCluster(Cluster):
 
     Parameters
     ----------
-    name: str (required)
-        Name given the Dask cluster.
+    name: str (optional)
```
jacobtomlinson, line 63:
```suggestion
    name: str
```

Setting `name` is required in the vast majority of uses given that `custom_cluster_spec` is a power-user feature. So I'm hesitant to explicitly mark this as `(optional)`. Maybe we just shouldn't specify and leave folks to read the description?

jacobtomlinson (approved):

PR #824, opened 2023-09-27
Avoid parsing resources when None

droctothorpe:
This PR fixes a small bug that I ran into trying to run the SDK from my local machine. If resources were not passed and there was no `~/.config/dask/kubernetes.yaml` file, it would try to parse `self.resources` when `self.resources` was set to its default, `None`.

### DIFF
--- a/dask_kubernetes/operator/kubecluster/kubecluster.py
+++ b/dask_kubernetes/operator/kubecluster/kubecluster.py
@@ -223,7 +223,7 @@ class KubeCluster(Cluster):
             "kubernetes.idle-timeout", override_with=idle_timeout
         )
 
-        if self._custom_cluster_spec:
+        if self._custom_cluster_spec is not None:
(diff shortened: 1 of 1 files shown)


dask_kubernetes/operator/kubecluster/kubecluster.py
```diff
@@ -232,21 +232,22 @@ def __init__(
...
+                ), f"resources must be dict type, found {type(resources)}"
+                for field in self.resources:
+                    if field in ("limits", "requests"):
+                        assert isinstance(
+                            self.resources[field], dict
+                        ), f"key of '{field}' must be dict type"
+                    else:
+                        raise ValueError(f"resources has unknown field '{field}'")
+            except AssertionError as e:
+                raise TypeError from e
```
droctothorpe, line 250: The only reason this is showing up in the diff is because it's indented.

jacobtomlinson (approved):
Awesome thanks

PR #833, opened 2023-10-12
Default resource_timeout to 60

droctothorpe:
This brings it into alignment with the docstring:

```
    resource_timeout: int (optional)
        Time in seconds to wait for Kubernetes resources to enter their expected state.
        Example: If the ``DaskCluster`` resource that gets created isn't moved into a known ``status.phase``
        by the controller then it is likely the controller isn't running or is malfunctioning and we time
        out and clean up with a useful error.
        Example 2: If the scheduler Pod enters a ``CrashBackoffLoop`` state for longer than this timeout we
        give up with a useful error.
        Defaults to ``60`` seconds.
```

It also resolves an exception I ran into on this line when it tried to add a float to None.

### DIFF
--- a/dask_kubernetes/operator/kubecluster/kubecluster.py
+++ b/dask_kubernetes/operator/kubecluster/kubecluster.py
@@ -168,7 +168,7 @@ class KubeCluster(Cluster):
         create_mode: Optional[CreateMode] = None,
         shutdown_on_close: Optional[bool] = None,
         idle_timeout: Optional[int] = None,
-        resource_timeout: Optional[int] = None,
+        resource_timeout: int = 60,
         scheduler_service_type: Optional[str] = None,
         custom_cluster_spec: Optional[str | dict] = None,
         scheduler_forward_port: Optional[int] = None,
--- a/doc/source/kubecluster_migrating.rst
+++ b/doc/source/kubecluster_migrating.rst
@@ -21,7 +21,7 @@ Here are some reasons why we decided to make this change:
 - Simpler Python API
 - More powerful YAML API
   - Create, scale and delete clusters with ``kubectl``
-- Detatch and reattactch from running clusters
+- Detach and reattach from running clusters
 - New resource types like ``DaskJob``
 - Multiple worker groups
 - Autoscaling handled by the controller and not the cluster manager


droctothorpe:
Added a tiny documentation fix.

jacobtomlinson (commented):
The default is set in the Dask config and we intentionally don't set it in the function. This is commonly the way Dask handles things. So I don't think we can accept this PR as the default will then be defined in two places.

https://github.com/dask/dask-kubernetes/blob/cb209c9477273e6debdf14e630036421d9f633c3/dask_kubernetes/kubernetes.yaml#L21

It would be useful to know more about the bug you're running into.

droctothorpe:
Sure thing. The corner case I ran into is when there is no Dask config available, which is probably not unusual, especially when users are using dask-kubernetes on their local machine for the first time (as opposed to a curated image). 

Basically, if the dask config is not available, [this](https://github.com/dask/dask-kubernetes/blob/main/dask_kubernetes/operator/kubecluster/kubecluster.py#L209) line returns `None`. I can replicate independently:

```python
In [5]: import dask

In [6]: dask.config.get("kubernetes.resource-timeout")
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[6], line 1
----> 1 dask.config.get("kubernetes.resource-timeout")

File /opt/homebrew/Caskroom/miniconda/base/envs/3.9/lib/python3.9/site-packages/dask/config.py:568, in get(key, default, config, override_with)
    566 k = canonical_name(k, result)
    567 try:
--> 568     result = result[k]
    569 except (TypeError, IndexError, KeyError):
    570     if default is no_default:

KeyError: 'kubernetes'
```

If `self._resource_timeout` is set to `None`, [this](https://github.com/dask/dask-kubernetes/blob/main/dask_kubernetes/operator/kubecluster/kubecluster.py#L459) line fails because it's trying to concatenate a `float` to `None`.

A few simple solutions come to mind. Maybe we can specify a default in the constructor but if `dask.config.get("kubernetes.resource-timeout")` returns a value, it overrides the default? Alternatively, maybe we  can override `None` if `dask.config.get("kubernetes.resource-timeout")` returns `None`? I think the former might make more sense but there is some implicit config sequencing (dask.config takes precedence over the default) which might be odd. Curious to hear your thoughts. Happy to implement whatever solution you prefer.

jacobtomlinson:
You need to import `dask_kubernetes` too to get the defaults to load in.

```python
In [1]: import dask

In [2]: import dask_kubernetes

In [3]: dask.config.get("kubernetes.resource-timeout")
Out[3]: 60
```

droctothorpe:
Ah thank you! My dask import is happening in my Kubeflow pipeline function, not in the namespace where I'm using dask-kubernetes. Much appreciated.

PR #837, opened 2023-10-19
Prepend user env vars

jacobtomlinson:
In #836 we see the default scheduler address is overriding the one supplied by the user. This happens because we extend the list of user-supplied env vars with the defaults, which Kubernetes then interprets as overriding.

This PR prepends the defaults instead so that the user-supplied env vars with the same name take precedent.


creste:
Hi @jacobtomlinson - Do you have a rough estimate of when this PR will be merged into main?

jacobtomlinson:
I think this may have been superseded by #874. I'm going to close this out but feel free to comment here if there is more that needs doing.

PR #840, opened 2023-11-03
Add quiet parameter to KubeCluster

droctothorpe:
This parameter is available in the parent class but not exposed in `KubeCluster`.

### DIFF
--- a/dask_kubernetes/operator/kubecluster/kubecluster.py
+++ b/dask_kubernetes/operator/kubecluster/kubecluster.py
@@ -115,6 +115,9 @@ class KubeCluster(Cluster):
         used to create the cluster instead of generating one from the other keyword arguments.
     scheduler_forward_port: int (optional)
         The port to use when forwarding the scheduler dashboard. Will utilize a random port by default
+    quiet: bool
+        If enabled, suppress rich console output.
+        Defaults to ``False``.
 
     **kwargs: dict
         Additional keyword arguments to pass to LocalCluster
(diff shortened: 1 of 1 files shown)


dask_kubernetes/operator/kubecluster/kubecluster.py
```diff
@@ -224,6 +228,7 @@ def __init__(
         self.idle_timeout = dask.config.get(
             "kubernetes.idle-timeout", override_with=idle_timeout
         )
+        self.quiet = quiet
```
jacobtomlinson, line 231: I think we should remove this and instead pass `quiet` along to the `super().__init__(..., quiet=quiet)` further down.

dask_kubernetes/operator/kubecluster/kubecluster.py
```diff
@@ -115,6 +115,9 @@ class KubeCluster(Cluster):
         used to create the cluster instead of generating one from the other keyword arguments.
     scheduler_forward_port: int (optional)
         The port to use when forwarding the scheduler dashboard. Will utilize a random port by default
+    quiet: bool
+        If enabled, suppress rich console output.
```
jacobtomlinson, line 119: I think we should be clear that this suppresses all printed output from the cluster, not just the rich console. I acknowledge that the only output we have today is the rich console but that may not always be true.

dask_kubernetes/operator/kubecluster/kubecluster.py
droctothorpe: Duh, heh. Not sure why I didn't think of that. Done.

jacobtomlinson (approved):
Awesome thanks!

droctothorpe:
Parsing the test failures. I don't think they're related to this PR.

jacobtomlinson:
Sorry for the delay in unblocking the CI here. Things should be good to go now.

droctothorpe:
Thanks, @jacobtomlinson!

Issue #842, opened 2023-11-22
Ability to add different scheduler address to workers outside of standard format


s-ducks:
for some reason our k8s clusters don't use the `cluster.local` dns suffix and we would like to be able to add our own dns suffix to allow us to work around this. 
I have tried setting this with the env var `DASK_SCHEDULER_ADDRESS`, but this [bug](https://github.com/dask/dask-kubernetes/issues/841) prevents that from working

jacobtomlinson:
I think this is a duplicate of #836  which will be closed by #837

s-ducks:
yup agreed - have put a PR in: https://github.com/dask/dask-kubernetes/pull/843
lmk what you think :)

wmay:
I just tried this with the most recent versions of the helm chart and dask kubernetes, and can confirm that it works now.

Issue #841, opened 2023-11-22
Env var duplication


s-ducks:
Environment variables `DASK_SCHEDULER_ADDRESS` and `DASK_WORKER_NAME` are duplicated on each successive worker pod created.

**Minimal Complete Verifiable Example**:

```python
from dask_kubernetes.operator import KubeCluster
cluster - KubeCluster(name="test", n_workers=3, namespace="external")
```

```sh
kubectl get pods -n external
NAME                                              READY   STATUS    RESTARTS   AGE
test-default-worker-07336f78ed-6495fb4f86-ltwvb   1/1     Running   0          10m
test-default-worker-babb2f1fed-7c6db6cdfc-lpfmf   1/1     Running   0          10m
test-default-worker-e2e95b0bb4-59f647889d-dpw7t   1/1     Running   0          10m
test-scheduler-89b645d66-hzdgr                    1/1     Running   0          10m
```
then:
```sh
kubectl get pod -n external -o yaml test-default-worker-e2e95b0bb4-59f647889d-dpw7t | grep -e DASK_SCHEDULER_ADDRESS -e DASK_WORKER_NAME -e value
    - $(DASK_WORKER_NAME)
    - name: DASK_WORKER_NAME
      value: test-default-worker-babb2f1fed
    - name: DASK_SCHEDULER_ADDRESS
      value: tcp://test-scheduler.external.svc.cluster.local:8786
    - name: DASK_WORKER_NAME
      value: test-default-worker-e2e95b0bb4
    - name: DASK_SCHEDULER_ADDRESS
      value: tcp://test-scheduler.external.svc.cluster.local:8786
```
we can see that this has the address and name of one of the other worker pods and a duplicated scheduler address. This does not effect the actual set values in the container however:

```sh
kubectl -n external exec test-default-worker-e2e95b0bb4-59f647889d-dpw7t -- env | grep DASK
DASK_VERSION=2023.11.0
DASK_WORKER_NAME=test-default-worker-e2e95b0bb4
DASK_SCHEDULER_ADDRESS=tcp://test-scheduler.external.svc.cluster.local:8786
```

**Environment**:

- Dask version: 2023.11.0
- Python version: 3.10.12
- Operating System: Mac 
- Install method (conda, pip, source): pip

efeboyaci:
Something related to this issue that has not been resolved.
If env is provided in the daskworkergroup, the env is duplicated in the deployment config. I believe the last entry applies to the pod. 
[...]

PR #843, opened 2023-11-27
fix duplicate env vars + ability to add custom address for scheduler

s-ducks:
fixes:
- https://github.com/dask/dask-kubernetes/issues/841
- https://github.com/dask/dask-kubernetes/issues/842 which duplicates:
- https://github.com/dask/dask-kubernetes/issues/836
alternative to:
- https://github.com/dask/dask-kubernetes/pull/837

_why this approach_? 
I think it would be good to "prefer" the user set env vars (or env vars set from elsewhere), which is what the addition of `_consolidate_env_vars`  does


jacobtomlinson:
I think this has been superseded. If I'm wrong please comment and I can reopen.

Issue #684, opened 2023-03-23
`ResourceQuota` handling


bstadlbauer:
**Describe the issue**:
Whist running the operator in production, we've recently noticed that sometimes a `DaskJob` is stuck in the `"JobCreated"` state. Upon further inspection, we've seen that the operator is not able to schedule one of the pods in most cases. The reason behind this is that we've got namespace-wise `ResourceQuota`s in place, which are then used up by the worker pods, leaving no space for the runner pod to launch. 

We're seeing an error similar to the following one:

```
kubernetes_asyncio.client.exceptions.ApiException: (403)
Reason: Forbidden
HTTP response headers: <CIMultiDictProxy('Audit-Id': '38a12602-5719-4d7e-bb02-325bd12dd85e', 'Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'X-Kubernetes-Pf-Flowschema-Uid': 'ec346534-3e72-417d-88b8-de9cd45795e0', 'X-Kubernetes-Pf-Prioritylevel-Uid': '6ee0e2ec-80ac-421f-a6ff-dae05057166a', 'Date': 'Thu, 23 Mar 2023 09:41:49 GMT', 'Content-Length': '417')>
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"pods \"simple-job-default-worker-8fd7f483ae\" is forbidden: exceeded quota: default-quota, requested: requests.cpu=500m,requests.memory=500Mi, used: requests.cpu=1,requests.memory=1000Mi, limited: requests.cpu=1,requests.memory=1Gi","reason":"Forbidden","details":{"name":"simple-job-default-worker-8fd7f483ae","kind":"pods"},"code":403}
```

@jacobtomlinson I am happy to fix this, but I am not quite sure what the best strategy is here or what general k8s best practices are when dealing with restricted resources. I am also not sure about the internals of `kopf` (i.e. how often it retries), because if it would retry indefinitely, the `job` should start running eventually? Two concerns that I have: 
1. If we "just" retry, a large cluster might be up, idling, just waiting for the `job-runner` to come up
2. In case the scheduler does not come up in time, we might run into timeout issues (we've already increased our overall timeout to ~10 minutes to deal with node cold start, etc.)

**Minimal Complete Verifiable Example**:
[...]

PR #845, opened 2023-12-18
Add priorityClassName value to operator deployment

Artimi:
In https://github.com/dask/dask-kubernetes/issues/684 we were discussing that adding `priorityClassName` might help with configurability of dask kubernetes operator. Especially, settings [priority class](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/) to some high priority class might help operator being unable to start after restart because of other resources. Tested manually locally both with `priorityClassName` set and left out.


jacobtomlinson (approved):
Sorry for the delay in reviewing here. This looks great, thanks!

PR #852, opened 2024-01-04
Investigate CI

jacobtomlinson:
CI seems unhappy here lately. Pushing a small comment change to a file that should trigger all CI jobs.


jacobtomlinson:
It seems to be just the operator CI so tweaking the change only to trigger those workflows.

jacobtomlinson:
Reinstating the kr8s pin seems to unblock things. Perhaps we switched to a looser constraint too early.

PR #862, opened 2024-02-19
Bump max Kubernetes in CI to 1.29 and minimum to 1.26

jacobtomlinson:
Seeing some CI failures on Kubernetes 1.24 because some charts installed in CI do not support it. Taking the opportunity to upgrade to more recent Kubernetes versions.


jacobtomlinson:
Read the docs failure is unrelated

Issue #870, opened 2024-02-26
Dask Cluster with name longer than 53 chars is stuck in Created state, cannot be deleted


JohannaGoergen:
**Describe the issue**:

Creating a Dask cluster with 53+ characters in the name results in: 
* Cluster resource has `Created` status
* Operator is infinitely retrying to create the scheduler deployment with name `{cluster_name}-scheduler`. It is failing with 422 exception.
* DaskCluster resource cannot be deleted -- `kubectl delete daskcluster {clusterName}` hangs forever 

**Minimal Complete Verifiable Example**:

Example -- DaskCluster with 56 characters in name.

The `delete` command does not complete.
![Screenshot from 2024-02-26 17-35-22](https://github.com/dask/dask-kubernetes/assets/15880886/f0daa964-0d95-4046-9147-fdfd35191bf2)

Logs from the operator pod show:
```
[2024-02-26 16:30:03,681] kopf.objects         [ERROR   ] [default/abcdefgddddddddddddddddddddddddddddddddddddddddddddddddddddd] Handler 'daskcluster_create_components/status.phase' failed with an exception. Will retry.
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/kopf/_core/actions/execution.py", line 276, in execute_handler_once
    result = await invoke_handler(
  File "/usr/local/lib/python3.10/site-packages/kopf/_core/actions/execution.py", line 371, in invoke_handler
    result = await invocation.invoke(
  File "/usr/local/lib/python3.10/site-packages/kopf/_core/actions/invocation.py", line 116, in invoke
    result = await fn(**kwargs)  # type: ignore
  File "/usr/local/lib/python3.10/site-packages/dask_kubernetes/operator/controller/controller.py", line 313, in daskcluster_create_components
    await scheduler_service.create()
  File "/usr/local/lib/python3.10/site-packages/kr8s/_objects.py", line 232, in create
    async with self.api.call_api(
  File "/usr/local/lib/python3.10/contextlib.py", line 199, in __aenter__
    return await anext(self.gen)
  File "/usr/local/lib/python3.10/site-packages/kr8s/_api.py", line 134, in call_api
    response.raise_for_status()
  File "/usr/local/lib/python3.10/site-packages/httpx/_models.py", line 761, in raise_for_status
    raise HTTPStatusError(message, request=request, response=self)
[...]

Issue #826, opened 2023-10-03
Validate dask cluster name to contian only allowed characters 


dbalabka:
Having a "." in the cluster name leads to a hung cluster in the "Created" state:
```text
❯ kubectl get daskclusters
NAME                                     WORKERS   STATUS    AGE
dmitry.balabka-cluster              1         Created   17h
```

While it is described in official k8s docs about Object Names, it would be great to have a validation in the dask library that raises at least a warning.
https://kubernetes.io/docs/concepts/overview/working-with-objects/names/

Code example:
```python
from dask_kubernetes.operator import KubeCluster
 
# Create a Dask cluster by calling Kubernetes API via kubectl command based on provided specificatoin
cluster = KubeCluster(
    name="dmitry.balakba-cluster",
    namespace="dask-operator",
)
 
client = cluster.get_client()
 
# Renders cluster configuration in the cell output
client
```

dbalabka:
It is possible that the Linux user name contains `.`. Basically, it might happen even w/o specifying a custom name because generated cluster name does not sanitize the name as well:
https://github.com/dask/dask-kubernetes/blob/03aa9dbcc5b0eb87284eeac90d651e17688e9f62/dask_kubernetes/operator/kubecluster/kubecluster.py#L254-L256

Not sure should the validation happen in `dask-kubernetes` or `kr8s`/`kopf`

jacobtomlinson:
It's interesting that the resource gets created, but I'm guessing if you look at the controller logs there are errors there when it tried to create the child components. Ideally the Kubernetes API would reject this name and give us an error that we can propagate back to the user. Maybe this is something we can configure in the CRD?

Otherwise yeah we will need to add some client-side checking. Although maybe we should do that in `kr8s` so that more folks can benefit from this.

PR #871, opened 2024-02-27
Validate Dask Cluster Names

JohannaGoergen:
### Issues
https://github.com/dask/dask-kubernetes/issues/870 
https://github.com/dask/dask-kubernetes/issues/826

### Details

This PR introduces cluster name validation in order to avoid the invalid state in which a `DaskCluster` resource with a too-long or RFC-1123-noncompliant name is created but cannot be deleted while the operator retries infinitely to create a scheduler service (see https://github.com/dask/dask-kubernetes/issues/826 for more details).

### DIFF
--- a/dask_kubernetes/operator/controller/controller.py
+++ b/dask_kubernetes/operator/controller/controller.py
@@ -1,4 +1,5 @@
 import asyncio
+import re
 import time
 from collections import defaultdict
 from contextlib import suppress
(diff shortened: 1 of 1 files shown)


jacobtomlinson (commented):
Thanks for raising this!

I agree an admission time check would be cleaner. I wonder if there is a way to use [validation rules](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules) to achieve this.

We have a long term god to add better support for hooks to the controller. I don't mind adding the dependencies, but the CI and test machinery is the biggest roadblock. Mainly because our workflow is to run the controller outside of the cluster when developing/testing.

In the meantime I'm happy to merge this PR. Would you mind adding some tests that check for clusters with invalid names do go into an error state as expected?

Also it would be nice to move the checking logic somewhere common and also use it in `KubeCluster` as a client side check.

jo-migo:
Thanks @jacobtomlinson for your quick review! 

I've added some tests and updated `KubeCluster`  to use the same validation.

Good point about the validation rules. It looks like you could specify `metadata` on the [DaskCluster OpenAPI schema](https://github.com/dask/dask-kubernetes/blob/main/dask_kubernetes/operator/customresources/templates.yaml#L52) to be a custom [ObjectMeta schema](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/object-meta/) having:
```
   properties:
      name:
          type: string
          pattern: '^(?=.,53$)[a-z0-9]([-a-z0-9]*[a-z0-9])?$'
``` 

The only minor downside of that approach is that the 53 would be hard-coded instead of computed using the scheduler service name (`63 - len("-scheduler")`). An update to the scheduler service name without update to this regex could cause a regression.

Let me know what you prefer to go with.

PR #872, opened 2024-03-04
Fix: KeyError on Building Scheduler Name in Replica Count Decrease

jo-migo:
### Fix

I introduced a bug in https://github.com/dask/dask-kubernetes/pull/871 which will cause decreasing replicas in a worker group to fail on a key error because `format` requires `cluster_name` to be provided with a key.


jacobtomlinson (approved):
Thanks for catching this!

jo-migo:
Thanks for the quick merge @jacobtomlinson ! Do you plan to do a patch release?

jacobtomlinson:
@jo-migo done!

Issue #873, opened 2024-03-06
Cannot Overwrite DASK_SCHEDULER_ADDRESS in Worker env


efeboyaci:
**Describe the issue**:
To use a different port or non default FQDN for the scheduler we provide `DASK_SCHEDULER_ADDRESS` env in `daskcluster` `.spec.worker.spec.containers[0].env`, however the operator creates 2 items in worker deployment env. The last one overrides the first in the pod's env. So we cannot change `DASK_SCHEDULER_ADDRESS` environment variable from dask kubernetes objects.

**Minimal Complete Verifiable Example**:

```python
from dask_kubernetes.operator import KubeCluster
cluster = KubeCluster(name="test", n_workers=1, namespace="dask-operator", env={"DASK_SCHEDULER_ADDRESS": "tcp://test-scheduler:8786"})
```
DaskWorkerGroup
```yaml
env:
- name: DASK_SCHEDULER_ADDRESS
  value: tcp://dev-cluster-scheduler:8786
```

Worker Pod
```yaml
env:
- name: DASK_SCHEDULER_ADDRESS
  value: tcp://dev-cluster-scheduler:8786
- name: DASK_WORKER_NAME
  value: dev-cluster-default-worker-5f4e5766de
- name: DASK_SCHEDULER_ADDRESS
  value: tcp://dev-cluster-scheduler.dask-operator.svc.cluster.local:8786
```

**Anything else we need to know?**:

Duplicate items with same names in container["env"] and env should be removed from env before adding it to configuration
Current:
https://github.com/dask/dask-kubernetes/blob/b668cc62944dc3eac225dc4c4ff01f2b3743c9da/dask_kubernetes/operator/controller/controller.py#L166-L170
https://github.com/dask/dask-kubernetes/blob/b668cc62944dc3eac225dc4c4ff01f2b3743c9da/dask_kubernetes/operator/controller/controller.py#L202-L206

Fix:
```py
    for container in deployment_spec["spec"]["template"]["spec"]["containers"]:
        if "env" in container:
            # Remove duplicate env vars
            container_env_names = [env_item["name"] for env_item in container["env"]]
            for env_item in env:
                if env_item["name"] in container_env_names:
                    env.remove(env_item)
            # Add the env vars
            container["env"].extend(env)
        else:
            container["env"] = env
```
**Environment**:
2024.3.0 dask operator

jacobtomlinson:
[...]

PR #874, opened 2024-03-08
Enable overwrites of default environment variables

jonded94:
Should solve https://github.com/dask/dask-kubernetes/issues/873

### DIFF
--- a/dask_kubernetes/operator/controller/controller.py
+++ b/dask_kubernetes/operator/controller/controller.py
@@ -154,8 +154,8 @@ def build_worker_deployment_spec(
         "spec": spec,
     }
     worker_env = {
-            "name": "DASK_WORKER_NAME",
-            "value": worker_name,
+        "name": "DASK_WORKER_NAME",
+        "value": worker_name,
     }
     scheduler_env = {
         "name": "DASK_SCHEDULER_ADDRESS",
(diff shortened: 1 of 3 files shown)


jacobtomlinson (commented):
This is great many thanks! Would you mind adding a small test to verify this behaviour?

jonded94:
Was sure you'd ask that :smile:  I'm a bit unsure about that though. 

Changing the Scheduler address from the default address will likely lead to that Cluster being broken in the CI test? 
We could test though for a custom non-default Dask Worker name I guess? What do you think?

jacobtomlinson:
> We could test though for a custom non-default Dask Worker name I guess?

Yep that sounds like a great way to test it.

jonded94:
@jacobtomlinson sorry for the delay, was busy with other things. PR should now be ready; I included some tests that will use the apparently previously unused "simpleworkergroup.yaml" for deploying an additional worker group that overrides the envvar "DASK_WORKER_GROUP".
In the tests, it will check if the DASK_SCHEDULER still is intact.

jacobtomlinson (approved):
This looks great thanks!

jonded94:
Ah, sure, sorry, will do that too tomorrow. This one doesn't need a test though (overriding the scheduler address is anyways a bit weird to test).

PR #877, opened 2024-03-28
Add a new, more careful option for downscaling

sil-lnagel:
# Background / Problem

After my forum post https://dask.discourse.group/t/shuffle-p2p-unstable-with-adaptive-k8s-operator/2600 @hendrikmakait confirmed my suspicion that downscaling and reshuffling (especially with p2p) seem not to work well together.

For our test workload the grafana dashboard shows the following:
default_behavior

It is clearly visible that as soon as the scheduler recommends to retire one worker (ie.: 5->4) the worker is retired and the shuffle is (at least partially) restarted, as can be seen in the increase in waiting tasks. In the end it just "finishes" because shuffle fails "shuffle_barrier failed during shuffle 5da46c9c80547f11890d48656a58c7f4". Sometimes it works, but in most cases it either needs multiple shuffle restarts or fails at some point.

Since this is a feature, we need for our workloads which heavily depend on reshuffling by nature, we further investigated this issue and came up with a potential improvement for the operator.

# Proposed Solution

In the existing implementation of the operator it just relies on the scheduler to propose workers to close. If the worker is part of a shuffle this causes the shuffle to restart.

Our idea was the following:
- don't immediately retire the workers the scheduler suggests
- instead query their /metrics endpoint and verify that no tasks are running
- only retire if they do nothing, otherwise delay the scaling

This resolves the restart issues in all our test workloads, while still allowing scale down after work is finished.

Potential issues we forsee
- workers are never shut down until they are done even if they only have 1 task they are working on
- we attempted to assign worker states to `ACTIVE_STATES` and `IDLE_STATES` but we would appreciate any input if this partition was correct
- the get_workers rpc is not available as HTTP endpoint in dask.distributed, so we can only use RPC here

## Additional changes

Made a couple of hard coded values configurable, that we used to fine tune the scaling behavior (defaults should be unchanged)
- `kubernetes.controller.autoscaler.cooldown`
- `kubernetes.controller.worker.dashboard-port`
- `kubernetes.controller.worker-allocation.batch-size`
- `kubernetes.controller.worker-allocation.delay`

Added parameters to enable the scaling:
- `kubernetes.controller.autoscaler.method` (left the existing behavior as default, set to "careful" to enable)
- `kubernetes.controller.autoscaler.retry-delay`

careful
A picture of the proposed operator running smoothly :)

We are very interested in any feedback and improvement suggestions. Please let us know if this something that could improve the operator or if there are any questions / concerns.

### DIFF
--- a/dask_kubernetes/operator/controller/controller.py
+++ b/dask_kubernetes/operator/controller/controller.py
@@ -1,6 +1,7 @@
 import asyncio
 import copy
 import time
+from enum import Enum
 from collections import defaultdict
 from contextlib import suppress
 from datetime import datetime
@@ -41,14 +42,11 @@ KUBERNETES_DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
 
 DASK_AUTOSCALER_COOLDOWN_UNTIL_ANNOTATION = "kubernetes.dask.org/cooldown-until"
 
-ACTIVE_STATES = [
-    "cancelled",
+ACTIVE_STATES = (
     "constrained",
-    "error",
     "executing",
     "fetch",
     "flight",
-    "forgotten",
     "long-running",
     "memory",
     "missing",
@@ -56,9 +54,18 @@ ACTIVE_STATES = [
     "rescheduled",
     "resumed",
     "waiting",
-]
-DONE_STATES = ["released"]
+)
+IDLE_STATES = (
+    "released",
+    "error",
+    "cancelled",
+    "forgotton"
+)
 
+class WorkerState(Enum):
+    IDLE = 1
+    BUSY = 2
+    UNCERTAIN = 3
 
 
 # Load operator plugins from other packages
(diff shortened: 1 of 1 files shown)


hendrikmakait:
> Our idea was the following:
> - don't immediately retire the workers the scheduler suggests
> - instead query their /metrics endpoint and verify that no tasks are running
> - only retire if they do nothing, otherwise delay the scaling

I'm currently out on PTO, so I haven't had a closer look at the PR. I'll just note that the idea sketched out above is not _sufficient_ to prevent P2P restarts as workers might be idle due to straggling transfer tasks, but still required for the remainder of the shuffle. cc @fjetter

See also dask/distributed#8579 for a related issue/possible solution.

hendrikmakait:
No worries, I chose to answer after all :)

> a) If the controller or anyone asks the scheduler "what can I close" (rpc: workers_to_close), it should only suggest workers that do not take participate a P2P shuffle.

Yes, that's the basic idea behind @fjetter's comment on the other issue.

jacobtomlinson (commented):
I really appreciate the effort going into this here. I haven't looked deeply at the code yet as I'm concerned that this problem is being solved in the wrong place.

Dask Kubernetes shouldn't be worrying about the details of the downward scaling. It should just ask the scheduler periodically if the cluster needs to scale up or down and act accordingly. I feel like all of this logic around avoiding disrupting operations like shuffles should be handled in the scheduler. If a shuffle is in progress the scheduler shouldn't be suggesting to retire workers.

What do you think @fjetter @hendrikmakait?

PR #880, opened 2024-04-03
Bump kr8s to 0.14.2

### LINKED ISSUE #853: Update dask-kubernetes to a newer kr8s
Recently we unpinned `kr8s` but things broke, so the pin was set back to `0.9.0`. We need to manually update the pin to something more recent (and in theory more stable) before unpinning again.

jacobtomlinson:
Closes #853

### DIFF
--- a/dask_kubernetes/operator/_objects.py
+++ b/dask_kubernetes/operator/_objects.py
@@ -71,7 +71,7 @@ class DaskCluster(APIObject):
         return services[0]
 
     async def ready(self) -> bool:
-        await self._refresh()
+        await self.async_refresh()
         return (
             "status" in self.raw
             and "phase" in self.status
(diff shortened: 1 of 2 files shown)


jacobtomlinson:
I spent some time in #883 trying to figure out what change in `kr8s` was causing the CI deadlock.

It looks like the problem is because when we run the `dask-kubernetes` test suite both the controller and client code run within the same process, but on separate event loops. The client caching in `kr8s` meant that the same API client and other async objects was being shared between these loops, which caused the deadlock. Annoyingly this isn't a real world use case for `dask-kubernetes` and only happens in the test suite.

There was a fix in `kr8s` a while back which ensured that the client caching checked the current thread to ensure this doesn't happen, due to an assumption that different event loops must exist in different threads. However that seems to not be true in this case and both loops were reporting the same thread ID.

I tested out a fix directly from the `kr8s` PR in #883 and things seem to be passing. So I've made a `v0.14.2` release of `kr8s` and updated this PR to point to the new release which hopefully gets things unblocked.

PR #881, opened 2024-04-03
Draft: Introduce typehints in controller

jonded94:
General code quality improvements for the controller part of this repository.

- Make sure exported symbols are correctly set in `__init__.py` with `__all__`.
- Renamed pytest fixture for creating a k8s namespace from `ns` to `namespace` since `ns` is a bit overloaded, especially in `test_controller.py`
- Introduce typehints in `controller.py` and `test_controller.py` to massively improve developer experience (previously, one just really had to guess what functions are for and which types are expected)
- Removed unused function parameters
- Made function headers (i.e. parameter [type] defintion) compatible with what `kopf.on.[create/change/...]' expects
- Introduce mypy (and missing stubs of external packages if available)
- Introduce mypy ignores for (still) untyped parts of the codebase for now

### DIFF
--- a/dask_kubernetes/operator/controller/controller.py
+++ b/dask_kubernetes/operator/controller/controller.py
@@ -1,19 +1,22 @@
+from __future__ import annotations
+
 import asyncio
 import copy
 import time
 from collections import defaultdict
 from contextlib import suppress
 from datetime import datetime
+from typing import TYPE_CHECKING, Any, Awaitable, Final, cast
 from uuid import uuid4
 
 import aiohttp
 import dask.config
 import kopf
-import kr8s
+import kr8s  # type: ignore[import-untyped]
 import pkg_resources
 from distributed.core import clean_exception, rpc
 from distributed.protocol.pickle import dumps
-from kr8s.asyncio.objects import Deployment, Pod, Service
+from kr8s.asyncio.objects import Deployment, Pod, Service  # type: ignore[import-untyped]
 
 from dask_kubernetes.constants import SCHEDULER_NAME_TEMPLATE
 from dask_kubernetes.exceptions import ValidationError
@@ -26,18 +29,23 @@ from dask_kubernetes.operator._objects import (
 from dask_kubernetes.operator.networking import get_scheduler_address
 from dask_kubernetes.operator.validation import validate_cluster_name
 
-_ANNOTATION_NAMESPACES_TO_IGNORE = (
+if TYPE_CHECKING:
+    from distributed import Scheduler
+
+_ANNOTATION_NAMESPACES_TO_IGNORE: Final[tuple[str, ...]] = (
     "kopf.zalando.org",
     "kubectl.kubernetes.io",
 )
-_LABEL_NAMESPACES_TO_IGNORE = ()
+_LABEL_NAMESPACES_TO_IGNORE: Final[tuple[str, ...]] = ()
 
-KUBERNETES_DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
+KUBERNETES_DATETIME_FORMAT: Final[str] = "%Y-%m-%dT%H:%M:%SZ"
 
-DASK_AUTOSCALER_COOLDOWN_UNTIL_ANNOTATION = "kubernetes.dask.org/cooldown-until"
+DASK_AUTOSCALER_COOLDOWN_UNTIL_ANNOTATION: Final[
+    str
+] = "kubernetes.dask.org/cooldown-until"
 
 # Load operator plugins from other packages
-PLUGINS = []
+PLUGINS: list[Any] = []
 for ep in pkg_resources.iter_entry_points(group="dask_operator_plugin"):
     with suppress(AttributeError, ImportError):
         PLUGINS.append(ep.load())
@@ -47,7 +55,7 @@ class SchedulerCommError(Exception):
     """Raised when unable to communicate with a scheduler."""
 
 
-def _get_annotations(meta):
+def _get_annotations(meta: kopf.Meta) -> dict[str, str]:
     return {
         annotation_key: annotation_value
         for annotation_key, annotation_value in meta.annotations.items()
@@ -58,7 +66,7 @@ def _get_annotations(meta):
     }
 
 
-def _get_labels(meta):
+def _get_labels(meta: kopf.Meta) -> dict[str, str]:
     return {
         label_key: label_value
         for label_key, label_value in meta.labels.items()
@@ -69,15 +77,16 @@ def _get_labels(meta):
 
 
 def build_scheduler_deployment_spec(
-    cluster_name, namespace, pod_spec, annotations, labels
-):
-    labels.update(
-        **{
-            "dask.org/cluster-name": cluster_name,
-            "dask.org/component": "scheduler",
-            "sidecar.istio.io/inject": "false",
-        }
-    )
+    cluster_name: str,
+    pod_spec: kopf.Spec,
+    annotations: kopf.Annotations,
+    labels: kopf.Labels,
+) -> dict[str, Any]:
+    labels = dict(labels) | {
+        "dask.org/cluster-name": cluster_name,
+        "dask.org/component": "scheduler",
+        "sidecar.istio.io/inject": "false",
+    }
     metadata = {
         "name": SCHEDULER_NAME_TEMPLATE.format(cluster_name=cluster_name),
         "labels": labels,
@@ -101,13 +110,16 @@ def build_scheduler_deployment_spec(
     }
 
 
-def build_scheduler_service_spec(cluster_name, spec, annotations, labels):
-    labels.update(
-        **{
-            "dask.org/cluster-name": cluster_name,
-            "dask.org/component": "scheduler",
-        }
-    )
+def build_scheduler_service_spec(
+    cluster_name: str,
+    spec: kopf.Spec,
+    annotations: kopf.Annotations,
+    labels: kopf.Labels,
+) -> dict[str, Any]:
+    labels = dict(labels) | {
+        "dask.org/cluster-name": cluster_name,
+        "dask.org/component": "scheduler",
+    }
     return {
         "apiVersion": "v1",
         "kind": "Service",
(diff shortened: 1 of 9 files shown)


jonded94:
@jacobtomlinson I converged to a somewhat rounded midpoint now; there still are a handful of `mypy` errors in `controller.py` (`test_controller.py` now shows no errors even with `mypy --strict`), but solving these isn't too easy right now.

Let me know what you think about my PR in general; it is unfortunately quite large, but I did it to improve developer experience (I honestly found contributing to this repository a bit hard since it is almost impossible to know which types are used where). I would prefer not increasing the scope of this PR even further for now but to do only clean up.

Also, I did not include `mypy` in CI for now since there are still hundreds of errors because of other parts of the codebase of this repo. Let me know what you think about that.

Maybe I can find some time to also add type hints to other parts of this repo in the distant future.

dask_kubernetes/operator/controller/controller.py
```diff
@@ -69,15 +77,16 @@ def _get_labels(meta):
...
-            "dask.org/component": "scheduler",
-            "sidecar.istio.io/inject": "false",
-        }
-    )
+    cluster_name: str,
+    pod_spec: kopf.Spec,
+    annotations: kopf.Annotations,
+    labels: kopf.Labels,
+) -> dict[str, Any]:
+    labels = dict(labels) | {
```
jacobtomlinson, line 85: This is exciting! I hadn't realised union operators were in 3.9, which is our minimum version!

dask_kubernetes/operator/controller/controller.py
```diff
@@ -410,18 +472,18 @@ async def retire_workers(
     )
     # Dask version mismatches between the operator and scheduler may cause this to fail in any number of unexpected ways
     with suppress(Exception):
-        comm_address = await get_scheduler_address(
+        comm_address = await get_scheduler_address(  # type: ignore[no-untyped-call]
```
jacobtomlinson, line 475: There are a lot of these ignore throughout. Perhaps it makes sense to ignore this globally via the project config?

dask_kubernetes/operator/controller/tests/test_controller.py
```diff
@@ -1,15 +1,26 @@
...
+    Iterator,
+    TYPE_CHECKING,
+)
 
 import dask.config
 import pytest
 import yaml
 from dask.distributed import Client
-from kr8s.asyncio.objects import Deployment, Pod, Service
+from kr8s.asyncio.objects import Deployment, Pod, Service  # type: ignore[import-untyped]
```
jacobtomlinson, line 23: Same with this ignore, should we disable this for the whole project?

jacobtomlinson (commented):
Thanks for taking the time here.

Generally our policy in Dask projects is to do type hinting on a best efforts basis. It is by no means mandatory and I don't think we ever want to enforce it here as it will likely obstruct novice contributors.

I'm more than happy to accept PRs like this that improve hinting provided the code becomes more readable and continues to function the same. If you want to invest more time on this in the future that would be great too, but accepting this PR isn't conditional on you following it through to `mypy` being 100% happy.

Unfortunately the CI is failing so I'm guessing these changes have broken something. Would you be able to investigate the errors and fix things up?

dask_kubernetes/operator/controller/controller.py
jonded94: The change here actually also removes some remaining in-place modifications of mutable variables :)

dask_kubernetes/operator/controller/controller.py
jonded94: I'm personally more on the "the more explicit and correct, the better"-side (because ignoring it globally *could* potentially mask bugs, depending on which error you want to ignore) but I'm happy to use the way that works for everybody. Note that these untyped calls would actually vanish as soon as the entire repo would be properly typed.

jonded94:
@jacobtomlinson could you help me out with this piece of code here:

In the top-level `__init__.py` the `KubeCluster` is defined under `__all__`, but it is actually not imported "properly" but instead a `__getattr__` is defined that is doing a manual `importlib.import_module` pointing to `dask_kubernetes.classic.KubeCluster`:
https://github.com/dask/dask-kubernetes/blob/main/dask_kubernetes/__init__.py#L22

This class actually is very much different from the `KubeCluster` class defined [here](https://github.com/dask/dask-kubernetes/blob/main/dask_kubernetes/operator/kubecluster/kubecluster.py#L54) which I used in all my projects. 

Why is there a differentation, why is the default pointing to the "classic" `KubeCluster` and why is it done in this way? This smells very anti-pattern, but maybe I'm just not seeing something important here.

The CI errors actually stemmed from me trying to importing `KubeCluster` from the place which I thought to be "right", e.g. the `dask_kubernetes.operator.kubecluster` one.

jacobtomlinson:
@jonded94 it was done this way when we deprecated the classic `KubeCluster` implementation.

In the past users would do `from dask_kubernetes import KubeCluster`. But when we rewrote things with the operator we pushed things down into submodules so you would import `KubeCluster` from either `dask_kubernetes.classic` or `dask_kubernetes.operator`. The `getattr` was added so we could tell people to migrate over.

Enough time has passed now that we can do away with all of this. The classic codebase can be totally removed. Then we need to make a decision about should we allow `from dask_kubernetes import KubeCluster` to import the operator version of the class, or should we enforce everyone to do `from dask_kubernetes.operator import KubeCluster`. This is mainly to avoid very old codebases breaking.

jonded94:
Okay, so the idea would be to include a DeprecationWarning *in the future* then as soon somebody is importing that classic thing, got it (because right now it simply imports it with no further warning or output). This is something separate then from this PR :) Thanks for your clarification!

jacobtomlinson:
Yeah, it was actually there previously, but we moved it again because there were some usage patterns that meant it didn't show up.

https://github.com/dask/dask-kubernetes/blob/2a0465273f0ce539e8405c4a4eb1d25e8f92c3d8/dask_kubernetes/classic/kubecluster.py#L455

I basically consider the classic code overdue for removal. I just haven't had the time lately to do the work. Hoping to get to it very soon.

jacobtomlinson:
I took the plunge and removed the classic codebase in #890. Would you mind rebasing this PR?

jonded94:
Very nice to hear! :) Rebased my PR.

dask_kubernetes/operator/controller/controller.py
jacobtomlinson: I think the thing I would say here is that Dask strives to be a project that is welcoming to novice developers. My concern with type hinting in general is it can be very daunting for junior developers, even if it helps them avoid shooting their foot off.

Removing inline ignores I feel is a good way to remove some visual clutter that may be daunting for some folks. So even though it may mask a bug here and there I would prefer a global ignore.

dask_kubernetes/operator/controller/controller.py
```diff
@@ -453,7 +519,7 @@ async def check_scheduler_idle(scheduler_service_name, namespace, logger):
                 idle_since = (await resp.json())["idle_since"]
                 if idle_since:
                     logger.debug("Scheduler idle since: %s", idle_since)
-                return idle_since
+                return cast(str, idle_since)
```
jacobtomlinson, line 522: I'm curious why not `str(idle_since)`?

dask_kubernetes/operator/controller/controller.py
```diff
@@ -525,35 +602,52 @@ async def get_desired_workers(scheduler_service_name, namespace, logger):
...
-    name, namespace, old, new, **kwargs
-):
+    name: str | None,
+    namespace: str | None,
+    old: Any | None,
+    new: Any | None,
+    **__: Any,
+) -> None:
+    if name is None or namespace is None:
+        raise kopf.PermanentError("Name or Namespace not given.")
```
jacobtomlinson, line 630: Are there really cases where `kopf` would call this with `None` values? I'm not sure we should be so aggressive with checking hooks from dependencies like this.

jacobtomlinson (commented):
Thanks! I've given it another pass and have a few more comments.

dask_kubernetes/operator/controller/controller.py
jonded94: I wasn't actually too sure if this really is a `str` here (after all, we're just accessing some field in a JSON dictionary, could be anything really). `cast` will make the type checker happy without changing any runtime behaviour.

Honestly, this was more of a hack to finalize getting `mypy` to 0 errors, we can do here whatever you want:
1. Leaving out everything, will result in a `mypy` error
2. `assert isinstance(idle_since, str)` before the `return`
3. `return cast(str, idle_since)`
4. `return str(idle_since)`

Let me know what you'd prefer. I'd personally go for 2. or 4., but this would introduce a slight runtime behaviour change.

dask_kubernetes/operator/controller/controller.py
jonded94: I'm actually not too familiar with `kopf` and what a circumstance is where `name` or `namespace` could be `None` (probably some error cases where *something* still wants to be cleaned up or at least logged).

The test for the correct Callable is defined here in kopf: https://github.com/nolar/kopf/blob/9243c8fbb2a36c1b7cad1f65a647d3b9876a5b17/kopf/_core/intents/callbacks.py#L109

As you can see, it expects any "on change" Callable to be typed that it also could expect `None` for `name` or `namespace`. Because Callables are contravariant, giving it a Callable where `name` can't be `None` will result in a `mypy` error on applying the decorator `kopf.on.[...]` to this function here.
The other solution would be to leave out checking for `None` *inside* the function, but then every access to these variables would result in `mypy` errors since they potentially could be `None`.

Maybe we could replace this with an `assert`, then it at least would be a bit less visually verbose?

dask_kubernetes/operator/controller/controller.py
jonded94: Sure, if you'd prefer a global ignore for `no-untyped-call` & `import-untyped`, we can integrate this into the `pyproject.toml`. I'll do it when I find time for it.

dask_kubernetes/operator/controller/controller.py
jacobtomlinson: I think we should be flexible here. Users will likely pass either a `str` or `int` to this config option, so 4 seems to meet my expectations. Improving the runtime behaviour is fine with me.

dask_kubernetes/operator/controller/controller.py
jacobtomlinson: Adding assertions where we make assumptions like this make the most sense to me. Thanks!

jonded94:
I now fixed the remarks, i.e.:
- removed `#type ignore: [...]` comments to exchange them with global ignores in `pyproject.toml`
- fixed a wrong type annotation of me regarding `idle_since` - that thing is a `float` not a `str`, sorry
- removed visually taxing tests for `None`-ness and manual Exception raises and replaced them with simple `assert`'s
- added type annotations for functionality that was introduced in the mean time

After merging this MR, the next step surely would be to introduce `mypy` in the CI, at least for `dask_kubernetes.operator.controller`? (since that now shows 0 errrors)

dask_kubernetes/operator/controller/controller.py
jonded94: Added `asserts` now.

dask_kubernetes/operator/controller/controller.py
jonded94: Was `float` all along, sorry. This removed the need for some casts (simply because they were wrong before).

jacobtomlinson (approved):
This looks great to me. 

Very happy to add `mypy` to the `pre-commit` checks just for the controller for now. Let's do that in a follow up PR.

PR #883, opened 2024-04-18
Experiment with bumping kr8s version

jacobtomlinson:
In #880 I've tried bumping to the latest kr8s, but something is causing the test suite to hang. I can reproduce the hang locally too, but when running tests individually they all pass. When running tests with a debugger they all pass. 

In this PR I'm going to try and gradually bump the version forward to see if I can isolate which release causes the issues.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,4 +5,4 @@ kubernetes-asyncio>=12.0.1
 kopf>=1.35.3
 pykube-ng>=22.9.0
 rich>=12.5.1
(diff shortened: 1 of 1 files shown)


jacobtomlinson:
Ok looks like the first commit that starts to cause CI hangs is https://github.com/kr8s-org/kr8s/pull/189

jacobtomlinson:
Switching back to #880 to land this change.

PR #889, opened 2024-04-29
Migrate to hatch

jacobtomlinson:
Upgrading project to a modern build system.

- Use `hatch` for building
- Use `hatch new --init` to migrate `setup.py`, `setup.cfg` and `requirements.txt` to `pyproject.toml`
- Replace `versioneer` with `hatch-vcs`
- Use `hatch build` in CI release process
- Added workflow to [PyPI trusted publish management](https://docs.pypi.org/trusted-publishers/) and remove secret


dask_kubernetes/__init__.py
```diff
@@ -15,7 +15,13 @@
...
 
 __all__ = ["HelmCluster", "KubeCluster"]
-__version__ = _version.get_versions()["version"]
+
+try:
+    from ._version import version as __version__  # noqa
+    from ._version import version_tuple as __version_tuple__  # noqa
+except ImportError:
+    __version__ = "0.0.0"
+    __version_tuple__ = (0, 0, 0)
```
jacobtomlinson, line 24: The `_version.py` file is no longer kept in version control and instead is generated when running `pip install -e .` or when packaging the distribution for PyPI/Conda Forge. So it's possible to clone the repo and never install it, so we catch the `ImportError` and use a null version so imports at least work.

.github/workflows/release.yml
```diff
@@ -8,15 +8,17 @@ jobs:
 
     steps:
       - name: Checkout source
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
jacobtomlinson, line 13: Need the full history so that `setuptools-scm` can calculate the version correctly.

.dockerignore
```diff
@@ -1,2 +1,2 @@
-.*
```
jacobtomlinson, line 1: This was stopping the `.git` directory from being copied in which is needed to correctly resolve the version number. To avoid this blowing up the container size I've switched the `Dockerfile` to do a multi-stage build where we build the distributions and then only copy those over to the final container image.

PR #905, opened 2024-09-02
Fix Autoscaler example code snippet on the docs index page

### LINKED ISSUE #904: Incorrect example code snippet in the Autoscalers section of the docs index page
**Describe the issue**:

In the [Autoscalers section of the docs index page](https://kubernetes.dask.org/en/latest/#autoscalers), there's an incorrect example code snippet on how to create a `KubeCluster` and scale it. Here is screenshot:

image

Initially, I thought it was a duplicate snippet but @jacobtomlinson has [pointed out it's actually an incorrect example and it should depict how to adapt a cluster to use the autoscaler](https://github.com/dask/dask-kubernetes/pull/905#pullrequestreview-2276582837).

marcelovilla:
Fixes #904

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -152,6 +152,12 @@ Autoscalers
 
 A ``DaskAutoscaler`` resource will communicate with the scheduler periodically and auto scale the default ``DaskWorkerGroup`` to the desired number of workers.
 
+.. code-block:: python
+
+    from dask_kubernetes.operator import KubeCluster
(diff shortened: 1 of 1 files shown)


jacobtomlinson (changes requested):
Thanks for spotting this. It's not actually a duplicate, it is just a broken example.

Instead of saying `cluster.scale(10)` it should say `cluster.adapt(minimum=1, maximum=10)`.

Would you mind updating the example?

jacobtomlinson (approved):
Thanks for the quick turnaround here!

PR #908, opened 2024-09-17
Fix loadBalancer attribute name

### LINKED ISSUE #906: KubeCluster with service type LoadBalancer on AKS stuck in "Pending" state on startup
**Describe the issue**:
I'm trying to create a KubeCluster with AKS LoadBalancer, but the DaskCluster is stuck in a pending state at startup. 

Trying to create a cluster like this: 

```python
def make_sim_cluster(
        name: str,
        cpu_cores: int,
        memory_in_gb: int, 
        image: str = '<image>',
        tag: str = 'x.x.x',
        namespace: str = 'namespace',
):
    base_cluster_spec = make_cluster_spec(
        name=name,
        image=f'{image}:{tag}', 
        n_workers=cpu_cores,
        scheduler_service_type="LoadBalancer",
    )
    base_cluster_spec['metadata']['annotations'] = {
        "service.beta.kubernetes.io/azure-load-balancer-internal": "true",
    }

    return KubeCluster(custom_cluster_spec=base_cluster_spec, namespace=namespace)
```

but the DaskCluster is stuck in Pending State until it times out 
image
image

I looked around at our service spec and it seems like our resource definition has our load balancer under a key "loadBalancer" when the scheduler service comes up (as referenced in kubernetes doc https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer) 
image
 while dask-kubernetes seems to look around for "load_balancer" in status: https://github.com/dask/dask-kubernetes/blob/2ecfdcd41b0e09665d4ff98cb40528218fea10fe/dask_kubernetes/operator/controller/controller.py#L403-L406

for what it's worth, the kr8s library also seems to look around for "load_balancer".

When I switched these over to using "loadBalancer" the DaskCluster was able to come up without any issues. Actually even when I didn't change the code, the service, pods, deployments etc were all able to be setup and in a running state, only the DaskCluster was pending. 

Any helps is great, thanks!

**Environment**:

- Dask version:
    - dask-core: 2024.2.1
    - dask-kubernetes: 2024.5.0
- Python version: 3.11.9
- Operating System: RHEL 7.9 (Maipo)
- Install method (conda, pip, source): conda

catzc:
Closes #906

### DIFF
--- a/dask_kubernetes/operator/controller/controller.py
+++ b/dask_kubernetes/operator/controller/controller.py
@@ -401,7 +401,7 @@ async def handle_scheduler_service_status(
     assert namespace
     # If the Service is a LoadBalancer with no ingress endpoints mark the cluster as Pending
     if spec["type"] == "LoadBalancer" and not len(
-        status.get("load_balancer", {}).get("ingress", [])
+        status.get("loadBalancer", {}).get("ingress", [])
     ):
         phase = "Pending"
     # Otherwise mark it as Running
--- a/dask_kubernetes/operator/networking.py
+++ b/dask_kubernetes/operator/networking.py
@@ -50,8 +50,8 @@ async def get_external_address_for_scheduler_service(
     """Take a service object and return the scheduler address."""
     if service.spec.type == "LoadBalancer":
         port = _get_port(service, port_name)
-        lb = service.status.load_balancer.ingress[0]
-        host = lb.hostname or lb.ip
+        lb = service.status.loadBalancer.ingress[0]
+        host = lb.get("hostname", None) or lb.ip
     elif service.spec.type == "NodePort":
         port = _get_port(service, port_name, is_node_port=True)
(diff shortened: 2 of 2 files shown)


catzc:
On top of https://github.com/dask/dask-kubernetes/pull/907

dask_kubernetes/operator/controller/controller.py
```diff
@@ -387,7 +387,7 @@ async def daskcluster_create_components(
         await worker_group.create()
     logger.info(f"Worker group {data['metadata']['name']} created in {namespace}.")
 
-    patch.status["phase"] = "Pending"
+    patch.status["phase"] = "Running"
```
catzc, line 390: I need to build the controller and test that it works

jacobtomlinson (approved):
This looks fine to me. Thanks @catzc

Issue #855, opened 2024-01-08
TCTU Bug while scaling down workers 


BitTheByte:
The currently implemented logic for deletion can be summarized into:

1) The operator asks the scheduler to retire n of the workers 
2) The scheduler retires them (process exits) and returns to the operator names of retired workers 
3) Operator deletes worker deployment sequentially

Ref: https://github.com/dask/dask-kubernetes/blob/main/dask_kubernetes/operator/controller/controller.py#L600-L611

However, between 2 and 3 the Kubernetes API may interfere and restart the worker deployment so a new pod will be created and join the cluster for some time before the operator deletes the deployment effectively interpreting the pod mid-run.

BitTheByte:
Is it possible to allow the scheduler to delete the deployment directly? maybe with something like a plugin or so?

jacobtomlinson:
Are you seeing this in practice? Or is this a hypothetical race condition?

Each worker should have a unique ID, and when the scheduler retires it then a worker with that ID cannot reconnect in the future. So even if Kubernetes restarts the Pod in the time before the delete call happens and the Pod is cascade deleted the new worker should just repeatedly fail to connect to the scheduler.

> Is it possible to allow the scheduler to delete the deployment directly?

I don't think we should be giving the scheduler the ability to interact with the Kubernetes API as that would require us to give permissions to the scheduler Pod which can execute arbitrary user code.

BitTheByte:
> Are you seeing this in practice? Or is this a hypothetical race condition?

Yes, I'm seeing this during large-scale cluster scaling e.g. 100-200 workers

briceruzand:
I'am having that behavior each time I try to scale down even with 2 workers (see #856), so my cluster never scale down.
I need to delete the cluster and to create a new one with less replicas ;-(

briceruzand:
Hi,
Here are the logs of my behavior.



Scheduler logs : 

```logs
[2024-02-06 11:02:24] INFO     distributed/scheduler.py:1685            : State start
[2024-02-06 11:02:24] INFO     distributed/scheduler.py:3998            :   Scheduler at:    tcp://100.64.5.23:8786
[2024-02-06 11:02:24] INFO     distributed/scheduler.py:4013            :   dashboard at:  http://100.64.5.23:8787/status
[2024-02-06 11:02:24] INFO     distributed/scheduler.py:7526            : Registering Worker plugin shuffle
[...]

Issue #910, opened 2024-10-14
Operator unable to delete Kubernetes Deployment 


thaisarcanjo-ow:
There is an issue with the default settings available from the docs where the Operator tries to delete a Kubernetes Deployment using the wrong name and therefore cannot find. The Operator tries to delete a Deployment that is named like the Pod name, which doesn't exist.

**Reproducing steps:**
1. Install the Operator with `helm install --repo https://helm.dask.org --create-namespace -n dask-operator --generate-name dask-kubernetes-operator`, ie [this quick start step](https://kubernetes.dask.org/en/latest/index.html#quickstart).
2. Create the cluster using the default yaml available from [this guide](https://kubernetes.dask.org/en/latest/operator_resources.html#daskcluster) as is. At this stage, two workers would be available from two deployments.
3. Create an autoscaler with the min workers set to 0 and install it

```
# autoscaler.yaml
apiVersion: kubernetes.dask.org/v1
kind: DaskAutoscaler
metadata:
  name: simple
spec:
  cluster: "simple"
  minimum: 0  # we recommend always having a minimum of 1 worker so that an idle cluster can start working on tasks immediately
  maximum: 10 # you can place a hard limit on the number of workers regardless of what the scheduler requests
```
4. Apply this AutoScaler settings:
```
kubectl apply -f autoscaler.yaml
daskautoscaler.kubernetes.dask.org/simple created
```
5. At this stage, the operator would already try to remove some deployments, but it is attempting to delete a Deployment resouirce that matches the Pod name, which doesn't exist:
```
[2024-10-14 09:22:42,559] kopf.objects         [INFO    ] [default/simple] Autoscaler updated simple worker count from 2 to 1
[2024-10-14 09:22:42,559] kopf.objects         [INFO    ] [default/simple] Timer 'daskautoscaler_adapt' succeeded.
[2024-10-14 09:22:42,662] httpx                [INFO    ] HTTP Request: GET https://10.96.0.1/apis/kubernetes.dask.org/v1/namespaces/default/daskclusters?fieldSelector=metadata.name%3Dsimple "HTTP/1.1 200 OK"
[2024-10-14 09:22:42,668] httpx                [INFO    ] HTTP Request: GET https://10.96.0.1/apis/apps/v1/namespaces/default/deployments?labelSelector=dask.org%2Fworkergroup-name%3Dsimple-default "HTTP/1.1 200 OK"
[...]

PR #912, opened 2024-10-15
Improve retire_worker fallback by finding deployments name instead of pod name

RossmacD:
Fallback to finding deployments rather than worker for deletion, as otherwise the pod can have the incorrect name, leading to workers not scaling down in adaptive mode.

Believe this was part of the issue on https://github.com/dask/dask-kubernetes/issues/910

### DIFF
--- a/dask_kubernetes/operator/controller/controller.py
+++ b/dask_kubernetes/operator/controller/controller.py
@@ -488,7 +488,7 @@ async def retire_workers(
         "This can result in lost data, see https://kubernetes.dask.org/en/latest/operator_troubleshooting.html."
     )
     workers = await kr8s.asyncio.get(
-        "pods",
+        "deployments",
         namespace=namespace,
         label_selector={"dask.org/workergroup-name": worker_group_name},
     )


jacobtomlinson (approved):

briceruzand:
Thanks @RossmacD , it looks like my trouble https://github.com/dask/dask-kubernetes/issues/855#issuecomment-1909607721 (`404` on deployments, because using pods name instead of deployments name)
I will try this fix on next release.

briceruzand:
Try the fix. Thx a lot, my dask cluster can now scale down. :fireworks: 

But it never scale down to 0, do you have any idea ?
I need to do some adjustment to avoid scale up/down flapping ;-)

Do you plan to release a new version with that fix ?

briceruzand:
resolve #855

fcourtial:
@jacobtomlinson could we make a release with this fix ?

The current version prevents the dask auto scaler to down scale, so if the cluster creates 100 workers, they will never be deleted and be restarted given the pods comes from a Deployment.

We cannot really use the auto scaling because of the cost of the Pods kept alive.

Best regards.

jacobtomlinson:
Sure @fcourtial I just tagged `2025.1.0`.

fcourtial:
Thanks @jacobtomlinson !

Issue #927, opened 2025-01-16
Golang linter is unhappy


jacobtomlinson:
Linting is failing with some golang issues

https://github.com/dask/dask-kubernetes/actions/runs/12813426873/job/35727461396

@bstadlbauer would you mind taking a look?

bstadlbauer:
Sorry for this taking a bit, should be fixed in https://github.com/dask/dask-kubernetes/pull/928

PR #928, opened 2025-01-21
Remove generated `go` code before regeneration

bstadlbauer:
In https://github.com/dask/dask-kubernetes/issues/927, we've seen that the go-linter is failing. The reason being that some of the files did not update their version from `v2024` to `v2025` in the yearly update PR (https://github.com/dask/dask-kubernetes/pull/925). After closer inspection, it seems like the code generation is non-determistic. While the version of the code-generator is pinned to a specific tag ([here](https://github.com/dask/dask-kubernetes/blob/27ef9592339fc095cdd6537d8e820eb9933dd48b/dask_kubernetes/operator/go_client/hack/Dockerfile#L5)), there seem to be some loose dependencies which cause this. The root problem was that some files did not get regenerated (but also not deleted), thus, their version was not updated.

I didn't have the time to look further into the non-determinism of the issue, but made sure that this won't happen again by first removing the generated code before regeneration. The files that did not get regenerated in this version (the `client/applyconfiguration` submodule) should not be too important.

Closes https://github.com/dask/dask-kubernetes/issues/927


jacobtomlinson (approved):
This is great thanks!

PR #934, opened 2025-02-23
Reflect that `env` will set on scheduler too

CarterFendley:
Minor docs change to reflect that the `env` parameter of `KubeCluster` will set the environment variables on the scheduler (in addition to workers).

See that this is the case [here](https://github.com/dask/dask-kubernetes/blob/30cc719850afb6a5d4482e0df867d2368db000e0/dask_kubernetes/operator/kubecluster/kubecluster.py#L877) in `make_cluster_spec(...)`, that function's doc string is correct. But if you are like me you may only look at the doc string on `KubeCluster` in an LSP annotation / [on the docs](https://kubernetes.dask.org/en/latest/operator_kubecluster.html#api)


jacobtomlinson (approved):
Thanks!

Issue #895, opened 2024-06-14
Handle `image`, `env` and `args` fields updates in DaskCluster in k8s operator


Fogapod:
I have a permanent dask cluster in kubernetes. Current operator ignores all changes to manifest.
There has been an issue about supporting spec updates, it got closed as resolved after implementing scale field support: https://github.com/dask/dask-kubernetes/issues/636.

The only fields that cause changes to deployment after applying updated manifest are `spec.worker.replicas`, `DaskAutoscaler` min/max.

Is it possible to support other fields, specifically `image`, `args`, `env`, `volumes/mounts`?
If not, what could be the optimal way to gracefully shut down and update cluster?

Cluster manifest (mostly copypasted from example):
```yaml
---
apiVersion: kubernetes.dask.org/v1
kind: DaskCluster
metadata:
  name: dask-primary
spec:
  worker:
    replicas: 1
    spec:
      containers:
      - name: worker
        image: "//backend/dask:image"
        imagePullPolicy: Always
        args:
          - worker
          - --name
          - $(DASK_WORKER_NAME)
          - --dashboard
          - --dashboard-address
          - "8788"
        ports:
          - name: http-dashboard
            containerPort: 8788
            protocol: TCP
        env:
          - name: ENV_1
            value: 1
          - name: ENV_2
            value: 2
        volumeMounts:
          - name: kafka-certs
            mountPath: /etc/ssl/kafka/ca.crt
            subPath: ca.crt
            readOnly: true

      volumes:
        - name: kafka-certs
          configMap:
            name: kafka-certs

  scheduler:
    spec:
      containers:
      - name: scheduler
        image: "//backend/dask:image"
        imagePullPolicy: Always
        args:
          - scheduler
        ports:
          - name: tcp-comm
            containerPort: 8786
            protocol: TCP
          - name: http-dashboard
            containerPort: 8787
            protocol: TCP
        readinessProbe:
          httpGet:
            port: http-dashboard
            path: /health
          initialDelaySeconds: 5
          periodSeconds: 10
        livenessProbe:
          httpGet:
            port: http-dashboard
            path: /health
          initialDelaySeconds: 15
          periodSeconds: 20
      imagePullSecrets:
        - name: regcred
    service:
      type: ClusterIP
      selector:
        dask.org/cluster-name: dask-primary
        dask.org/component: scheduler
      ports:
      - name: tcp-comm
        protocol: TCP
        port: 8786
        targetPort: "tcp-comm"
[...]

PR #936, opened 2025-04-02
Patch long lived DaskCluster components on update

edgar-s-silva-alb:
Implementation for #895.

I'm not super familiar with kopf and kubernetes code in general, so some input would be appreciated.

### DIFF
--- a/dask_kubernetes/operator/controller/controller.py
+++ b/dask_kubernetes/operator/controller/controller.py
@@ -40,9 +40,7 @@ _LABEL_NAMESPACES_TO_IGNORE: Final[tuple[str, ...]] = ()
 
 KUBERNETES_DATETIME_FORMAT: Final[str] = "%Y-%m-%dT%H:%M:%SZ"
 
-DASK_AUTOSCALER_COOLDOWN_UNTIL_ANNOTATION: Final[
-    str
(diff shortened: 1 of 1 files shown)


edgar-s-silva-alb:
@jacobtomlinson Hey, from my testing this solution is not currently working fully, but does work on some cases with the scheduler. 
I would like to get on your input if this looks like the right path.

dask_kubernetes/operator/controller/controller.py
```diff
@@ -389,6 +376,92 @@ async def daskcluster_create_components(
...
+    name: str | None,    
+    namespace: str | None,
+    diff: kopf.Diff,     
+    patch: kopf.Patch,   
+    logger: kopf.Logger,
+    **__: Any
+):
+    """When the DaskCluster resource is updated update all the components."""
+    assert name
+    assert namespace
```
jacobtomlinson, line 393: If either of these assertions fail then kopf will retry automatically forever. We probably need to catch the assertion error and just return if this isn't what we want to hapen.

jacobtomlinson (commented):
This approach generally looks reasonable to me. Feel free to ping me for full review when ready.

Could you make sure you're running `pre-commit` to fix up linting issues. It looks like your editor has changed a bunch of things that are formatting related.

PR #954, opened 2025-06-16
Docs/jupyterhub dask dashboard

Montes-L:
I added some explanations to clarify and provide a concrete example that can be used directly to secure Kubernetes deployments.


jacobtomlinson (commented):
Thanks. It looks like there are merge conflicts here. Could you pull from main and resolve them?

PR #961, opened 2025-10-10
Test support for grouped workers

jacobtomlinson:
Running CI here with changes from https://github.com/dask/distributed/pull/9121 to check compatibility.


jacobtomlinson:
I opened ip #962 to see if these errors are related and so far it looks like they are not as I can see the same errors in [both](https://github.com/dask/dask-kubernetes/actions/runs/18401431189/job/52431423980?pr=962). I'll get those cleaned up and then update this PR.

PR #963, opened 2025-10-14
Update docs theme and remove docs env pins

jacobtomlinson:
- Removed pins from #865 to allow docs builds with newer Sphinx versions now that we have `dask-sphinx-theme` version `4`.
- Fixed some broken docs references and syntax highlighting

xref https://github.com/dask/dask-sphinx-theme/issues/68#issuecomment-1893198477


jacobtomlinson:
Docs build looks good and is passing so going to merge

PR #967, opened 2026-01-06
Update operator Dockerfile to Python 3.12

chrisspalm:
## Summary
- Bump Python version in operator Dockerfile from 3.10 to 3.12
- Updates both builder and runtime stages

## Motivation
Python 3.12 brings performance improvements and is well-supported by the project (already in CI matrix).

## Test plan
- [ ] CI passes
- [ ] Operator builds successfully with Python 3.12 base image

### DIFF
--- a/dask_kubernetes/operator/controller/tests/test_controller.py
+++ b/dask_kubernetes/operator/controller/tests/test_controller.py
@@ -169,7 +169,7 @@ def test_operator_plugins(kopf_runner: KopfRunner) -> None:
 
     assert runner.exit_code == 0
     assert runner.exception is None
-    assert "Plugin 'noop' running." in runner.stdout
+    assert "Plugin 'noop' running." in runner.output
 
 
 @pytest.mark.timeout(180)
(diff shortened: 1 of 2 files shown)


chrisspalm:
The test failures appear to be pre-existing flaky tests unrelated to this change:

- `test_operator_plugins` - Checks `runner.stdout` but the plugin message goes to captured log instead
- `test_job` / `test_failed_job` - Timing issues with kopf runner

These same tests fail across all Python versions (3.10, 3.11, 3.12) in the matrix, and the main branch also shows [similar test failures](https://github.com/dask/dask-kubernetes/actions/runs/18496837287) from October 2025.

The actual change (Dockerfile base image update) is straightforward and doesn't affect test behavior.

chrisspalm:
Added a fix for the flaky tests in the second commit.

**Root cause:** Tests were checking `runner.stdout` but KopfRunner exposes captured output via `runner.output`. The `stdout` attribute was always empty.

**Fix:** Changed all occurrences of `runner.stdout` to `runner.output` in:
- `test_operator_plugins`
- `test_job`  
- `test_failed_job`

This should fix the intermittent test failures.

jacobtomlinson (approved):

